Issue #134, opened 2015-01-14
Get rid of `Req <: HttpRequest` in Endpoint


vkostyukov:
A very interesting issue come up today after the discussion on Gitter. For now, if you want to use rich request type you have to extends `httpx.Request` since `Endpoint` has a constraint on its type parameter `Req` so `Req <: HttpRequest`. This is needed for routing (more precisely for converting the endpoint into service) since we access `path` and `method` members on request.

So far so good. But the thing is that `httpx` has no longer contain `RequestProxy` class, which you used to use for making rich requests. Finagle is trying to say that it's not a good idea to subtype requests, but you can always to wrap it with something. Ideally, `HttpRequest` should be immutable case class.

So the problem here is if users wish to use rich requests types the only right way to do that is to wrap with case class like this:

```scala
case class OAuth2Request(underlying: httpx.Request, userId: Long)
```

But the thing is you can't really use it as a request type in the endpoint since it's no longer extends `Request`. In this case you would probably make your endpoint takes `HttpRequest` instead and mix your `authorize` filer before service like this:

```scala
val authorize: Filter[HttpRequest, Rep, OAuth2Request, Rep] = ???
val getUsers: Service[OAuth2Request, Rep] = ???
val e = Endpoint[HttpRequest, Rep] {
  case Get / "users" => authorize ! getUser
}
```

Which is doesn't look like "Do not repeat yourself" approach.

So, my opinion on this issue is that we have to somehow allow the `Endpoint` accept just `Req` so users can create rich request types with composition not inheritance. So, the question is how can we do routing then.

This tickets relates to #89.

@pasviegas, @benjumanji, @rpless, @rodrigopr, @BenWhitehead: What do you guys think about this issue?

vkostyukov:
What I'm thinking about right now is to have an endpoint defined as `Endpoint[Req, Rep]` and have method `toService` defined as:

```scala
def toService(implicit val c: ReqToPathAndMethod) = route(c(req))
```

where `ReqToPathAndMethod` (name may be better) looks like:

```scala
trait ReqToPathAndMethod[Req] {
  def apply(req: Req): (Path, Method)
}
```

We can define it for `httpx.Request` so users won't see any difference. But in case if you want to implement your own rich http type: you have to provide a converter for it.

amartinsn:
I have implemented a RequestProxy for httpx Request type. It's based on the original http RequestProxy.

```scala
abstract class RequestProxy extends Request {
  def request: Request
[...]

Issue #89, opened 2014-12-09
Endpoint must extends a service


vkostyukov:
This is an open question. I personally believe that it might be a simple service.

PR #135, opened 2015-01-14
Address #134 and #89: no http dependency for Endpoint

vkostyukov:
See details #134 and #89.

@pasviegas, @benjumanji, @rodrigopr, @rpless, @BenWhitehead please review.


finch-core/src/main/scala/io/finch/Endpoint.scala
```diff
@@ -28,12 +28,13 @@ import com.twitter.finagle.httpx.path.Path
 import com.twitter.finagle.httpx.service.NotFoundService
 
 /**
- * A REST API endpoint that primary defines a ''route'' and might be converted
- * into a finagled service with ''toService'' method.
+ * A REST API endpoint that primary defines a ''route'' and might be
+ * implicitly converted into Finagle service ''Service[Req, Rep]''.
```
rodrigopr: I think it's better to describe that it requires an implicit `Route` to be converted to a Service

finch-core/src/main/scala/io/finch/package.scala
```diff
@@ -147,11 +148,36 @@ package object finch {
...
+    new Service[Req, Rep] {
+      def apply(req: Req) = f
+    }
+
+  /**
+   * A function that fetches the route information from the given request ''req''.
+   *
+   * @tparam Req the request type
+   */
+  trait Route[-Req] {
```
rodrigopr: We could use `@implicitNotFound` annotation with further explanation which why the implicit is required.

vkostyukov: Nice! I didn't know that!

finch-core/src/main/scala/io/finch/Endpoint.scala
vkostyukov: Sure!

vkostyukov:
BTW, I was quite surprised how easy it is to make such compete and big changes in the Finch's code base. Not sure who to thank: Scala or Finch itself :)

rodrigopr:
@vkostyukov, yeah hehe, also great that the api remained the same. nice job.

beside my last two comments it looks good to me.

finch-core/src/main/scala/io/finch/package.scala
```diff
@@ -139,19 +141,42 @@ package object finch {
...
+   * Allows to use an ''Endpoint'' when ''Service'' is expected. The implicit
+   * conversion is possible if there is an implicit view from ''Req'' to
+   * ''HttpRequest'' available in the scope.
+   *
+   * {{{
+   *   val e: Endpoint[HttpRequest, HttpResponse] = ???
+   *   Httpx.serve(new InetSocketAddress(8081), e)
+   * }}}
+   *
+   * @param e the
```
rodrigopr: incomplete doc here

vkostyukov:
Well, there is a new issue with `RequiredBody` and `OptionalBody` since they depends on two type parameters we can't really use them as previously: `RequiredBody[Json](req)`. I added two tests which are failed now. Hope I will come with something that will allow to keep the API unchanged. Travis CI should show the error message. For now it's:

```
error] /Users/vkostyukov/workspace/finch/finch-core/src/test/scala/io/finch/request/BodySpec.scala:101: type mismatch;
[error]  found   : io.finch.HttpRequest
[error]     (which expands to)  com.twitter.finagle.httpx.Request
[error]  required: io.finch.request.DecodeRequest[Int]
[error]     val i: Future[Int] = RequiredBody[Int](req)
[error]                                            ^
[error] /Users/vkostyukov/workspace/finch/finch-core/src/test/scala/io/finch/request/BodySpec.scala:101: type mismatch;
[error]  found   : io.finch.request.RequestReader[Int]
[error]  required: com.twitter.util.Future[Int]
[error]     val i: Future[Int] = RequiredBody[Int](req)
[error]                                           ^
```

@rodrigopr you have a better understanding of implicits then me. Do you have any idea what we can do here?

vkostyukov:
There is a workaround: just use `RequiredBody(req)` instead of `RequiredBody[Json](req)`. I personally liked the original version much more since it indicates which type you're reading from the request.

rodrigopr:
Yeah, I prefer the original version (`RequiredBody[Json](req)`) too.
Thinking about it, we could keep the code before the last change [1], as the user will have an implicit view `Req => HttpRequest` in scope anyway, it shouldn't be a problem.

[1] `def apply[A](req: HttpRequest)(implicit decode: DecodeRequest[A])`

vkostyukov:
Right. I made the changes and added `TODO`. This will definitely work correct, since there is an implicit conversion available. It's just a bit confusing, but, for now it's a good solution.

vkostyukov:
Last call @rpless, @benjumanji, @BenWhitehead, @pasviegas. Please review. I'm pretty happy with the state of this PR.

BenWhitehead:
@vkostyukov Looks good to me, nice job on keeping the Api they while updating the type bounds.

rpless:
:+1: for maintaining the API.

vkostyukov:
There is just one thing I don't really like in the whole story: I added `TODO` for it. For now I don't really have a solution for that issue but I hope will come up with something later.

PR #139, opened 2015-01-16
Move to new build.sbt format and conventions; update Scala and other versions

travisbrown:
I've migrated the `Build.scala` configuration to the new `build.sbt` format, and have adopted some of its conventions. The most obvious is that the directory names for subprojects now do not include the `finch-` prefix, since if no path is explicitly specified, SBT uses the project variable name, and while it's possible to use hyphens in a Scala variable name, it's not really standard.

The SBT project identifiers also no longer include the prefix, so from the SBT console you'd now write e.g. `project core` instead of `project finch-core`.

I've also made a few other small changes: I've updated the Scala, ScalaTest, and Jawn versions, and I've fixed the Travis configuration to run 2.10 tests (previously only 2.11 was being tested, even for the tests that were marked "2.10.4"). I'd be happy to break these changes out into a separate PR if you'd like.


vkostyukov:
This is super cool @travisbrown! There is broken links from README file (from the list of the sub-projects) but I can fix it later.

PR #140, opened 2015-01-16
Fix subproject links

travisbrown:
Sorry I missed these the first time around. I've left the names the same, since the artifacts will have the prefix.


vkostyukov:
:+1:

PR #141, opened 2015-01-16
removed a DS_STORE that slipped in and added it to the gitignore

rpless:


vkostyukov:
Argh! Just deleted the file in the previous commit. But we still have to cleanup `gitignore`.

Issue #129, opened 2015-01-09
Use crazy scalac flags


vkostyukov:
See details here: http://tpolecat.github.io/2014/04/11/scalac-flags.html. We probably might want to use those flags.

rpless:
I've started adding flags to the compiler and have run up against some interesting warnings:
- `Xlint` - "it is not recommended to define classes/objects inside of package objects" -- This happens all over the project
- `xLint` - "Selecting value Db from object Main, which extends scala.DelayedInit, is likely to yield an uninitialized value" -- -Xlint causing trouble again
- `Ywarn-value-discard` -- This actually raises a bunch of warnings in our tests, especially around calling `intercept`.

The first warning suggests a pretty massive restructuring of Finch might need to happen. Basically, we can't define classes/traits inside package objects.
For the second warning, I probably just need to fiddle with where DB is defined (will probably put it in its own `object`).
The third warning is more annoying than anything else because we are intentionally discarding those values.

Additionally, I'd like to `-Xfatal-warnings` to make warnings to be fatal, but cannot at the moment because of #86.

Thoughts on these (especially issue 1, 3, and using `-Xfatal-warnings`)?

vkostyukov:
Nice job @rpless on this comprehensive analysis!

I think we can postpone enabling `Xlint` flag for now (I have some plans on how to reduce the amount of code in `package.scala` classes, but this is not urgent at all). We can definitely enable `xLint` and `-Xfatal-warnings` (once we fix #86). I'm not sure I understood `Ywarn-value-discard` flag semantic. Do we really use `intercept`? I was thinking I replaced them all with `an [Exception] should be thrownBy`. But it's probably the same as `intercept`.

rpless:
We still have a couple of `intercept`s hanging around and its also happening on `an [Exception] should be thrownBy`.

vkostyukov:
Done by @rpless.

PR #142, opened 2015-01-16
Scalac flags

rpless:
For #129. Doesn't include `Xlint` and "Xfatal-warnings" for reasons discussed in the issue.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -13,7 +13,20 @@ val baseSettings = Defaults.defaultSettings ++ Seq(
     "com.twitter" %% "finagle-httpx" % "6.24.0",
     "org.scalatest" %% "scalatest" % "2.2.3" % "test"
   ),
-  scalacOptions ++= Seq( "-unchecked", "-deprecation", "-feature"),
+  scalacOptions ++= Seq(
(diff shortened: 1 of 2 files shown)


vkostyukov:
:+1: Thanks Ryan @rpless!

vkostyukov:
Looks like it's failed to compile for Scala 2.10.4. Ryan @rpless could you please take a look?

vkostyukov:
Hm. Interesting. Looks like there no flag `[error] bad option: '-Ywarn-unused-import'` for Scala 2.10.

travisbrown:
You can add the flag conditionally using `CrossVersion.partialVersion` (see e.g. [Finagle's current configuration](https://github.com/twitter/finagle/blob/1e06db17ca2de4b85209dd2fbc18e635815e994b/project/Build.scala#L17) for Jackson versions).

rpless:
Just saw this. It looks like that flag is only available in 2.11. I'll update with @travisbrown's suggestion.

travisbrown:
:+1: from me.

vkostyukov:
:+1: Superb!

Issue #133, opened 2015-01-13
Add support for the Jackson library


vkostyukov:
See details here: https://github.com/FasterXML/jackson-module-scala

jeremyoday:
Any schedule estimate on this?  I spent some time working a solution with Finch 0.4.0 but ran into type erasure problems (sample included, DefaultObjectMapper is just a pre-configured mapper).

```scala
  implicit def decodeJacksonJson[A: Manifest] = new DecodeRequest[A] {
    def apply(json: String): Option[A] = {
      try {
        Some(DefaultObjectMapper.readValue[A](json))
      } catch {
        case ex: Throwable => {
          ex.printStackTrace()
          None
        }
      }
    }
  }
```
Any help would be appreciated as we are currently skipping the RequiredBody path to parse directly with Jackson.  Let me know if my Scala is off as I am coming over from Java.  Thank you, we are enjoying our work with Finch for our new platform.

vkostyukov:
Thanks @jeremyoday! Nice too know that we're heading the right direction. I've moved this ticket to this milestone. I expect the 0.4.0 release in a couple of weeks. Not too much work is left.

I'm not an expert in Jackson, but the I believe you have use [ClassTag](http://www.scala-lang.org/files/archive/nightly/docs/library/index.html#scala.reflect.ClassTag) which contains erased information. Both `Manifest` and `ClassManifest` are going to be deprecated soon. The solution would be close to:

```scala
  import scala.reflect.ClassTag
  implicit def decodeJacksonJson[A: ClassTag] = new DecodeRequest[A] {
    def apply(json: String): Option[A] = {
      val classTag = implicitly[A]
      try {
        Some(DefaultObjectMapper.readValue[A](json, classTag.runtimeClass).asInstanceOf[A])
      } catch {
        case ex: Throwable => {
          ex.printStackTrace()
          None
        }
      }
    }
  }
```
I haven't compile it, but this should show you the direction. Let me know how it would work. And feel free to contribute your implementation!

jeremyoday:
Thank you for the note, really appreciate it.  Unfortunately that won't compile because classTag is just a type.  I played with it a bit but every permutation seems to yield scala.Nothing as the type.

This where I left off but the type seems to always be gone at runtime so was wondering if something further up in the Finch stack needed to altered to pass the type through:
```scala
  implicit def decodeJacksonJson[A: TypeTag : ClassTag] = new DecodeRequest[A] {
[...]

PR #143, opened 2015-01-17
Add Jackson support #133

vkostyukov:
I implemented the basic Jackson support. Everything works almost fine except using Jackson via `RequestReader` (but I've found suitable [workaround](https://github.com/finagle/finch/pull/143/files#diff-5c501b7d1666db0ef25e62bb7109fc75R51)): `implicit val codec = decodeJackson[Foo]`. We don't need to do the opposite thing for encoding sine it works just fine. 

Any thoughts no how to avoid declaring the `decodeJackson` explicitly?

### DIFF
--- a/jackson/src/main/scala/io/finch/jackson/package.scala
+++ b/jackson/src/main/scala/io/finch/jackson/package.scala
@@ -1,0 +1,44 @@
+/*
+ * Copyright 2014, by Vladimir Kostyukov and Contributors.
+ *
+ * This file is a part of a Finch library that may be found at
+ *
+ *      https://github.com/finagle/finch
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * You may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * Contributor(s): -
+ */
+
+package io.finch
+
+import com.fasterxml.jackson.databind.ObjectMapper
+import io.finch.request.DecodeAnyRequest
+import io.finch.response.EncodeAnyResponse
+
+import scala.reflect.ClassTag
+
+package object jackson {
+
+  implicit def decodeJackson(implicit mapper: ObjectMapper) = new DecodeAnyRequest {
+    def apply[A: ClassTag](s: String): Option[A] = try {
+      val clazz = implicitly[ClassTag[A]].runtimeClass.asInstanceOf[Class[A]]
+      Some(mapper.readValue[A](s, clazz))
+    } catch { case _: Exception => None }
+  }
+
+  implicit def encodeJackson(implicit mapper: ObjectMapper) = new EncodeAnyResponse {
+    def apply[A](rep: A): String = mapper.writeValueAsString(rep)
+    def contentType = "application/json"
+  }
+}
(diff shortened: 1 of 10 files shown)


jackson/src/main/scala/io/finch/jackson/package.scala
```diff
@@ -0,0 +1,44 @@
...
+import io.finch.response.EncodeResponse
+
+import scala.reflect.ClassTag
+
+package object jackson {
+
+  implicit def decodeJackson[A](implicit mapper: ObjectMapper, tag: ClassTag[A]) = new DecodeRequest[A] {
+    def apply(s: String): Option[A] = try {
+      val clazz = tag.runtimeClass.asInstanceOf[Class[A]]
+      Some(mapper.readValue[A](s, clazz))
```
rodrigopr: Check if it work for decode something that requires type erasure (`List[String]`)? 
ScalaObjectMapper uses a different method (`def readValue[T: Manifest](content: String): T`): 
https://github.com/FasterXML/jackson-module-scala/blob/master/src/main/scala/com/fasterxml/jackson/module/scala/experimental/ScalaObjectMapper.scala#L183

vkostyukov: Right. I will add a test.

vkostyukov:
@rodrigopr please have a look at the updated PR. It has a bit of code duplication but works as charm. I used magnet patter for `ResponseBuilder.apply(m: Magnet)`. But I decided to use some different approach in other places the idea is to have a fake class that does nothing rather then holding type `A` (see `TurnIntHttp`, `RequiredBody`, `OptionalBody`). In addition we defined a bunch of implicit conversations fro this fake class to `RequestReader` (if there is an implicit decoder). This allowed to keep the API unchanged.

vkostyukov:
And btw I managed to break a compiler while was experimenting with this approach:

```
java.lang.NullPointerException
	at scala.tools.nsc.typechecker.Implicits$ImplicitSearch.loop$1(Implicits.scala:554)
	at scala.tools.nsc.typechecker.Implicits$ImplicitSearch.checkCompatibility(Implicits.scala:556)
	at scala.tools.nsc.typechecker.Implicits$ImplicitSearch.normSubType(Implicits.scala:370)
	at scala.tools.nsc.typechecker.Implicits$ImplicitSearch.matchesPt(Implicits.scala:474)
	at scala.tools.nsc.typechecker.Implicits$ImplicitSearch.typedImplicit1(Implicits.scala:669)
	at scala.tools.nsc.typechecker.Implicits$ImplicitSearch.typedImplicit0(Implicits.scala:582)
	at scala.tools.nsc.typechecker.Implicits$ImplicitSearch.scala$tools$nsc$typechecker$Implicits$ImplicitSearch$$typedImplicit(Implicits.scala:451)
```

Trying to make a reproducer right now.

vkostyukov:
All right, I should've test it as `sbt +test` not `sbt test`. Still doesn't work for Scala 2.10. Argh!

vkostyukov:
It's green! Yay!

I need a drink.

rodrigopr:
Hehe, great job. Overall looks good. 
Just one thing bothered me, I think `DecodeMagnet` implementations are unnecessary tied to `OptionalStringBody`.

To change that, we could decouple DecodeMagnet from the reader, as:
```scala
trait DecodeMagnet[A] {
  def apply(json: String): Option[A]
}
```

And compose it with the reader when needed:
```scala
object OptionalBody {
  def apply[A](implicit decode: DecodeMagnet[A]): RequestReader[Option[A]] = OptionalStringBody map decode
}
```

What do you think?

vkostyukov:
Thanks @rodrigopr! What is the difference between `DecodeMagnet` and `DecodeRequest`? I was firstly trying to implicitly convert `DecodeAnyRequest` to `DecodeRequest` it didn't really work.

I personally don't think it's a problem to have a dependency to `OptionalStringBody` from `OptionalBody`.  We just compose readers together to get more richer readers. That's how we implemented at least half of the readers.

rodrigopr:
`OptionalStringBody` dependency in `OptionalBody` is good.
What I meant was that `DecodeMagnet` is actually the `OptionalBody` implementation, If we could decouple the decode logic from where the data comes from would be better.

Yeah, in my example `DecodeMagnet` and `DecodeRequest` were identical, only change was the variance in `A`.
Another option is to change `DecodeMagnet` to: `def apply(): DecodeRequest[A]`.

Or we could rename to `DecodeBodyMagnet` and keep as it is.

vkostyukov:
@rodrigopr thanks for the feedback! I made the changes to make `DecodeMagnet` return `DecodeRequest`. This simplified the implementations a bit! What a nice idea!

rodrigopr:
Nice, looks good to me.

vkostyukov:
Cool! I'm merging this.

PR #144, opened 2015-01-19
Fix build by removing deprecated default configuration

travisbrown:
The artifact name problem was my fault—when I ported from `Build.scala` to `build.sbt` I forgot to remove the deprecated `Defaults.defaultSettings`. This meant that if `moduleName` was set before `baseSettings`, it got ignored. Removing the old-style default configuration fixes the build (artifact names include the prefix) and is more idiomatic than changing the project identifiers.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -25,7 +25,7 @@ lazy val compilerOptions = scalacOptions ++= Seq(
   case _ => Seq.empty
 })
 
-val baseSettings = Defaults.defaultSettings ++ Seq(
+val baseSettings = Seq(
   libraryDependencies ++= Seq(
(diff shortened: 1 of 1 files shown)


travisbrown:
The build failures (403s for a Cobertura DTD while running `coveralls`) look unrelated to this change, but it's weird that they don't happen for Oracle's JDK 7. May be worth waiting an hour or two to see if they clear up.

vkostyukov:
Cool! I restarted the Ci build.

travisbrown:
Well that's a little better. Just restarted again.

vkostyukov:
Thanks! Merging this.

Issue #146, opened 2015-01-22
Route Combinators


vkostyukov:
I was thinking quite a long time on combinators for `Route` abstraction. The idea is to replace the partial function `route` in  `Endpoint` by special type `Route`, whose instance may be build to plenty of predefined combinators. Thus the particular `Endpoint` is actually something like this:

```scala
type Endpoint[Req, Rep] = (Route[R], Service[Req, Rep])
```

So, we still will be able to compose a bunch of endpoints by `orElse` operator.

The `Route[R]` abstraction is tactually a function `String => Option[R], `. It takes a path and fetches some `R` from it or `None` if the given path doesn't correspond the given `Route`. Thus we some how should be able to flatMap (i.e., using the `~>` operator) the `Route` result to underlying service. Like this:

```scala
def service(i: Int): Service[A, B]
val r: Route[Int] = ??? // route that fetches Int value
val endpoint1: Endpoint[A, B] = r ~>  service(_)
val endpoint2: Endpoint[A, B] = r ~> { i => service(i) }
```

We might have plenty of predefined combinators like `Route[Int]`, `Route[String]`, ect and use them build an entire route. Considering that `int()` and `string()` are combinators the API might looks as follows:

```scala
val r: Route[Int, String] = Get / "v" / int("version") / "users" / string("name")
```

Now we can define an `Endpoint`:
```scala
val e: Endpoint[A, B] = r ~> { (version, name) => service(version, name)}
```

This approach solves a lot of issues:
1. We no longer need a type alias for route: #114 
2. We may decouple `HttpRequest` from `Route`
3. We would be able to share the sub-route with other routes via composition
4. We finally can define a proper reverse routes
5. We can fetch the path of concrete route which was asked recently on Gitter.

I believe this feature, as well as `RequestReader` will be cornerstones of Finch 1.0.0.

Let me know what do you think @pasviegas, @benjumanji, @rpless, @rodrigopr.

rpless:
I like the direction. One of the potential benefits is that we'll be able to inspect routing and users could extend it with meta data (as desired in #73). The other issue with routing currently is we can't really define it independently of the service it maps to. I had a lot of issues with this when I was experimenting with #79. Right now all of our routing is built on a set of extractors that are defined in Finagle. Its fine for simple use cases, but becomes a problem when you want more complicated things (i.e. composition and path variables).

The only thing I would suggest at the moment is a shorthand for type combinators in addition to longer form suggested above. Something like:
```scala
[...]

Issue #145, opened 2015-01-21
Applicative RequestReader


benjumanji:
I think it would be useful to provide an applicative interface to request readers, as this would allow for validation that collected all errors rather than just the first one.

benjumanji:
I am on holiday at the moment, but would be happy to have a go at making this happen when I get back. The obvious question here, is what this should look like. Given that applicative functions aren't in scala's predef, and having scalaz-core as a dependency feels a little heavy, I guess we could just have a function for chaining request readers together that return an aggregate exception?

vkostyukov:
Hi @benjumanji! I was trying to avoid the scalaz dependency as long as possible. The only thing we needed so far is `Reader` which is a `RequestReader`. It would be nice to avoid this dependency in this time also. Let's wait for the third time we need scalaz. It might be a good sign to add it.

I agree, we should provide a way to collect all the exceptions thrown by `RequestReader`. This will allow for example to verify all the fields from the request and immediately mark them red in the UI. While now, we can mark them one-by-one, which is kinda annoying.

I would love to see a PR from you with this important change. Do you have any particular sketch in mind on how would the API look like?

vkostyukov:
BTW, there is a light-weight version of scalaz that appears like 1 hour ago: https://github.com/mpilquist/fundamentum. The project is still in progress, but I think it's worth to look at it.

jenshalm:
Note that this would very likely also require to change the DecodeRequest API. It currently just returns a None in case of parsing/validation error, so even if we combine the parameter and header errors, the error handling would still be incomplete if we cannot report body parser errors more precisely than with a generic "parsing JSON body failed".

vkostyukov:
Thanks for the feedback @jenshalm! You're right, we might consider redesigning the `DecodeRequest` to return `Either` or `Try` instead of `Option`.

travisbrown:
Just wondering—how would people feel about an Algebird dependency? You'd probably still want to add some syntax on top of its `Applicative`, and it's not a super light dependency, but it'd give you this abstraction (and others—in fact it includes a reader monad implementation, although I don't know off the top of my head how easy it'd be to drop in here) in a Twitter-friendly form.

"That's a terrible idea and not in line with the goals of this project" is a completely reasonable answer, of course.

vkostyukov:
[...]

Issue #114, opened 2014-12-14
Type alias for endpoint routes


pasviegas:
So, I just started with finch, and I am really happy with the project.

Just as a beautifier, what do you think about having a type alias like the following?

```scala
type Routes[Req <: HttpRequest, Rep] = PartialFunction[(Method, Path), Service[Req, Rep]]

object GreetingEndpoint extends Endpoint[HttpRequest, Greeting] {
  def route: Routes[HttpRequest, Greeting] = {
    case Method.Get -> Root / "hello" / name => HelloService(name)
    case Method.Get -> Root / "hurray" / name => HurrayService(name)
  }
}
```

vkostyukov:
@pasviegas do you mind if I move this to 0.5.0 release. Since we're trying to hide the dependency on `httpx` we might want to have our own abstraction for route in future. So it would not be an _alias_ but a standalone abstraction. Maybe.

vkostyukov:
Alright, moving it to the next release.

vkostyukov:
I think this is no longer actual due to #146. @pasviegas do you think I can close it?

vkostyukov:
I think we can close it for now, since the route combinators approach is preferred.

pasviegas:
yes, we should close it :)

Issue #79, opened 2014-11-22
Brand new HTTP Routing DSL


vkostyukov:
We still should be as close to the Finagle API as possible. But having a nicer routing DSL is a really cool feature. I think we can reuse a lot from default routing DSL but just make is a bit shorter. Like this:

```scala
val e = new Endpoint[Req, Rep] {
  def route = {
    case Get / "users" / id => ???
    case Post / "users" => ???
  }
}
```

Instead of:

```scala
val e = new Endpoint[Req, Rep] {
  def route = {
    case Method.Get -> Root / "users" / id => ???
    case Method.Post -> Root / "users" => ???
  }
}
```

rpless:
I don't actually think we need to anything more than import `import com.twitter.finagle.httpx.Method._`.
All of the Methods are exposed from the message object: https://github.com/twitter/finagle/blob/master/finagle-httpx/src/main/scala/com/twitter/finagle/httpx/Method.scala

vkostyukov:
I was wondering if we can get rid of `-> Root /` in the path. To use it like `case Get / "users" / id => ???`. Ryan @rpless, do you think this is possible to do?

rpless:
I had missed that part. I think it might be. We'll probably have to extend the `Path` object to include a `Method`. I'll experiment with it a little and let you know what I find.

vkostyukov:
No longer actual because #146.

PR #147, opened 2015-01-24
Intoroduce io.finch.route package

vkostyukov:
This is an initial implementation of route combinators as discussed in #146. There is no DSL layer yet - just raw classes. So, for now we have to explicitly create new instances like `StringMatcher` and `StringExtractor`. The next step would be to write a bunch of implicit conversion in order to enable full-featured DSL.

But now I wanted to collect a feedback and ask for an advice. I don't really think we need both `RouteMatcher` and `RouteExtractor`. My primary goal now is to replace them both with just `Router`. I would love to hear your advices @pasviegas, @benjumanji, @rodrigopr and @rpless on how can we achieve it.

The main problem here is that there are routers that don't extract anything but match something: `Get`, `StringMatcher`. So, the problem is in `RouteExtractor.flatMap` method. Let's say we have just `Router` and define `Get` as `Router[Unit]`. And when we flatMap it to `Router[Int]` we will get `Router[Unit / Int]`, which is wrong.

Anyway. The main idea i was trying to show is that router is just a function `Route => A`. And if `A` is `Service[Req, Rep]` we can call such router an `Endpoint`, since it takes a route and returns a service. This is exactly what endpoint is.

### DIFF
--- a/core/src/test/scala/io/finch/route/RouterSpec.scala
+++ b/core/src/test/scala/io/finch/route/RouterSpec.scala
@@ -1,0 +1,184 @@
+/*
+ * Copyright 2015, by Vladimir Kostyukov and Contributors.
+ *
+ * This file is a part of a Finch library that may be found at
+ *
+ *      https://github.com/finagle/finch
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * You may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * Contributor(s):
+ */
+
+package io.finch.route
+
+import com.twitter.util.Await
+
+import io.finch._
+import io.finch.response._
+
+import com.twitter.finagle.httpx
+import com.twitter.finagle.httpx.Method
+import com.twitter.finagle.Service
+import org.scalatest.{Matchers, FlatSpec}
+
+class RouterSpec extends FlatSpec with Matchers {
+
+  val route = List(
+    MethodToken(Method.Get), PathToken("a"), PathToken("1"), PathToken("b"), PathToken("2")
+  )
+
+  "A Router" should "extract single string" in {
+    val r = Get / string
+    r(route) shouldBe Some((route.drop(2), "a"))
+  }
+
+  it should "extract multiple strings" in {
+    val r = Get / string / "1" / string
+    r(route) shouldBe Some((route.drop(4), /("a", "b")))
+  }
+
+  it should "match method" in {
+    Get(route) shouldBe Some(route.tail)
+  }
+
+  it should "match method only once" in {
+    val r = Get / Get
+    r(route) shouldBe None
+  }
+
+  it should "match string" in {
+    val r = Get / "a"
+    r(route) shouldBe Some(route.drop(2))
+  }
+
+  it should "match 2 or more strings" in {
+    val r = Get / "a" / 1 / "b"
+    r(route) shouldBe Some(route.drop(4))
+  }
+
+  it should "not match if one of the routers failed" in {
+    val r = Get / "a" / 2
+    r(route) shouldBe None
+  }
+
+  it should "not match if the method is missing" in {
+    val r = "a" / "b"
+    r(route) shouldBe None
+  }
+
+  it should "be able to skip route tokens" in {
+    val r = * / "a"
+    r(route) shouldBe Some(route.drop(2))
+  }
+
+  it should "match either one or other matcher" in {
+    val r = (Get | Post) / ("a" | "b")
+    r(route) shouldBe Some(route.drop(2))
+  }
+
+  it should "match int" in {
+    val r = Get / "a" / 1
+    r(route) shouldBe Some(route.drop(3))
+  }
+
+  it should "be able to not match int if it's a different value" in {
+    val r = Get / "a" / 2
+    r(route) shouldBe None
+  }
+
+  it should "be able to skip one route tokens" in {
+    val r = Get / *
+    r (route) shouldBe Some(route.drop(2))
+  }
+
+  it should "be able to match the whole route" in {
+    val r1 = Get / * / $
+    val r2 = Get / * / * / * / * / $
+    r1(route) shouldBe None
+    r2(route) shouldBe Some(Nil)
+  }
+
+  it should "support DSL for string and int extractors" in {
+    val r1 = Get / "a" / int / string
+    val r2 = Get / "a" / int("1") / "b" / int("2") / $
+
+    r1(route) shouldBe Some((route.drop(4), /(1, "b")))
+    r2(route) shouldBe Some((Nil, /(1, 2)))
+  }
+
+  it should "support DSL for boolean marchers and extractors" in {
+    val route = List(PathToken("flag"), PathToken("true"))
+    val r1 = "flag" / boolean
+    val r2 = "flag" / true
+
+    r1(route) shouldBe Some((Nil, true))
+    r2(route) shouldBe Some(Nil)
+  }
+
+  it should "be implicitly converted into a service" in {
+    def echo(s: String) = new Service[HttpRequest, String] {
+      def apply(request: HttpRequest) = s.toFuture
+    }
+
+    val service: Service[HttpRequest, String] = Get / string /> echo
+    Await.result(service(httpx.Request("/foo"))) shouldBe "foo"
+  }
+
+  it should "be composable as an endpoint" in {
+    val r1 = Get / "a" / int /> { _ + 10 }
+    val r2 = Get / "b" / int / int /> { case a / b => a + b }
+    val r3 = r1 | r2
+
+    r3(route) shouldBe Some((route.drop(3), 11))
+  }
+
+  it should "maps to value" in {
+    val r = Get /> 10
+    r(route) shouldBe Some((route.tail, 10))
+  }
+
+  it should "converts into a string" in {
+    val r1 = Get / $
+    val r2 = Get / "a" / true / 1
+    val r3 = Get / ("a" | "b") / int / long / string
+    val r4 = Get / string("name") / int("id") / boolean("flag") / "foo"
+    val r5 = Post / * / $
+
+    r1.toString shouldBe "GET/"
+    r2.toString shouldBe "GET/a/true/1"
+    r3.toString shouldBe "GET/(a|b)/:int/:long/:string"
+    r4.toString shouldBe "GET/:name/:id/:flag/foo"
+    r5.toString shouldBe "POST/*/"
+  }
+
+  it should "support the for-comprehension syntax" in {
+    val r1 = for { a / b <- Get / string / int } yield a + b
+    val r2 = for { a / b / c <- Get / "a" / int / string / int } yield b + c + a
+    val r3 = r1 | r2
+
+    r1(route) shouldBe Some((route.drop(3), "a1"))
+    r2(route) shouldBe Some((Nil, "b21"))
+    r3(route) shouldBe r1(route)
+  }
+
+  it should "be implicitly convertible into service from future" in {
+    val s: Service[HttpRequest, HttpResponse] =
+      (Get / "foo" /> Ok("bar").toFuture) |
+      (Get / "bar" /> Ok("foo").toFuture)
+
+    Await.result(s(httpx.Request("/foo"))).contentString shouldBe "bar"
+    Await.result(s(httpx.Request("/bar"))).contentString shouldBe "foo"
+    a [RouteNotFound] should be thrownBy Await.result(s(httpx.Request("/baz")))
+  }
+}
--- a/core/src/main/scala/io/finch/route/package.scala
+++ b/core/src/main/scala/io/finch/route/package.scala
@@ -1,0 +1,386 @@
+/*
+ * Copyright 2015, by Vladimir Kostyukov and Contributors.
+ *
+ * This file is a part of a Finch library that may be found at
+ *
+ *      https://github.com/finagle/finch
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * You may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * Contributor(s): -
+ */
+
+package io.finch
+
+import com.twitter.finagle.Service
+import com.twitter.finagle.httpx.Method
+import com.twitter.util.Future
+
+/**
+ * This package contains various of functions and types that enable _router
+ * combinators_ in Finch. A Finch [[route.Router]] is an abstraction that
+ * is responsible for routing the HTTP requests using their method and path
+ * information. There are two types of routers in Finch: [[route.Router0]]
+ * and [[route.RouterN]]. `Router0` matches the route and returns an `Option`
+ * of the rest of the route. `RouterN[A]` (or just `Router[A]`) in addition
+ * to the `Router0` behaviour extracts a value of type `A` from the route.
+ *
+ * A [[route.Router]] that extracts [[Service]] is called an endpoint. An
+ * endpoint `Req => Rep` might be implicitly converted into `Service[Req, Rep]`.
+ * Thus, the following example is a valid Finch code:
+ *
+ * {{{
+ *   def hello(s: String) = new Service[HttRequest, HttpResponse] {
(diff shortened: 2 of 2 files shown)


vkostyukov:
If it's turned to be impossible we may rename things:
* `RouteExtractor[A]` -> `RouterN[A]`
* `RouteMatcher` -> `Router0`

And then create an aliases: 
```scala
type Router[A, B] = RouterN[A, B]
type Endpoint[A, B] = Router[Service[A, B]]
```
Usually we don't need write types but we need we should write them in a nice way:

```scala
val r: Router[Int / Int / String] = ???
```

vkostyukov:
@jenshalm it would also be nice to get your feedback on this.

rpless:
I would recommend not using `/` as the infix case class in the Router's parameters. It might conflict with an existing class name in the Finagle route api: https://github.com/twitter/finagle/blob/master/finagle-httpx/src/main/scala/com/twitter/finagle/httpx/path/Path.scala#L92.

vkostyukov:
Right. But I think that with route combinators users no longer need to import `finagle.httpx.path._` class. Why I do like `/` as the infix case class is that is looks like an HTTP path delimiter, which makes it fitting in the DSL thing really well.

vkostyukov:
What I also plan to do here is:

* `Long` extractor
* `Router.toString` should return a string representation of the router. Something like `GET "/users/:user"` (this has been asked recently on Gitter)
* More tests
* A DSL layer - please, go ahead and leave a comment on how do see the DSL for route combinators.

jenshalm:
Looks like a good start. But it's hard to judge the feature before seeing the planned DSL, as I assume that the base work available now is more intended to be internal API (if it would be public API I'd find it a bit too verbose).

Other than that two things come to mind:

* Are there plans to let the routing layer figure out 405 errors? In this case it might require to offer nesting method declarations inside path matchers

* A little thing: Like in some other places in Finch there is the change to reduce boilerplate, e.g. through creating a base extractor like this:

```
  class Extractor[A](f: String => Option[A]) extends RouterN[Int] {
    def apply(route: Route): Option[(Route, Int)] = for {
      (r, ii) <- StringExtractor(i)(route)
      iii <- f(ii)
    } yield (r, iii)
  }
```

which would then allow concise definitions like this:

```  
  case class IntExtractor(i: String) extends Extractor(stringToSomeNumber(_.toInt))
```

(this would require to change the stringToSomeNumber signature accordingly)

vkostyukov:
Thanks for the feedback @jenshalm! You're right. This is an internal API, which will be covered by a DSL layer. Wha I managed to get so far is implicit conversions for string/int matchers. So, we can already write something like this:

```scala
val router = Get / "users" / 100
``

Regarding to the universal extractor. I tried to implement the similar approach to `StringToValueOrFail` from request reader abstraction. So, the [`StringToValue`](https://github.com/finagle/finch/pull/147/files#diff-013dead8f7e981221fbdc102176fa21dR252) router does nothing that tries to evaluate the given conversion. This approach worked well for request readers, so I decided just to reuse this abstraction in some way.

Regarding the `405` error. This is an interesting question. What you can do for now (having the route combinators) is have a router that represents the rest of the path. Lets say you have an endpoint: `? /users/:id`, where `?` is either `POST` or `PATH`. So, we want to response 405 in case of other methods. Note, that we still don't have a DSL for extractors only for matchers.

```scala
val rest = "users" / IntExtractor("id")
val r = ((Path | Post) / rest ~> goodService()) | (SkipOne / rest ~> respond405())
```

vkostyukov:
Added DSL for `SkipOne` and replaced `SkipAll` with the end of the route `$`. See [here](https://github.com/finagle/finch/pull/147/files#diff-15fab5bf685af131056294bdbbcd0950R103) for usage details.

jenshalm:
I'm afraid we are talking at cross purposes regarding the generic extractor. I am not suggesting any changes for the API of the existing extractors, but only two things: a) avoid copy-paste coding as found in the RequestReaders and Extractors (once you add a LongExtractor similar to the IntExtractor) and b) as a side-effect allow users to define custom RequestReaders/Extractors in a concise API. What you wrote somehow does not sound like a counter-argument to my suggestion.

vkostyukov:
I think this is arguable that users will end up writing their own extractors. There is not to much information  is usually passed via URI. 

And I don't really think users write their own request readers. The `RequestReader` abstraction allows to compose readers together using the monadic API and get any readers we might want, which is a functional way to compose things but not inherit them. But I totally agree with you about the boilerplate code that exists in both packages. I'm just saying that  if user want to extend `RequestReader` to write hist own one, he is probably doing something wrong.

In case of `Router` I decide to add a `maybeMap` function into `RouterN`, which looks as follows:
```scala
    def maybeMap[B](fn: A => Option[B]): RouterN[B] = new RouterN[B] {
      def apply(route: Route): Option[(Route, B)] = for {
        (r, a) <- self(route)
        b <- fn(a)
      } yield (r, b)
    }
```

So, we can define `Extractor` and `IntExtractor` as follows:

```scala
  class Extractor[A](name: String, f: String => Option[A]) extends RouterN[A] {
    def apply(route: Route): Option[(Route, A)] = StringExtractor(name).maybeMap(f)(route)
  }
  case class IntExtractor(name: String) extends Extractor(name, stringToSomeValue(_.toInt))
```

BenWhitehead:
Here is another scala library that provides a DSL for URI's perhaps some of the things it does could be leveraged here, or provide guidance.

https://github.com/NET-A-PORTER/scala-uri

vkostyukov:
Thanks @BenWhitehead for sharing! Meanwhile, what do you guys thing about the following DSL for extractors:

```scala
val router = Get / "users" / "id" @ int / "name" @ string
```

jenshalm:
Thanks for your explanations. I agree regarding preferring composition over inheritance. However I wasn't necessarily thinking of something that would require inheritance, but more of a lack of flexibility I see in the current request readers, but that does not belong here, so I'll stop and maybe create another issue or PR that explains what I meant in a better way.

Regarding your example for handling a 405, that would certainly work, but for something that I would usually want for all of my routes it looks like too much plumbing, still. I think handling 405 is something I'd expect to be provided by the web stack for free.

The extractor DSL looks good, just that I am not sure I would expect to be forced to provide a name for the dynamic path elements (I think many other routing DSLs do not do that). I think having something as concise as this:

```
val router = Get / "users" / int / string
```

would be neat. For rendering the route as a string, the framework could pick defaults based on the type, e.g. `:int`. But then again, it's not a big deal if that's not possible.

vkostyukov:
Thanks @jenshalm! Please, go ahead with the issue/PR about request readers. I would love to understand your concern and fix it in Finch.

Regarding the 405 error. I belie most of the frameworks/libraries will respond 404 in this case. If you know libraries that support 405 out of the box please let me know. I'm pretty sure there is a way to do that using route compositor, I'm just trying to understand whether or not it worths doing this right now. I mean, we are going to cover 99% use cases of most of the users and there is also 1% that requires more changes. Alright. Let me think about this. I will try to understand the size of changes to support 405 error.

We can definitely just have `string`, `int` etc. I was just thinking about string representations. We need this for a different reasons. First of all there is a ticket #73 that will require it. And also a Finch user was asked this on Gitter recently. What I'm planing to do is to support both ways to write routers: with custom name and with default name:

```scala
val r1 = Get / "users" / int
val r2 = Get / "users" / int("id")
val r3 = Get / "users" / "id" @ int
```

The question here is what looks better `r2` or `r3`? It's definitely simpler to implement `r2`.

vkostyukov:
I think this 405 error handling is almost impossible to do using pure route combinators, since it will require to add a new _special_ type of combinator that handles method and when we need to support any other error we will end up adding new type/class of combinators. The reason for this is that we can't really replace `Option` with `Either` or `Try` to propagate the exception since this is not what router do and in case of router composition with `orElse` it's not really clear how to handle the exception. 

What I think we can do here is to use Finagle filters + Finch route combinators. I believe, this would be the most right way to do it. Let's say you have a filer `MakeSureMethodIs(methods: Seq[Method])` that makes sure that given request has an acceptable method and throws/responds 405 otherwise. So we can easily use it like this:

```scala
val r = * / "users" / int ~> MakeSureMethodIs(Seq(Post, Patch)) andThen editUser(_)
```

I think this is a reasonable boilerplate and more importantly we use the strong sides of both libraries: Finch and Finagle. What do you think @jenshalm? I was trying to google any frameworks that support it but found only Django that does it via decorators written by user, which is quite similar to the filters approach.

vkostyukov:
Enabled DSL layer for everything and rewrote all the tests using DSL:

```scala
val r1 = Get / "a" / int / string
val r2 = Get / "a" / "id" @@ int / "b" / "2" / $
```

jenshalm:
I think r2 and r3 work equally well in terms of being intuitive, r2 maybe just feels a tiny bit more aligned with r1?

Regarding 405 I agree in terms of design consequences potentially being severe. For that reason it is probably not amongst the most critical features for now. Regarding frameworks that support it out of the box, I think spray does (but I'm not 100% certain, this is more from reading their docs than from verifying it myself).

Regarding the PR, I'll wait a bit before I do the one for RequestReader flexibility, as first I'll be doing one for making RequestReader applicative (issue #145). I have a fully working solution here on my machine, including tests, just have to polish it a bit. (I'm currently on the same team as Ben who suggested that functionality).

vkostyukov:
@jenshalm this is super cool! Please, go ahead with that PR! I'm really looking forward to it.

And I just realised that it's not really possible to have both `int` and `int("name")` things in the same scope. So, I decided to go forward with `@@` syntax.

Regarding the 405 error. I will have a look on spray. Probably it will help me to understand how to integrate it in Finch. But for now, I think the Finagle filter is probably the best way to do that.

vkostyukov:
I had to replace `~>` operator with `/>` operator to allow us to write things w/o braces:

```scala
val r = Get / "a" /> 10 /> { _ + 20 }
```

This wont compile with `~>` since it has a higher precedence then `/`.

vkostyukov:
I'm pretty happy with the state of this PR. The only thing I wanted to investigate is the ability to use futures directly in the endpoint rather then services. Otherwise, I think this is ready to be shipped.

Also, please, let me know what do you guys think about `\>` alias for `map`.

benjumanji:
I am very wary of the operator soup that we seem to be generating. I am
traveling for the next couple of days, but I would really prefer it if we
could figure out some way to have int("a") over another operator. Surely we
can do some trick with an overloaded apply method? I won't have time to
look into it before the weekend but if this isn't merges before then I will
have a go.

vkostyukov:
Thanks @benjumanji for the feedback! I will try to work out `int("name")` then. Do you have any preferences on `/>` operator? Does it look reasonable?

pasviegas:
Maybe it would be a good idea to move the operators to a different package, like "syntax", so the "operator soup" is optional.

vkostyukov:
But route combinators w/o syntax make no sense. Nobody wants to write `Get flatMap int flatMap string`. Thus, I think we should provide a reasonable set of operators for route combinator out of the box.

vkostyukov:
If somebody has a good candidate for `map` instead of `/>`, please let me know,

vkostyukov:
I turned to be super easy to replace `@@` syntax with `int("name")`.

vkostyukov:
Added `toString` method that converts routers to string:

```scala
val r4 = Get / string("name") / int("id") / boolean("flag") / "foo"
r4.toString shouldBe "GET /:name/:id/:flag/foo"
```

rodrigopr:
Cool @vkostyukov, the new API is looking good already.

I personally prefer the clarity of partial functions when describing multiple routes,
we may be able to continue using it with this api if `Extractor` had a `unapply` method.

Along with a helper function like:
```scala
object Endpoint { 
  def build[Rep, Rep](router: PartialFunction[Router[_], Service[Req, Rep]]): Endpoint[Req, Rep]
}
```

What do you think?

vkostyukov:
@rodrigopr Looks interesting. I think we keep the old-style `Endpoint` for some time, so you can continue using it. Once we consider deprecating `Endpoint`, we might think about the approach you suggested. But I personally think route combinators are the nicest way to do that. In case of multiple routes if might looks like (using for-comprehension):

```scala
val users = for { id / name <- Get / "users" / int / string } yield GetUser(id, name) |
            for { id <- Post / "users"} yield PostUser(id) 

val backend = users | tickets | tags
```

or using braces:

```scala
val users = (Get / "users" / int / string /> { case a / b => GetUser(id, name)}) |
            (Post / "users" / int /> PostUser(_))
```

Finch will be definitely supporting the old-school `Endpoint` but the new API will be preferred.

@rodrigopr what do you think about this syntax?

BenWhitehead:
I think I'm with @rodrigopr, I think I prefer the Endpoint with partial function approach from early finch.

For example I've got an existing service written in finch that looks similar to this:
```scala
  def route = {
    case Method.Get -> Root / "item" / Integer(itemId) / "recipe" / "summary" => 
      itemRecipeSummary(itemId) ! json
    case Method.Get -> Root / "item" / Integer(itemId) / "ingredient" / "summary" => 
      itemIngredientSummary(itemId) ! json
    case Method.Get -> Root / "item" / Integer(itemId) / "ingredient" / "tree" => 
      itemIngredientTree(itemId) ! json
  }
```

I'm really not sure I like the new approach better (which in my understanding would like like the following):
```scala
  val routes = (Get / "item" / int / "recipe" / "summary" /> itemRecipeSummary(_) ! json) |
                     (Get / "item" / int / "ingredient" / "summary" /> itemIngredientSummary(_) ! json) |
                     (Get / "item" / int / "ingredient" / "tree" /> itemIngredientTree(_) ! json)
```

One of the things that originally appealed to me with finch is that the routing system just made sense with a basic understanding of scala, it wasn't complicated with a bunch of special operators that I had to learn.  Also, given the fact that routes are defined as partial functions it makes it harder (in my opinion) to mess things up.

Before I chose to use finch the first time I looked at spary and had used unfiltered on a small project.  One of the things I hated with all the fancy syntax of these projects is how easy it was to create code that had bugs in it.

Take for example the following and see if you can spot the bug, mind you this code would compile cleanly:
```scala
def routes = {
  (Get / "user" / int /> user(_) ! json) |
  (Get / "item" / int / "details" /> userDetails(_) ! json)
  (Get / "item" / int / "groups" /> userGroups(_) ! json)
}
```

The bug in the previous example is that I forgot the `|` after the second route definition.  If not paying close attention it's something that is easy to miss and could result is wasted time debugging.

Admittedly, I'm a bit of a noob when it comes to designing libraries in Scala, but the idea of being able to continue using partial functions is something that I would very much like to see stay in the core design of finch.

That said, I think it's probably a good idea to decouple things from the netty and httpx packages.

core/src/test/scala/io/finch/route/RouterSpec.scala
```diff
@@ -0,0 +1,172 @@
...
+  }
+
+  it should "converts into a string" in {
+    val r1 = Get
+    val r2 = Get / "a" / true / 1
+    val r3 = Get / ("a" | "b") / int / long / string
+    val r4 = Get / string("name") / int("id") / boolean("flag") / "foo"
+    val r5 = Post / * / $
+
+    r1.toString shouldBe "GET "
```
BenWhitehead: I don't think `GET` is a valid route for an http server, I think the minimum acceptable would be `GET /` which means we probably shouldn't be able to declare a route that consists of only a method.

vkostyukov:
Thanks for the feedback @BenWhitehead! And special thanks for the example of the bug. I also like pattern matching on routes: it's simple and easy to understand. But it also a big lack of flexibility that we can just see by the number tickets created against it (copy from #146):

1. We no longer need a type alias for route: #114 
2. We may decouple HttpRequest from Route
3. We would be able to share the sub-route with other routes via composition
4. We finally can define a proper reverse routes
5. We can fetch the path of concrete route which was asked recently on Gitter.

Unfortunately, we won't be able to resolve all of this w/o combinators.

However, I totally agree with you guys @rodrigopr and @BenWhitehead that we have to keep things simple. Finch is a super simple tool so far (I got a feedback recently from a person who wrote a Finch application with zero knowledge of Scala and Finagle, which is the best feedback that I've got so far). So, I would love to see how can we use pattern matching on combinators.

So, I would suggest the following plan:

1. We merge this for now
2. @rodrigopr or I will introduce a PR that allows to use pattern matching over combinators so we can deprecate the old-style `Endpoint` in 0.5.0 release.

What do you say?

BenWhitehead:
I think that sounds like a reasonable approach. For me, something almost important to me as simplicity is that the library helps me to write correct code and make it difficult to write incorrect code.

vkostyukov:
Thus, we should probably end up having something similar to #79, which was planned as just a fancy version of the current approach.

Regarding the `Get` route. The thing is that `Get` itself is a valid route, but it's string representation looks weird. And I don't really see a simple way to fix it now. So, I think we don't really need to have a string representation that looks exactly like in framework X, but we can try to make it looks similar, which means that we probably want to make it simple. I thing string representation of the router like `GET/:name/:id` looks pretty reasonable for now.

Regarding the `Get` router semantic. The idea is following:

```scala
val r1 = Get // match all the GET requests
val r2 = Get / $ // match GET requests with empty path, like GET /
```

You can still expose any constraints on your routes.

vkostyukov:
I thing it's ready to be merged. Last call. Let me know if have any concerns left. Especially about `/>` operator for map. Although, we can easily rename it later.

vkostyukov:
Btw, here is a valid code that enables an HTTP echo server (with routing) in Finch:

```scala
Httpx.serve(":8080", Get / "echo" / string /> Ok(_).toFuture)
```

rodrigopr:
I'll play a bit with it later this week, see if I can make the pattern-matching work with the new route combinators.
For this PR, I think it's ready to be merged.

vkostyukov:
I'm pretty sure we will figure something out. I will also have a look on it this weekend. Now I'm planing to merge this PR and publish a snapshot version.

PR #148, opened 2015-01-26
add applicative behaviour to RequestReader. Fixes #145

### LINKED ISSUE #145: Applicative RequestReader

jenshalm:
This adds applicative behaviour to the RequestReader. It is essential for comprehensive error handling as it collects all errors instead of the existing monadic behaviour which is fail-fast.

A monadic reader written with the current API:

``` scala
  val intTuple: RequestReader[(Int, Int)] = for {
    a <- RequiredIntParam("a")
    b <- RequiredIntParam("b")
  } yield (a, b)
```

can be rewritten in applicative style like this:

``` scala
  val intTupleApplicative: RequestReader[(Int, Int)] = 
    (RequiredIntParam("a") ~
     RequiredIntParam("b")) { 
       case a ~ b => (a, b)
     }
```

None of the existing APIs are affected by this addition.

The functionality has been added without introducing new dependencies like scalaz. The generic applicative builder has been kept separate from the (very little) code which is specific to RequestReader.

Some things remain open for debate and could quickly be adjusted depending on feedback:

1) I picked `~` as the combinator for chaining readers, as this is a very common symbol for this kind of combinators. However, we could also be forward-compatible and use the `|@|` symbol from scalaz, should we expect that scalaz might be introduced as a new dependency at some point anyway, which would make the applicative builder utility added with this PR obsolete.

2) Using the applicative behaviour would require an additional import `io.finch.applicative._`. If this new functionality should be available for everyone without additional imports, it could alternatively be integrated into the `io.finch.request._` import.

### DIFF
--- a/core/src/test/scala/io/finch/request/ApplicativeRequestReaderSpec.scala
+++ b/core/src/test/scala/io/finch/request/ApplicativeRequestReaderSpec.scala
@@ -1,0 +1,77 @@
+/*
+ * Copyright 2014, by Vladimir Kostyukov and Contributors.
+ *
+ * This file is a part of a Finch library that may be found at
+ *
+ *      https://github.com/finagle/finch
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * You may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * Contributor(s):
+ * Jens Halm
+ */
+package io.finch.request
+
+import org.scalatest.{FlatSpec,Matchers}
+
+import com.twitter.finagle.httpx.Request
+import com.twitter.util.{Await,Throw}
+
+class ApplicativeRequestReaderSpec extends FlatSpec with Matchers {
+
+  
+  val reader: RequestReader[(Int, Double, Int)] = 
+    (RequiredIntParam("a") ~
+     RequiredDoubleParam("b") ~
+     RequiredIntParam("c")) map { 
+       case a ~ b ~ c => (a, b, c)
+     }
+  
+  
+  "The applicative reader" should "produce three errors if all three numbers cannot be parsed" in {
+    val request = Request.apply("a"->"foo", "b"->"foo", "c"->"foo")
+    Await.result(reader(request).liftToTry) should be (Throw(RequestReaderErrors(Seq(
+      ValidationFailed("a", "should be integer"),
+      ValidationFailed("b", "should be double"),
+      ValidationFailed("c", "should be integer")
+    ))))
+  }
+  
+  it should "produce two validation errors if two numbers cannot be parsed" in {
+    val request = Request.apply("a"->"foo", "b"->"7.7", "c"->"foo")
+    Await.result(reader(request).liftToTry) should be (Throw(RequestReaderErrors(Seq(
+      ValidationFailed("a", "should be integer"),
+      ValidationFailed("c", "should be integer")
+    ))))
+  }
+  
+  it should "produce two ParamNotFound errors if two parameters are missing" in {
+    val request = Request.apply("b"->"7.7")
+    Await.result(reader(request).liftToTry) should be (Throw(RequestReaderErrors(Seq(
+      ParamNotFound("a"),
+      ParamNotFound("c")
+    ))))
+  }
+  
+  it should "produce one error if the last parameter cannot be parsed to an integer" in {
+    val request = Request.apply("a"->"9", "b"->"7.7", "c"->"foo")
+    Await.result(reader(request).liftToTry) should be (Throw(ValidationFailed("c","should be integer")))
+  }
+  
+  it should "parse all integers and doubles" in {
+    val request = Request.apply("a"->"9", "b"->"7.7", "c"->"5")
+    Await.result(reader(request)) should be ((9,7.7,5))
+  }
+    
+  
+}
--- a/core/src/main/scala/io/finch/request/package.scala
+++ b/core/src/main/scala/io/finch/request/package.scala
@@ -21,12 +21,13 @@
  * Ben Whitehead
  * Ryan Plessner
  * Pedro Viegas
+ * Jens Halm
  */
 
 package io.finch
 
 import com.twitter.finagle.httpx.Cookie
-import com.twitter.util.Future
+import com.twitter.util.{Future,Return,Throw,Try}
 
 import scala.reflect.ClassTag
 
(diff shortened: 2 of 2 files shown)


jenshalm:
The decrease in coverage is due to the tests not using all builders from 2 to 12. It would be quite a bit of stupid plumbing to test them all, but still, would be quick to do if requested.

core/src/test/scala/io/finch/request/ApplicativeRequestReaderSpec.scala
```diff
@@ -0,0 +1,79 @@
...
+import io.finch.applicative._
+
+class ApplicativeRequestReaderSpec extends FlatSpec with Matchers {
+
+  
+  val reader: RequestReader[(Int,Double,Int)] = 
+    (RequiredIntParam("a") ~
+     RequiredDoubleParam("b") ~
+     RequiredIntParam("c")) { 
+       (a,b,c) => (a,b,c)
```
vkostyukov: Do you think we can avoid definition of `CanBuildX` and let users pattern match the result instead? Like in `Router`:

```scala
val reader: RequestReader[(Int,Double,Int)] = 
  (RequiredIntParam("a") ~
   RequiredDoubleParam("b") ~
   RequiredIntParam("c")) { case a ~ b ~ c => (a,b,c) }
```

I think this is a reasonable syntax.

vkostyukov:
Looks impressive! Awesome job @jenshalm! I will have a detailed look in a couple of hours, once I get in the office. For now, I have just one question about definitions `CanBuild2-10` and pattern matching of the `~` case class. Do you think we can do that?

jenshalm:
Ah, yes, it's actually funny I did not think about it myself, as it was me who suggested this technique in the other discussion. :-) I think I never thought about it as I adopted the pattern from Sadek Drobi's gist linked in the scaladoc of ApplicativeBuilder and tried to stay close to his concept. I'll see if I can quickly change that.

Do you have any preferences regarding 1) and 2) above?

vkostyukov:
I think we can put the applicative implementation in `request` package, since there is no conflicts with current implementation. Regarding the `~` combinator. How would the pattern matching look like for `|@|` symbol? I'm thinking that it's a really good idea of being forward compatible with scalaz.

jenshalm:
I just switch to the pattern match syntax. The implementation is now very minimalistic. Thanks for reminding me. 

Being forward compatible is no longer possible when we keep the pattern match syntax, as scalaz would require the old syntax I just removed. So we can as well stick with the `~` symbol.

Final step would be some repackaging to make it available with the `io.finch.request._` import.

vkostyukov:
Sounds good for me. Could you also please take a quick look at the places like `a,b,c` and rewrite them as `a, b, c` just to stick the Scala style.

jenshalm:
Ok, moved everything into request package and added spaces.

vkostyukov:
Cool! Thanks a lot @jenshalm! Just one question. Do you think we can replace `Builder` and `BuilderOps` with just one implicit class. They both looks pretty similar, I was wondering if we could avoid this tiny duplication. I'm talking about

```scala
implicit class Builder[M[_], A](ma: M[A])(implicit canBuild: CanBuild[M]) {
  def ~[B](mb: M[B]): M[A ~ B] = canBuild(ma, mb)
  def apply[B](f: A => B): M[B] = canBuild.map(ma, f)
}
```

jenshalm:
One now extends the other to reduce duplication. I think we should not fully merge them, as you probably would not want to add the apply method to all supported container types implicitly, they only need the `~` method.

vkostyukov:
Agreed. We don't need an extra apply method for every request reader.

vkostyukov:
Looks very good for me. Let's ask other contributors for the review. 

@benjumanji, @travisbrown, @rodrigopr, @rpless could you please have a look at this change?

vkostyukov:
@jenshalm I have an idea on how to simplify this. I don't really think we need `map` method in `CanBuild` since it just repeats request reader's `map`. Thus, we might define `CanBuild` and `Builder` as follows:

```scala
trait CanBuild[M[_]]{
  def apply[A,B](ma: M[A], mb: M[B]): M[A ~ B]
}

implicit class Builder[M[_], A](ma: M[A])(implicit canBuild: CanBuild[M]) {
  def ~[B](mb: M[B]): M[A ~ B] = canBuild(ma, mb)
}
```

So, the usage would involve request reader `map` method instead of `CanBuild.apply`. Like this:

```scala
val reader: RequestReader[(Int, Double, Int)] = 
  (RequiredIntParam("a") ~
   RequiredDoubleParam("b") ~
   RequiredIntParam("c")) map { 
     case a ~ b ~ c => (a, b, c)
   }
```

Which is very clear, since everybody is familiar with `map` method on `RequestReader`. I'm saying that `Builder.apply()` is just an additional way to do the things we already know how to do. Users, familiar with `RequestReader` with find `map` method more intuitive then `Builder.apply`. What do you think? I'm pretty sure we will break an applicative functor as a fundamental abstraction, but our goal is to make the things simple and make them work.

vkostyukov:
BTW, I just realized that we can make it much simpler and lightweight by replacing both `CanBuild` and `Builder` with `~` method on `RequestReader`.

jenshalm:
Yes, of course you can reduce the lines of code. The `CanBuild` abstraction was meant to add support for applicative behaviour in a generic way using the type class pattern. If you don't want or need that, you can always hard-code everything directly into RequestReader. I don't mind either way.

vkostyukov:
I'm just thinking that we don't really need a generic applicative functor for now, since there is only one place we use it. So, we can probably go with simple method. Once we decide to migrate to some functional library we may consider reusing the generic applicative functor from them.

jenshalm:
Ok, this is the hard-coded variant now.

vkostyukov:
Thanks a lot! I'm merging this.

vkostyukov:
Thanks agian Jens @jenshalm! I really appreciate your time on this. This is one of the most significant PRs that I've merged in. I hope you will continue contributing into the project. Feel free to open issues or PRs in future. I feel like your vision match well the project roadmap.

benjumanji:
Myself and Jens are colleagues and I am planning on basing a whole raft of
infrastructure on finagle + finch. So I am sure you will see more of us.

vkostyukov:
This sounds awesome! You guys rock :) Im pretty sure we will end up having more functional patches like this but we should not break bad and end up having `finchz`, since there are Finch users who don't know the category theory but still able to write of nice and state-less services.

vkostyukov:
Also a genteelly reminder: I would love to see you guys in the adopters list.

benjumanji:
I am going to do it as soon as we have the first thing in production.

jenshalm:
I already have another PR in the back of mind, the one I was hinting at regarding making RequestReader more flexible and consistent. I'll probably get around to preparing that over the weekend. That one cannot keep 100% compatibility with existing APIs though, so it'll probably require some debate. But I'd love to have a working suggestion before starting a debate first.

Regarding your comment about becoming `finchz`: In case it wasn't clear, my main motivation for this PR was improved error handling, not to please the FP gods. :-) I'll probably never use the monadic style myself unless one reader depends on the result of another. Our client devs just do not appreciate to get one error reported when they actually made five.

Besides, I just realized that I forgot to add a section about the new functionality to `docs.md`. Should I do another PR for that?

vkostyukov:
Please go ahed with the PR about request readers. I'm totally fine with minor API changes if it worth doing them.

Looks like I wasn't clear in my joke about `finchz`. Applicative behaviour (error collection) is something that I wanted to do since version 0.0.5 but didn't really know how to do that using the monadic API, which is impossible of course. Thus, I really appreciate your work on this. I'm just saying that Finch users are not super-functional programmers, so we have to be carefully with it and try to not scare them :) 

Yeah. That would be great. Otherwise, I was planing to revise the docs file by myself. But I would be happy get some help wit it.

jenshalm:
I'll write something up today or tomorrow.

Issue #26, opened 2014-05-26
Composable ValidationRule's


vkostyukov:
```scala
for {
 e <- RequiredParam("email")
 p <- RequiredParam("password")
 _ <- "password" shouldNot "be empty" when { p.nonEmpty }
 _ <- "password" should "be longer then 10 symbols" when { p.length > 10 }
} yield (e, p)
```

vkostyukov:
Since we introduce the applicate request reader we might think about refactoring the validation rule. I think we can go with something like `when` method on a request reader.

val r = RequiredInt("a") when("greater then 10") { _ > 10 }

rpless:
I like this a lot better than adding implicits to the String type. I can try to put something together tomorrow (I should have time seeing as I'm snowed in, #snowpocalypse2015).

vkostyukov:
Yeah. Ryan could you also make a research regarding `withFilter` method? Maybe we can use it instead of `when` and get some syntactic sugar in for-comprehension.

jenshalm:
Ok, continuing the discussion here. We can keep validation errors with two properties, but we cannot have a single class with a string parameter for the item. I think we have two options:

Option 1, little explosion of error types:
```scala
ParamNotValid(name: String, rule: String)
HeaderNotValid(name: String, rule: String)
CookieNotValid(name: String, rule: String)
BodyNotValid(rule: String)
RequestNotValid(rule: String)
```

Option 2, keep one error, but use a little ADT for the item:
```scala
ValidationFailed(item: RequestItem, rule: String)

sealed abstract class RequestItem
case class Param(name: String) extends RequestItem
case class Header(name: String) extends RequestItem
case class Cookie(name: String) extends RequestItem
case object Body extends RequestItem
case object Request extends RequestItem
```

Both options have pros and cons. Option 1 is better for scaladoc, as in option 2 the types like `Param` show up and users would not immediately realize that they are only used in exceptions. OTOH, for the internal design option 1 would probably lead to a cleaner impl as we can pass these `RequestItem` instances around until we need them.

Note that in both options we need the generic case `Request` or `RequestNotValid` for all cases where we `flatMap` or combine multiple items where we cannot (and usually would not want to) keep track of the original items.

jenshalm:
There is an option 3 to keep it simple and continue using the existing type, but the `item` property could no longer just hold the name then. So if we keep:

```scala
ValidationFailed(item: String, rule: String)
```

[...]

PR #151, opened 2015-01-28
Validation DSL

rpless:
This PR has a basic validation dsl for request readers as in #26. It takes the form of a `when` method on `RequestReader`s.
Basic usage looks like:
```scala
  val user: RequestReader[User] = for {
    name <- RequiredParam("name").when("should be greater then 5 symbols") { name.length > 5 }
  } yield User(userId, name, Seq.empty[Ticket])
```
Would love to get some feedback on this.

### DIFF
--- a/core/src/main/scala/io/finch/request/package.scala
+++ b/core/src/main/scala/io/finch/request/package.scala
@@ -27,7 +27,7 @@
 package io.finch
 
 import com.twitter.finagle.httpx.Cookie
-import com.twitter.util.{Future,Return,Throw,Try}
+import com.twitter.util.{Future,Return,Throw}
 
 import scala.reflect.ClassTag
 
@@ -76,7 +76,25 @@ package object request {
     }
 
     // A workaround for https://issues.scala-lang.org/browse/SI-1336
-    def withFilter(p: A => Boolean) = self
+    def withFilter(p: A => Boolean) = self.should("not fail validation")(p)
+
+    /**
+     * Validate the result of this ''RequestReader'' using a predicate. The rule is used for error reporting.
+     *
+     * @param rule Text describing the rule being validated
+     * @param predicate Predicate that returns true if the data is valid
+     *
+     * @return Return a ''RequestReader'' that will return the value of from this reader if it is valid.
+     *         Otherwise a ''RequestReaderError'' is returned as ''FutureException''.
+     */
+    def should(rule: String)(predicate: A => Boolean): RequestReader[A] = new RequestReader[A] {
+      def apply[Req](req: Req)(implicit ev: Req => HttpRequest): Future[A] = {
+        self(req).flatMap { a =>
+          if (predicate(a)) a.toFuture
+          else new RequestReaderError(s"${self.toString} should $rule.").toFutureException
+        }
+      }
+    }
   }
 
   /**
@@ -162,6 +180,8 @@ package object request {
       def apply[Req](req: Req)(implicit ev: Req => HttpRequest) =
         try number.toFuture
         catch { case _: IllegalArgumentException => ValidationFailed(param, rule).toFutureException }
+
+      override def toString: String = s"param: $param"
     }
   }
 
@@ -213,6 +233,8 @@ package object request {
           case Some(value) => value.toFuture
           case None => ParamNotFound(param).toFutureException
         }
+
+      override def toString: String = s"Required parameter '$param'"
     }
   }
 
@@ -334,6 +356,8 @@ package object request {
       new RequestReader[Option[String]] {
         def apply[Req](req: Req)(implicit ev: Req => HttpRequest) =
           req.params.get(param).toFuture
+
+        override def toString: String = s"Optional parameter '$param'"
       }
   }
 
@@ -479,6 +503,8 @@ package object request {
               case filtered => filtered.toFuture
             }
           }
+
+        override def toString: String = s"Required parameters '$param'"
       }
   }
 
@@ -600,6 +626,8 @@ package object request {
       new RequestReader[List[String]] {
         def apply[Req](req: Req)(implicit ev: Req => HttpRequest) =
           req.params.getAll(param).toList.flatMap(_.split(",")).filter(_ != "").toFuture
+
+        override def toString: String = s"Optional parameters '$param'"
       }
   }
 
@@ -722,6 +750,8 @@ package object request {
           case Some(value) => value.toFuture
           case None => HeaderNotFound(header).toFutureException
         }
+
+      override def toString: String = s"Required header '$header'"
     }
   }
 
@@ -743,6 +773,8 @@ package object request {
       new RequestReader[Option[String]] {
         def apply[Req](req: Req)(implicit ev: Req => HttpRequest) =
           req.headerMap.get(header).toFuture
+
+        override def toString: String = s"Optional header '$header'"
       }
   }
 
@@ -756,6 +788,8 @@ package object request {
         case Some(body) => body.toFuture
         case None => BodyNotFound.toFutureException
       }
+
+    override def toString: String = "Required body"
   }
 
   /**
@@ -768,6 +802,8 @@ package object request {
         case Some(length) if length > 0 => Some(RequestBody(req)).toFuture
         case _ => Future.None
       }
+
+    override def toString: String = "Optional body"
   }
 
   /**
@@ -813,7 +849,10 @@ package object request {
    */
   object RequiredBody {
     def apply[A](implicit m: DecodeMagnet[A]): RequestReader[A] = OptionalBody[A].flatMap {
-      case Some(body) => ConstReader(body.toFuture)
+      case Some(body) => new RequestReader[A] {
+        def apply[Req](req: Req)(implicit ev: Req => HttpRequest) = body.toFuture
+        override def toString: String = "Required body"
+      }
       case None => ConstReader(BodyNotParsed.toFutureException)
     }
     // TODO: Make it accept `Req` instead
@@ -834,6 +873,8 @@ package object request {
     def apply(cookieName: String) = new RequestReader[Option[Cookie]] {
       def apply[Req](req: Req)(implicit ev: Req => HttpRequest): Future[Option[Cookie]] =
         req.cookies.get(cookieName).toFuture
+
+      override def toString: String = s"Optional cookie '$cookieName'"
     }
   }
 
@@ -855,6 +896,8 @@ package object request {
           case None => CookieNotFound(cookieName).toFutureException
           case Some(cookie: Cookie) => cookie.toFuture
         }
+
+      override def toString: String = s"Required cookie '$cookieName'"
     }
   }
 
--- a/core/src/test/scala/io/finch/request/BodySpec.scala
+++ b/core/src/test/scala/io/finch/request/BodySpec.scala
@@ -46,6 +46,10 @@ class BodySpec extends FlatSpec with Matchers {
     a [BodyNotFound.type] should be thrownBy Await.result(futureResult)
   }
 
+  it should "have a toString that produces a string representation of itself" in {
+    RequiredArrayBody.toString should equal(s"Required body")
+  }
+
   "An OptionalArrayBody" should "be properly read if it exists" in {
     val request: HttpRequest = requestWithBody(fooBytes)
     val futureResult: Future[Option[Array[Byte]]] = OptionalArrayBody(request)
@@ -56,6 +60,10 @@ class BodySpec extends FlatSpec with Matchers {
     val request: HttpRequest = requestWithBody(Array[Byte]())
     val futureResult: Future[Option[Array[Byte]]] = OptionalArrayBody(request)
     Await.result(futureResult) should equal(None)
+  }
+
+  it should "have a toString that produces a string representation of itself" in {
+    OptionalArrayBody.toString should equal(s"Optional body")
   }
 
   "A RequiredStringBody" should "be properly read if it exists" in {
--- a/core/src/test/scala/io/finch/request/RequestReaderValidationSpec.scala
+++ b/core/src/test/scala/io/finch/request/RequestReaderValidationSpec.scala
@@ -1,0 +1,58 @@
+/*
+ * Copyright 2014, by Vladimir Kostyukov and Contributors.
+ *
+ * This file is a part of a Finch library that may be found at
+ *
+ *      https://github.com/finagle/finch
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * You may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * Contributor(s):
+ * Ryan Plessner
+ */
+
+package io.finch.request
+
+import com.twitter.finagle.httpx.Request
+import com.twitter.util.Await
+import org.scalatest.{Matchers, FlatSpec}
+
+class RequestReaderValidationSpec extends FlatSpec with Matchers {
+
+  val request = Request(("foo", "6"))
+  val reader = RequiredIntParam("foo")
+
+  "A RequestReader" should "allow valid values" in {
+    val evenReader = reader.should("be even") { _ % 2 == 0 }
+    Await.result(evenReader(request)) shouldBe 6
+  }
+
+  it should "raise a RequestReader error for invalid values" in {
+    val oddReader = reader.should("be odd") { _ % 2 != 0 }
+    a [RequestReaderError] should be thrownBy Await.result(oddReader(request))
+  }
+
+  it should "allow valid values in a for-comprehension" in {
+    val readFoo: RequestReader[Int] = for {
+      foo <- reader if foo % 2 == 0
+    } yield foo
+    Await.result(readFoo(request)) shouldBe 6
+  }
+
+  it should "raise a RequestReader error for invalid values in a for-comprehension" in {
+    val readFoo: RequestReader[Int] = for {
+      foo <- reader if foo % 2 != 0
+    } yield foo
+    a [RequestReaderError] should be thrownBy Await.result(readFoo(request))
+  }
+}
--- a/core/src/test/scala/io/finch/request/CookieSpec.scala
+++ b/core/src/test/scala/io/finch/request/CookieSpec.scala
@@ -63,4 +63,9 @@ class CookieSpec extends FlatSpec with Matchers {
 
     Await.result(futureResult) should equal(None)
   }
+
+  it should "have a toString that produces a string representation of itself" in {
+    RequiredCookie(cookieName).toString should equal(s"Required cookie '$cookieName'")
+    OptionalCookie(cookieName).toString should equal(s"Optional cookie '$cookieName'")
+  }
 }
--- a/core/src/test/scala/io/finch/request/HeaderSpec.scala
+++ b/core/src/test/scala/io/finch/request/HeaderSpec.scala
@@ -18,6 +18,7 @@
  * limitations under the License.
  *
  * Contributor(s):
+ * Ryan Plessner
  */
 
 package io.finch.request
@@ -53,5 +54,11 @@ class HeaderSpec extends FlatSpec with Matchers {
     val request = Request()
     val futureResult = OptionalHeader("Location")(request)
     Await.result(futureResult) should be (None)
+  }
+
+  "A Header Reader" should "have a toString that produces a string representation of itself" in {
+    val header = "Location"
+    RequiredHeader(header).toString should equal(s"Required header '$header'")
+    OptionalHeader(header).toString should equal(s"Optional header '$header'")
   }
 }
--- a/core/src/test/scala/io/finch/request/OptionalParamSpec.scala
+++ b/core/src/test/scala/io/finch/request/OptionalParamSpec.scala
@@ -42,6 +42,11 @@ class OptionalParamSpec extends FlatSpec with Matchers {
     Await.result(futureResult) should equal(None)
   }
 
+  it should "have a toString that produces a string representation of itself" in {
+    val param = "foo"
+    OptionalParam(param).toString should equal(s"Optional parameter '$param'")
+  }
+
 
   "An OptionalBooleanParam" should "be parsed as an integer" in {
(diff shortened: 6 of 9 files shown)


rpless:
@vkostyukov regarding your comment on using `withFilter`. There are a couple of issues. The first issue I see is that it won't carry a useful error message. In addition, to get it to work we would probably need to make the RequestReader both a `FilterMonadic` and a `GenTraversableOnce`. It already has most of the methods to be a `FilterMonadic` but the `GenTraversableOnce` has quite a number of methods we would need to implement. I think the being able to leverage the `for` syntax would be really nice, but I'm not sure its worth it. Let me know what you think.

vkostyukov:
@rpless are you sure we need to implement those interfaces? I just did some experiments:

```scala
trait R[A] { self =>
  def apply(): A

  def flatMap[B](fn: A => R[B]): R[B] = new R[B] {
    def apply() = fn(self())()
  }

  def map[B](fn: A => B): R[B] = new R[B] {
    def apply() = fn(self())
  }

  def withFilter(p: A => Boolean): R[A] = new R[A] {
    def apply() = if (p(self())) self() else throw new Exception()
  }
}

def r[A](a: A) = new R[A] {
  def apply() = a
}

val rr = for {
  x <- r(10) if x > 5
} yield x // compiles into r(10).withFilter(_ > 5).map(_)

println(rr())
```

It works great. And I really enjoy this lightweight syntax for writing simple validation rules. The only issue here is as you said it's unuseful error message. Do you think it's possible to propagate a `param` name into the error somehow?

rpless:
Interesting, I was under the impression that it should work the way you are doing it, but I was getting some issues with `PredicateDoesNotObtain`. Let me take another crack at it. I like the syntax as well. 

We might be able to propagate the param name, I'd need to think about it a little more. There is another problem if it though. Now that I think about it propagating the param name does necessarily make sense in all cases. Take for example the case where we are reading Json off the request.

vkostyukov:
Right. How about to have something like `name` val/def in `RequestReader`? We can define it as `body` for the request readers that read request's body. Otherwise, it will be always be param/header/cookie.

rpless:
Just added `withFilter` to the PR. Working on names in the RequestReader next.

vkostyukov:
In your example, first `radFoo` should has type `Future[Int]` instead. I think we're ok here, since `Future` is not our responsibility, but we may provide an nice syntactic sugar for our users who compose readers. 

Regarding the `name` field. I was solving a similar problem while working on route combinators. I decided to make it easy and use `toString` method that returns a string representation of the part of a route. Thus, I didn't need to have an extra field in the trait. I think we can go with similar approach and have `toString` method on request reader that return a string representation of the reader. It may be "a@param", "b@header", "c@cookie" or similar. So, we can use it in "withFilter" method.

I just realized that we have to refactor `ValidationFailed` exception since now it's actually a `ParamValidationFailed`. We should probably consider refactor it to avoid confusions.

rpless:
Yes, I agree `Future` isn't our issue.
I like using `toString` better, so I'll do that instead.
As for your point on `ValidationFailed`, I can add that. Do you want a separate PR.

Also seems the build is failing because Coveralls doesn't handle the Travis build matrix well. Not sure what to do about that.

vkostyukov:
Just restarted it. Refactoring `ValidationFailed` is kinda tricky problem. I'm not sure how to refactor it. We can address it later.

vkostyukov:
Ryan, what do you think about `shouldAlso` instead of `when` method:

```scala
for {
  a <- RequiredParam("a") shouldAlso("be less then 10") { _ < 10 } shouldAlso("not equal to zero") { _ != 0 } 
  b <- RequiredParam("b") shouldAlso("be positive") {_ > 0 } 
} yield (a, b)
```

vkostyukov:
We should probably ask @jenshalm what he had in mind about improvements in request reader. This `withFilter` thing showed that request validation is a lack of flexibility. I have sort of feelings that we probably deprecate `ValidationRule` in favour of `withFilter`/whatever and applicative request reader. Thus users would use `withFiler` approach with monadic API for validation of a single thing (param, header) and use applicative API for validating two or more things.

`a` does not depend of `b`
```scala
val ab {
  a <- RequiredParam("a") if a > 0
  b <- RequiredParam("b") if b > 0
} yield (a, b)
```

`a` depends on `b`
```scala
val ab = for {
  a ~ b <- RequiredParam("a") ~ RequiredParam("b") if a > 0 && b > 0 && a > b
} yield (a, b)
```

jenshalm:
Yes, that was actually part of my plan for the weekend, the suggestions I planned for RequestReader would include something like rpless's first code example. The name needs to be discussed, I don't have a strong opinion, but I would also think about reuse of validators (combining message and predicate once and then applying). Something more generic like `when`, `ensure` or `validate` might work better for cases like negation, which would read strangely when the method name is `shouldBe` or something like that. 

I personally find the `withFilter` approach very unattractive due to the lack of meaningful error messages.

rpless:
I personally like `should` for the method name. Using Vladimir's suggestion about passing `RequestReader` error information via the `toString` method, I can get error information from the both the method and `withFilter`. So if we have:
```scala
val user: RequestReader[User] = for {
  name <- RequiredParam("name").should("be greater then 5 symbols") { _.length > 5 }
} yield User(userId, name, Seq.empty[Ticket])
```
and we pass a name that is 5 characters or less we will get the error: "Required Param 'name' should be greater than 5 symbols."
We can use this with a `withFilter` too:
```scala
val NameReader = RequiredParam("name")
val user: RequestReader[User] = for {
  name <- NameReader if name.length > 5 
} yield User(userId, name, Seq.empty[Ticket])
```
and we pass a name that is 5 characters or less we will get the error: "Required Param 'name' should not fail validation." Obviously, not the greatest error, but we do get the name of the param/body/whatever that is failing to be validated.

@vkostyukov I agree with deprecating `ValidationFailed` in favor of either this PR (once its done) or something similar.

vkostyukov:
I'm with Ryan @rpless on supporting a lightweight syntax `if`. It's totally not a big deal to support `withFilter` once we have `should` or similar. I can think of use cases when you don't need a reasonable error message rather then indicator that something is broken. For such kind of cases `withFilter` is a perfect solution.

jenshalm:
I wasn't arguing against supporting `withFilter`, I was more thinking around user guidance and documentation. You said `if you need to validate two params you use applicative style", that is more the bit I tend to disagree with. I would use it in almost all cases, just for the better error handling (and I can at least say that view is shared by my entire team). 

It is perfectly fine to continue to support monadic style, but imho the docs should make it clear that the main difference between the two styles is not academic, but in the error handling. Then each user can decide how much she cares about full-blown error messages. Just mentioning it as I am about to write a paragraph about applicative style for the docs.

jenshalm:
Ah, and I agree that `should` works well as a method name.

vkostyukov:
@jenshalm Right. I was just worried about users who already have a codebase written in a monadic style. I'm personally also prefer applicative style. Thus, the plan of preferring applicative style but supporting both sounds like a reasonable roadmap. Looking forward to the paragraph about applicative style!

core/src/main/scala/io/finch/request/package.scala
```diff
@@ -76,7 +76,25 @@ package object request {
...
+     *
+     * @param rule Text describing the rule being validated
+     * @param predicate Predicate that returns true if the data is valid
+     *
+     * @return Return a ''RequestReader'' that will return the value of from this reader if it is valid.
+     *         Otherwise a ''RequestReaderError'' is returned as ''FutureException''.
+     */
+    def should(rule: String)(predicate: A => Boolean): RequestReader[A] = new RequestReader[A] {
+      def apply[Req](req: Req)(implicit ev: Req => HttpRequest): Future[A] = {
+        self(req).flatMap { a: A =>
```
vkostyukov: The type declaration `a: A` is not necessary here.

core/src/test/scala/io/finch/request/BodySpec.scala
```diff
@@ -58,6 +58,11 @@ class BodySpec extends FlatSpec with Matchers {
     Await.result(futureResult) should equal(None)
   }
 
+  it should "have a toString that produces a string representation of itself" in {
+    val param = "foo"
```
vkostyukov: Unused variable `param`.

core/src/test/scala/io/finch/request/RequestReaderValidationSpec.scala
```diff
@@ -0,0 +1,58 @@
...
+import com.twitter.util.Await
+import org.scalatest.{Matchers, FlatSpec}
+
+class RequestReaderValidationSpec extends FlatSpec with Matchers {
+
+  val request = Request(("foo", "6"))
+  val reader = RequiredIntParam("foo")
+
+  "A RequestReader" should "allow valid values" in {
+    val evenReader = reader.should("foo is even") { _ % 2 == 0 }
```
vkostyukov: The `rule` message should probably be `be even` vs. `be odd`.

core/src/test/scala/io/finch/request/RequestReaderValidationSpec.scala
```diff
@@ -0,0 +1,58 @@
...
+  it should "raise a RequestReader error for invalid values" in {
+    val oddReader = reader.should("foo is odd") { _ % 2 != 0 }
+    a [RequestReaderError] should be thrownBy Await.result(oddReader(request))
+  }
+
+  it should "allow valid values in a for-comprehension" in {
+    val readFoo: RequestReader[Int] = for {
+      foo <- reader if foo % 2 == 0
+    } yield foo
+    Await.result(readFoo(request)) should equal(6)
```
vkostyukov: I thought we were using `shouldBe X` syntax. Could you please use it instead of `should equal` in order to make things uniform?

core/src/main/scala/io/finch/request/package.scala
```diff
@@ -768,6 +800,8 @@ package object request {
         case Some(length) if length > 0 => Some(RequestBody(req)).toFuture
         case _ => Future.None
       }
+
+    override def toString: String = "body"
```
vkostyukov: Why body is just a `body` but not `Optional/Required body`?

core/src/main/scala/io/finch/request/package.scala
```diff
@@ -162,6 +180,8 @@ package object request {
       def apply[Req](req: Req)(implicit ev: Req => HttpRequest) =
         try number.toFuture
         catch { case _: IllegalArgumentException => ValidationFailed(param, rule).toFutureException }
+
+      override def toString: String = s"param: $param"
```
vkostyukov: Do we need `toString` method in this helper reader? And I just realised that I have no idea how `toString` should look like for complex readers like this:

```scala
for { 
  a <- RequiredParam("a")
  b <- OptionalParam("b")
} yield (a, b.getOrElse("foo"))
```
I would expect to see something like `RequiredParam 'a' & OptionalParam 'b'` or similar. But I have no idea how to achieve that.

core/src/test/scala/io/finch/request/RequestReaderValidationSpec.scala
rpless: I'll change it. We aren't very consistent throughout our tests with both `shouldBe` v `should equal` and `intercept` v `thrownby`. We probably want to a test clean up pass at some point.

vkostyukov: Yeah, right. We should definitely follow one convention. I'm trying to keep an eye on new tests coming via PRs. I can make a ticket about this to address it later.

rpless: Maybe we should come up with contributing guidelines. Github recommends having a CONTRIBUTING.md file (https://github.com/blog/1184-contributing-guidelines).

core/src/main/scala/io/finch/request/package.scala
rpless: Using `RequiredBody` actually returns a `ConstReader`. I didn't think that the `ConstReader`'s toString should return something about a request body.

core/src/test/scala/io/finch/request/RequestReaderValidationSpec.scala
vkostyukov: Wow! I just wrote the same thing here #156!

core/src/main/scala/io/finch/request/package.scala
rpless: I added it in anticipation of deprecating `ValidationFailed`. We don't necessarily need it now.
I also have no idea what to return for complex readers.

vkostyukov: Maybe it's not really possible to have the item (param, header) name derived from the request reader. We should probably roll back to the approach when we always specify the validated item. I thinking about the following syntax:

```scala
RequiredParam("a") suchThat ("a", "is positive") { _ > 0 }
```

Where `suchThat` is a math term `:` http://mathworld.wolfram.com/SuchThat.html.

We still may have `withFiler` implemented in terms of `suchThat`:

```scala
def withFilter(...) => suchThat("request", "is valid")(p)
```

@rpless, @jenshalm what do you think about this?

jenshalm: I'm not sure about `("a", "is positive")`. It was my plan to think about reuse of validators which would be neat to have in larger projects (I can do this in a separate PR so that we can get this through quicker). In that case you cannot hardcode any parameter names into the validator. I think I would prefer to have a single-item reader add the name of the item automatically while the combined ones do not. This is because if you write a validator for more than one parameter (which is the less common case anyway) you would probably want to write out the names yourself in the message (e.g. startDate should be smaller than endDate). This is hardly something the lib can figure out on its own. Does this make sense?

rpless: I like this better. I'll wait a little bit for other opinions, but if there are no complaints, I'll back out all of the toStrings and then do this instead.

jenshalm: In case it was too vague: I'd really like to leave an open door for the following:

```scala
val bePositive = ValidationRule[Int]("be positive"){ _ > 0}

for { 
  a <- RequiredIntParam("a").should(bePositive)
  b <- RequiredIntParam("b").should(bePositive)
  c <- RequiredIntParam("c").should(bePositive)
} yield (a, b, c)
```

I'm not sure I'd enjoy writing this out every time. And for allowing such a syntax, the param name cannot be in the rule.

jenshalm: Two more reasons against the `suchThat` syntax:

2) it looks like repetition where many users might wonder why they need to repeat the param name

3) how would you write a `suchThat` method on a reader for more than one param then?

rpless: I like the idea of being able reuse validators, but I think we're going to run into the same problem we're seeing in this pull request. How do we get information in the error? In the example above, if one of the RequestReaders fails to validate, how do we know which one it is? We're already having some difficulty with getting the Readers to provide the information we want.

jenshalm: That's what I tried to explain in my first message: For single-item readers the reader can provide this info for free (it knows what it is looking at after all), for the more rare cases of a reader for multiple values the user very likely has to fully phrase out something like `should have a startDate smaller than endDate`, where we could prefix a generic "the request". Does that make it clearer?

rpless: This works for Params, Cookies, and Headers but we're seeing that this doesn't work for a Request Body readers. Required Bodies are actually just `ConstReader`s. I think it would be wrong to make `ConstReader`s report it was a request body and I'm not sure we want to replicate the logic of `ConstReader` inside the Required Bodies.

vkostyukov: We could solve this by wrapping the const reader inside other request reader, so we can override the `toString`/whatever method on the outer class.

jenshalm: The problem with the ConstReader is a minor implementation detail, imho, that can be easily circumnavigated. The current RequestReader implementation with all the copy-paste code is not a piece of beauty that needs to be preserved like the population of Panda bears. :-)

I cited three reasons against the `suchThat` syntax and in my eyes they have more weight than the few problems we'd need to overcome with the original idea. Do you have an elegant solution regarding 3) above for example?

Maybe we can get a few more opinions on that?

vkostyukov: I like the idea of reusable validation rules. We can define a bunch of predefined rules, so users can use them out of the box. The suggested syntax looks good to me. We also may consider the following syntax:
```scala
RequiredIntParam("a") which (ShouldBePositive and ShouldBeLessThen(100))
```
In case if we have method `and` defined on a validation rule.

jenshalm: Yes, that's nice, too

rpless: Yeah, I'm sold on reuseable rules. Let's close this PR.

vkostyukov: This sounds like a big amount of work. But I'm already in love with this feature. @jenshalm do you mind to take care of this task? I think it will be fair, since it's originally your idea. If so, please, consider splitting the `package.scala` file into `RequestReader.scala` and `ValidationRule` since we have two big basic blocks now. @rpless do you mind if Jens continue working on it?

rpless: Its fine. He seems to have a pretty clear idea of the direction it should go. I'm excited to see how it turns out.

jenshalm: Yes, I could take over, I wanted to do some work over the weekend anyway (two more proposals regarding RequestReader, one of them potentially controversial), so I could add validation to that. But I feel we almost could have merged this PR first (I think I could build the reusable rules on top).

Did we settle with renaming `should` to `which`? I think I don't have a strong opinion on the name.

rpless: I can reopen it to be merged. If we want to do that we should decide between `should` and `which`. I prefer `should`.

jenshalm: Yes, I think this PR is a good basis for reusable rules. I'm neutral regarding the name. So we could either wait for more opinions or ask @vkostyukov to merge.

vkostyukov: What I personally like about `which` is that it fits into the whole story pretty well:

```scala
RequiredIntParam("a") which ShouldBePositive // vs
RequiredIntParam("") should BePositive
```

1. It feels like a natural english `required int param which should be positive`
2. In case of `which` we might define the bunch of rules that all start with `Should`, so users will find them easily just by autocompletion "Should...". In case of `should` method we would have both `BePositive` and `NotBePositive`.

I would stick with `which` and rules started with `Should`.

rpless:
Reopening for @jenshalm to use a basis for reusable rules.

core/src/main/scala/io/finch/request/package.scala
rpless: Ok, I'll update it to use `which`.

BenWhitehead: I like the idea of having some validators pre-defined.

Though I have a nit-pick about the naming, it should be `ShouldBeLessThan(100)` rather than `ShouldBeLessThen(100)` (notice th**a**n vs th**e**n)

vkostyukov: On other hand `should` is shorter. Do you think btw, we need a negated variant `shouldNot`?

vkostyukov: I'm not 100% sure about what name is better. Maybe we should vote? :)

vkostyukov: I mean we might have both methods `should` and `shouldNot` defined on `RequestReader` so we can name all the rules as `BeX` like `BePositive`. It seems nice to me:

```scala
RequiredParam("a") shouldNot BePositive
```

jenshalm: Yes, that's nice. However, to add to the confusion :-) `shouldNot` does not combine well with `and`.

vkostyukov: Right. But it's typical confusion for those who write boolean expressions since C: `if (!(flagA && flagB))` :)

jenshalm: Yeah, with the negation `should` has an edge. `whichNot` does not sound so well... :-)

vkostyukov: That's right. Let's merge this for now but rename the `which` -> `should` later. How does it sound?

jenshalm: Sounds good!

rpless: Will change. I also pushed up the fix to get "Required body" and "Optional body" in the respective toString methods.

jenshalm: LGTM. Would be cool to get this merged tonight, so that I can build on top of that over the weekend. What do you think @vkostyukov?

vkostyukov: Sure! I will merge this in once @rpless feel it's ready.

rpless: Let's wait for the build to complete, but I think its there. We have both `should` and `withFilter`, as well as some meaningful errors.

vkostyukov: Thanks a lot Ryan! I'm merging this.

jenshalm: Thanks, that was cool timing, as I have some time over the weekend to work on Finch.

vkostyukov: Cool! Feel free to start an issue i.e., "Composable Validation Rules" if you want to discuss either  design or implementation in details. Otherwise, a PR would be a great reason to start a discussion.

jenshalm: There is one thing I just noticed now and might need to be adjusted in my follow-up work: The should method produces exceptions of type `RequestReaderError` which is a base class and not a case class which breaks the promise made in the documentation: "all the exception throw by `RequestReader` are case classes".

What type of Error should the rules throw instead? `ValidationFailed` is already used by the number converters and has two properties, whereas we need a case class with a simple message property.

vkostyukov: I was thinking about using `ValidationFailed` here. I was hoping that we can keep the feature of representing the error as pair: `item` (previously param) and `rule`. I found this feature pretty useful, since we're able to pattern-match it send the `item` information wrapped in JSON, so clients may easily parse and visualize it. Do you think it's possible to continue sending `ValidationFailed` from new `ValidationRule` with both `item` and `rule`. The frontend teams will be happy to keep this functionality.

vkostyukov: We can continue discussion in https://github.com/finagle/finch/issues/26.

PR #152, opened 2015-01-29
Greedy routers

vkostyukov:
Realised that I missed an important part in routers. In case of two similar routes like `/users/:id` and `/users` it wasn't really clear how to chose the right router in `orElse` combinator. The previous implementation was always taking the first succeed router, which is wrong in the following case:

```scala
val r = (Get / "users" /> GetAllUsers) | (Get / "users" / int /> GetUser)
r("/users/10") // returns GetAllUsers
``` 

This PR introduces a simple and clean approach: always take a router that eats more route tokens (sort of greedy router). I also made a routing more strict - we don't route the requests whose path wasn't fully matched.

### DIFF
--- a/core/src/main/scala/io/finch/route/package.scala
+++ b/core/src/main/scala/io/finch/route/package.scala
@@ -93,7 +93,7 @@ package object route {
     def apply(req: Req): Future[Rep] = {
       val path = requestToRoute(req)
       r(path) match {
-        case Some((_, service)) => service(req)
+        case Some((Nil, service)) => service(req)
         case _ => RouteNotFound(req.path).toFutureException
       }
     }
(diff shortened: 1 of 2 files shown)


vkostyukov:
If somebody knows a better way to find a minimum of values inside two options let me know.

vkostyukov:
Simplified it. I think it's ready to be merged.

vkostyukov:
I'm going to merge it tomorrow. Let me know if you have any concerns about this. //cc @rpless, @rodrigopr, @BenWhitehead, @jenshalm

BenWhitehead:
I think this is a great convenience rather than the user having to do the ordering manually.

rpless:
:+1: although we do need to document it.

vkostyukov:
Sure. I was planing to add a section about route combinators.

PR #154, opened 2015-01-29
document applicative syntax for request readers

jenshalm:
This is for initial documentation, just a single section inserted into the otherwise unchanged document to make sure users know about the feature.

Should there be agreement that the applicative style should be documented as the preferred syntax in the future, more changes would be necessary to rewrite more of the existing reader examples in an applicative style and use the monadic style less often than in the current version of the docs.

### DIFF
--- a/docs.md
+++ b/docs.md
@@ -6,6 +6,7 @@
   * [Custom Request Types](docs.md#custom-request-types)
   * [Request Reader](docs.md#request-reader-monad)
   * [Query String Params](docs.md#query-string-params)
+  * [Improved Error Handling with Applicative Syntax](docs.md#improved-error-handling-with-applicative-syntax)
   * [Param Validation](docs.md#param-validation)
   * [Multiple-Value Params](docs.md#multiple-value-params)
   * [HTTP Headers](docs.md#http-headers)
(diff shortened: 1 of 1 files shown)


vkostyukov:
Super cool! Could you please add a link into the table of content?

vkostyukov:
Nice! Thanks a a lot for writing this.

jenshalm:
Done.

I somehow took it for granted this would be automatic. As the creator of Laika (http://planet42.github.io/Laika/index.html), I am not used to these things requiring manual work. :)

vkostyukov:
Yeah! I've seen the Laika project. Pretty impressive! BTW, do you think it makes sense to generate an html of the `docs.md`, so we can publish it online. We can use Laika in this case :)

jenshalm:
It depends on your preferences - and priorities :-). I'm personally not a big fan of these endless documentation pages, neither for editing nor reading them. With Laika you would get multi-page input and output with cross-linking and auto-generated navigation, but you would probably miss syntax colouring for source code which is not available yet and would need to be added manually based on some client side tool. 

But in general, should the Finch documentation further grow I guess at some point it might become attractive to think about having some way of publishing a real site.

Issue #115, opened 2014-12-15
Improve finch-json


vkostyukov:
At least two improvements should be done:

* Use `tailrec` annotation whenever its possible to use
* Use `scala.Dynamic` for querying
* Provide mechanism of case-class to JSON conversion (this may be moved to `0.5.0` milestone).

vkostyukov:
No longer actual due to #233.

PR #161, opened 2015-02-02
Use SB in Json.ecnode

vkostyukov:
`Json.encode` will be faster with `StringBuilder`. The implementation looks imperative (ugly?) but fast as hell. This addresses bullet N2 in #115.


vkostyukov:
Merging this.

PR #162, opened 2015-02-02
Make the request package more flexible, consistent and simpler - a "Discussion PR" containing 5 potential future PRs for review

jenshalm:
This is a pure "discussion-PR" as it is too big to be applied with one merge. It is presented in this way,
as some of the PRs build on top of the others, so they cannot be split out.

It is therefore best to review the 5 commits one by one, as each of them is meant to become a separate PR.

It is the result of the work on a rainy weekend in London. :-)

## Motivation

The aim of this series of PRs is to improve the simplicity, consistency and flexibility of the functionality contained in the request package. The quick summary of the 5 commits:

1) Add support for reusable and combinable validation rules (e.g. `RequiredIntParam("foo").should(bePositive and beSmallerThan(5))`)

2) Improve `DecodeRequest` API to return `Try[T]` instead of `Option[T]` for improved error handling

3) Make `DecodeRequest` more broadly useful through adding an implicit `as[T]` method to `RequestReader[String]`, `RequestReader[Option[String]]` and `RequestReader[Seq[String]]`

4) Unify error handling based on just 3 error types and refactor the request package for a much simpler implementation

5) Introduce deprecations for the API elements which are now covered in a more generic way by the new features

The following sections provide more detail for each of the PRs

## PR 1: Add support for reusable and combinable validation rules

This corresponds to the feature set discussed here: https://github.com/finagle/finch/issues/26#issuecomment-72353887, so it should hopefully not be controversial.

This PR leaves everything in place that Ryan already added for validation logic. He recently introduced the option to declare validation rules inline:

```scala
RequiredIntParam("foo").should("be positive")(_ > 0)
```

This PR adds to this work the following:

Negation:
```scala
RequiredIntParam("foo").shouldNot("be positive")(_ > 0)  
```

Reusable Rules:
```scala
val bePositive = ValidationRule[Int]("be positive")(_ > 0)

RequiredIntParam("foo").should(bePositive)
```

Combinable Rules:
```scala
val bePositive = ValidationRule[Int]("be positive")(_ > 0)
def beSmallerThan(n: Int) = ValidationRule[Int]("be smaller than $n")(_ < n)

RequiredIntParam("foo").should(bePositive and beSmallerThan(5))
```

The old `ValidationRule` reader has been removed. Due to the name conflict there
was no option to deprecate it first.

This PR also includes a new section in `docs.md` for this feature.

Please ignore everything around error handling in this PR, this is covered by PR 4.

Breaking API changes: Only the removal of the old `ValidationRule`.

## PR 2: Improve `DecodeRequest` API to return `Try[T]` instead of `Option[T]` for improved error handling

This is also very unlikely controversial. The existing API was insufficient for proper error handling.
Returning an `Option` is usually not a good idea in APIs where details about the failure may be helpful.

The new API is simply:
```scala
trait DecodeAnyRequest {
  def apply[A: ClassTag](req: String): Try[A]
}
```

This PR also makes all necessary adaptions for the various JSON integrations (Argonaut, Jawn, Jackson, Finch JSON).

Note that Finch JSON still does not report any error detail, as the underlying public API (Scala SDK JSON parser) is very limited. It should be possible to extract more detail through using a more low-level API here, but this is not yet covered by this PR.  

Breaking API changes: Only for those who manually implemented a `DecodeRequest`. Anyone using the various JSON integration modules should not be affected.

## PR 3: Make `DecodeRequest` more broadly useful through adding an implicit `as[T]` method to all string based readers.

This has not yet been discussed anywhere, but is hopefully very obviously attractive. The use of the `DecodeRequest` API has previously been limited to body parsers. However, this limitation feels arbitrary. It is sometimes just as useful to have for parameters or for headers. As an example, in our current project almost each route contains two parameters which need to be converted to a Joda `DateTime`. Composing this outside of a reader would be cumbersome and feel like the wrong abstraction.

The new `as[T]` method makes it easy to do this together with other parameter validations and conversions:

```scala
val user: RequestReader[TimeSeries] = for {
  groupBy <- RequiredParam("groupBy")
  startDate <- RequiredParam("startDate").as[DateTime]
  endDate <- RequiredParam("startDate").as[DateTime]
} yield TimeSeries(groupBy, startDate, endDate)
```

The above will work as long as an implicit `DecodeRequest[DateTime]` is in scope. 

For headers this feature may come in handy, too. In particular for those where you write custom parsers based on RFCs, like the `Accept` header for example.

The beauty of this feature comes from the fact that the `as[T]` method is not hardcoded anywhere, but implicitly available for any string based reader. The `RequiredCookie` reader for example cannot be directly used with `as[T]` as it is of type `RequestReader[Cookie]`, but the following will work without problems:

```scala
RequiredCookie("visits").map(_.value).as[Int]
```

The `as[T]` method is available for the following reader types:
```scala
RequestReader[String]         --> RequestReader[T]
RequestReader[Option[String]] --> RequestReader[Option[T]]
RequestReader[Seq[String]]    --> RequestReader[Seq[T]]
```

Breaking API changes: Just one (hopefully minor): RequiredParams is now a `RequestReader[Seq[String]]` instead of `List` to make it more generic and matching to the implicit `as` for readers producing a `Seq[String]`.

Otherwise this PR is non-destructive, it only adds to the feature set. In my eyes the old `RequiredBody[T]` should at some point be deprecated in favor of `RequiredBody.as[T]` to make it all feel more unified, but this is not covered by this PR yet.

## PR 4: Unify error handling and simplify the implementation of the `request` package

This is really two PRs in one, but since both require a bigger refactoring it would have been too cumbersome to split this work

#### 1) Unify Error Handling

The error handling has already been discussed here: https://github.com/finagle/finch/issues/26#issuecomment-72385723

It unifies all error types of Finch to just these 4:

```scala
case class RequestReaderErrors(errors: Seq[Throwable]) // when multiple request item were invalid or missing

case class NotFound(item: RequestItem) // when a required request item (header, param, cookie, body) was missing

case class NotParsed(item: RequestItem, targetType: ClassTag[_], cause: Throwable) // when type conversion failed

case class NotValid(item: RequestItem, rule: String) // when a validation rule did not pass for a request item
```

where RequestItem is this ADT:

```scala
  val description = kind + nameOption.fold("")(" '"+_+"'")
}
case class ParamItem(name: String) extends RequestItem("param", Some(name))
case class HeaderItem(name: String) extends RequestItem("header", Some(name))
case class CookieItem(name: String) extends RequestItem("cookie", Some(name))
case object BodyItem extends RequestItem("body")
```

This makes the life for both Finch users and Finch maintainers easier, as it is very consistent and intuitive.

#### 2) Simplify the Implementation of the `request` Package

Apart from the change in error types, this refactoring does not change any public APIs, but it greatly simplifies the implementation of the request package, which should become immediately obvious when skimming over the diff. The simplification was made possible through:

* Introducing a `RequestReader` companion object with a convenient factory method to create new readers with less plumbing
* Introducing an `embedFlatMap` in `RequestReader` which made internal use of `flatMap` obsolete, which usually led to a much more verbose implementation and would not allow us to keep track of the `RequestItem`
* Introducing an implicit `failIfEmpty` on any `RequestReader[Option[A]]` that allows for easy reuse of the same base reader for the required and optional variants
* Using the `should` and `shouldNot` method for internal validation, too

It would be good if future contributions would follow these patterns, too.

Some of the existing reader implementations have been moved to a new `optional.scala` file as they are candidates for deprecation, but this will be discussed in PR 5.

Breaking API changes: Everything around `Exception` types
 

## PR 5: Introduce deprecations for the API elements which are now covered in a more generic way by the new features

This may be controversial, but I still hope to convince Finch maintainers and users to reduce the API.

In particular I'm in favor of deprecating all the conversion readers like `RequiredIntParam` in favor
of the more generic `RequiredParam.as[Int]`. I think it is more consistent, and more functional style
to allow for composition of these features instead of hard-coding just a bunch of them while making it
non-obvious how to do the same for other types if required.

PR 4 also introduced implicit `DecodeRequest` instances for all the types that were previously supported
throug hard-coded readers. The decoders are all just one line, replacing hundreds of lines of plumbing
for the old hard-coded readers.

Since those were very central APIs which might be used in dozens of places in some projects, I feel it is
a bit rude to just remove them. Even though we are pre-1.0 and could do theoretically do anything we want,deprecation is easy to do in this case. I'm already happy if we get these readers removed for 1.0.

With the cleanup and unification, the Finch request package can be explained in 10 minutes:

1) Use one of the following 6 base reader types (Xx stands for Required/Optional)

    * `XxParam`
    * `XxParams`
    * `XxHeader`
    * `XxCookie`
    * `XxStringBody`
    * `XxArrayBody`
    
2) Convert types of any string based reader with `as[T]` combined with an implicit `DecodeRequest[T]`

3) Validate with `should` or `shouldNot` methods, defining rules inline or with reusable validators

4) Compose in applicative syntax with `~` or in monadic syntax with for-comprehensions/`flatMap`

5) Transform readers with `map` or `mapFuture`

6) Handle any errors based on just 4 unified error types

The `mapFuture` method was introduced with PR 4. It is slightly unusual, but extremely useful, the internal implementation uses it a lot, as it is more convenient than `flatMap` in almost all cases.

Regarding naming `XxStringBody` could be renamed back to `XxBody` now, if we remove the old `XxBody[T]` type instead of just deprecating it. I also feel `XxBinaryBody` would be clearer than `XxArrayBody`, but this is just a minor detail.

This PR also updates `docs.md` to use the new API instead of the deprecated one.

Breaking API changes: None, this PR just introduces deprecations.

----------------------------------------

And again, as a reminder: __Please review the 5 commits one by one, it will be overwhelming otherwise__

Btw: as you noticed I had some weird problems with whitespace that make some files appear as completely changed


vkostyukov:
OMG! Let me review this one-by-one. Amazing amount of job!

vkostyukov:
:+1:  on PR1 rather then a couple of scaladoc comments.

vkostyukov:
:+1:  on PR2! Very solid job!

vkostyukov:
:+1: PR3 looks pretty good to me. Do you think we may consider using `RequiredParam("a").as[Int]` instead of `RequiredIntParam()`?

jenshalm:
Regarding PR 3, yes, I suggest deprecating RequiredIntReader in PR 5 for unifying the API. 

And `RequiredParams("ids").as[T]` works with a simple `DecodeRequest[T]`, see the docs I wrote for this PR and the tests. It also works for optional string readers.

Need to catch some sleep, will extract PR 1 tomorrow in the morning and do the little fixes you asked for.

vkostyukov:
Just saw it! Amazing! Feels like you scanned my mind! Don't worry. Take a break. I will carefully review the rest of the changes.

vkostyukov:
@rodrigopr, @rpless, @benjumanji, @BenWhitehead I need some help here :)

vkostyukov:
I would go with renaming `DecodeRequest` to `Decode` and `DecodeAnyRequest` to `DecodeAny` since it's no longer just about bodies but about any request parts/items.

vkostyukov:
Also, @jenshalm please go ahead with renaming `StringBody` to `Body` and `ArrayBody` to `BinaryBody`. This will make things more uniform since there is always `XY.as` where `X` is either `Required` or `Optional` and `Y` is either `Body`, `Param`, `Header` etc.

BenWhitehead:
I don't think I'll be able to review tonight, but will try and get to it tomorrow.

Based on skimming the description I think a lot of things make sense but I'll spend more time digging in tomorrow.

core/src/main/scala/io/finch/request/package.scala
```diff
@@ -49,14 +129,22 @@ package object request {
...
+      val item = MultipleItems
       def apply[Req](req: Req)(implicit ev: Req => HttpRequest) = self(req) flatMap { fn(_)(req) }
     }
 
     def map[B](fn: A => B) = new RequestReader[B] {
+      val item = self.item
       def apply[Req](req: Req)(implicit ev: Req => HttpRequest) = self(req) map fn
     }
     
+    def mapFuture[B](fn: A => Future[B]) = new RequestReader[B] {
```
vkostyukov: I had to add a similar method into `Router`. I called it `mabeMap`. It operates a different type: `Option` vs. `Future`, which is still `F[_]`. I think we should synchronise our names. I made some research on how this method should be named. A request reader, as well as router are particular cases of Kleisli arrow, with output higher kinded type fixed. It usually called just as `flatMap` or `andThen` (see [cat's example](https://github.com/non/cats/blob/master/core/src/main/scala/cats/arrow/Kleisli.scala#L33)). Do you think we both can go with just `flatMap` then? I can rename it in `Router`.

benjumanji: I can't agree with this, composition in the Kliesli category is for a fixed higher kinded type of kind `* -> *`. This is more like flatMap followed by a natural transformation (a function between two functors).

vkostyukov: @benjumanji do you think there is a common name we can use for both `Router` and `RequestReader`? I don't know. Maybe `constMap`?

benjumanji: It's a good question. it 5am, and I should probably be asleep. I'll have a think, but I am sure @jenshalm can come up with a nice name. He has a talent for that kind of thing.

benjumanji: maybe embed? You are taking a function that handles Futures, and using to transform a RequestReader, so it's embedding a function from one space in another?

vkostyukov: Or `hoist`, per article you mentioned. I like both names. Let's ask @jenshalm's opinion.

rodrigopr:
Awesome changes!
Missed most of the discussion, will try to catch up tomorrow.

vkostyukov:
One thing will be broken with new API. Previously we was able to read params/etc like this:

```scala
val a = RequiredIntParam("a")(req)
```

Which may be rewritten using the new API as:

```scala
val a = (RequiredParam("a").as[Int])(req)
```

We have to use extra-brackets around reader, since `as` method takes an implicit parameter. We can either try to solve  by adding an extra method in implicit classes with `as` function. We need a method like this:

```scala
def as[A](req: Req)(implicit ev: Req => HttpRequest, DecodeMagnet[A]): Future[A]
```

Although, This is not a big deal. I think we can just mention it in the documentation.

vkostyukov:
Also, @jenshalm could you please fix a build. It doesn't compile for Scala 2.10: 

```
[error] /home/travis/build/finagle/finch/core/src/main/scala/io/finch/request/package.scala:280: overriding method applyOrElse in trait PartialFunction of type [A1 <: com.twitter.util.Try[T], B1 >: io.finch.request.package.NotParsed](x: A1, default: A1 => B1)B1;
[error]  method applyOrElse has incompatible type
[error]       else RequestReaderErrors(converted collect { case Throw(e) => NotParsed(reader.item, tag, e) }).toFutureException
[error]                                                  ^
[error] one error found
[error] (core/compile:compile) Compilation failed. 
```

You can check it locally by running `sbt +test`.

jenshalm:
Regarding `RequiredIntParam("a")(req)`, yes I noticed you added the workaround for `RequiredBody[A](req)`. If this is seen as essential we can add similar workarounds for the three `as[T]` implementations. However, is this often used in app code? It's probably more likely a shortcut for unit tests?

Yes, I did not run the tests for 2.10, as this was first a mere discussion PR. I will have a closer eye on these things when I extract the individual PRs. I'll also have a look then which of them causes the decrease in coverage.

core/src/main/scala/io/finch/request/package.scala
jenshalm: I have to read the linked article. It's interesting that this method name stirs some debate, as I indeed did not like it much and only picked it due to lack of better ideas.

I'm not sure which name is the most intuitive for people who did not have category theory in primary. :-)

jenshalm:
Regarding the rename from `DecodeRequest` to `Decode`. I think for the semantics it is not strictly necessary, as headers, cookies and params can be seen as part of the request technically (it is not named `DecodeBody` after all). However, I find the shorter name nicer nevertheless. I would just keep it in sync with renaming `EncodeResponse` to `Encode` then.

vkostyukov:
Regarding `RequiredIntParam("a")(req)`. Let's just note in the docs that there're extra brackets required, if the request passed directly do primitive (means the reader, which is not composed by user but implemented directly in Finch) reader. 

Regarding the `Encode`/`Decode`-request. Let's just keep the names for now.

There is one thing that worries me about `as` method. We do have support of Jackson, which implements the `DecodeAny` and `EncodeAny`. So, it might be used as an implicit `decode` in `as` method, since it support any type (including `Int`, etc). @jenshalm could you please add a test to `jackson` package that reads a simple integer param with `as` method. I'm just curious how the compile will handle the situation if there two implicits available in the scope. If it doesn't compile, we at least have to provide a workaround for users who use Jakcson.

jenshalm:
Ah, this is good to know, I will make sure this is tested. This is for PR 3, until that is "hot" I have had the time to look into it (very likely tomorrow evening). 

Is the `DecodeAny` for Jackson optional or is it the only way to decode with Jackson? An implicit for `Any` is usually not a good thing to have. Even if the implicit resolution for `Int` for example works (e.g. through the compiler picking the closest match), the downside would still be that if someone uses a type for parameter decoding without matching implicit in `as[A]` she would get the Jackson Any decoder then instead of an error message.

vkostyukov:
Yeah. `DecodeAnyRequest` doesn't actually returns `Any`, but has a method with type parameter:

```scala
def apply[A](s: String): Try[A]
```

This is why there is a `decodeMagnet` everywhere. For Jackson we have to provide a `ClassTag` as well. Thus, `decodeMagnet` is actually created from both implicit values: `decodeAnyJackson` and `ClassTag`.

The reason we have both `Decode` and `DecodeAny` is that there are JSON libraries that works with case classes and other types directly via reflection. They don't really need new instances of `decode` per each class the used.

BenWhitehead:
Wow, thanks for all the hard work @jenshalm I'm going to reserve doing detailed reviews for each of the individual PRs so that it's easier for me to follow the contained changes.

Most things seem reasonable.

One thing that did stand out to me though is that the new `as[T]` can only be applied to strings.  It would be nice if it could also apply to `Array[Byte]` particularly for the scenario where a binary body is used (perhaps protobuf) that can't be deserialized from a string.  Maybe this is a change for a different PR but I think we should avoid limiting to string when it's completely reasonable to have something be binary.

vkostyukov:
I think we're totally fine with `as` method only on string-based readers. In fact, all the base readers we have are string-based, which feels pretty reasonable to me, since the HTTP protocol is string-based itself. There is only one thing that doesn't feet into this concept - `BinaryBody` or `ArrayBody` previously. What I think we should do is to deprecate/remove it in favour of just `Body`, which is string-based. Thus, users will always be able to get a binary representation by just mapping the reader to `_.getBytes()`.

This approach also works well with protobuf, since we're always ending up with writing an implementations of `DecodeRequest`, which is also string-based.

BenWhitehead:
One problem with that approach though is that there may be weirdness in how the string is encoded possibly leading to loss of data.  I agree that HTTP is primarily string based, however this characteristic does not hold true for the request/response body.

It is very practical to have binary bodies. An example of a binary response body might be to serve an image thumbnail of a users profile image. You don't want to encode to a string from a png, and in the reverse you don't want to decode to a string before trying to decode to a png.

Another scenario I can think of where the server would need to handle a binary body would be if the client posts a gzipped request body, using finch's readers it would very feasable to do something like the following:
```scala
val reader = for {
  bytes <- RequiredArrayBody
  decompressed <- Decompress(bytes)
  bigObject <- Decode(decompressed).as[BigObject]
} yield bigObject
```

jenshalm:
To be honest, I haven't thought about binary bodies for two reasons: I haven't used them for years and I've built `as[T]` on top of the existing `DecodeRequest[T]` functionality which is entirely string based.

I think we should keep `RequiredBinaryBody` as calling `getBytes` on a `RequestReader[String]` feels wrong to me and could be misleading, as in reality it is the other way round, we start with bytes and might turn them into a string.

I don't think we would want a fully generic `as[T]` as this would mean we'd need to move to `DecodeRequest[A,B]` which would overcomplicate the design and implicit resolution and for very specific transformations, you'd always have `map` and `flatMap` on the readers.

What we could do, if there is support for that idea, is to introduce `DecodeBinary[T]` and `EncodeBinary[T]` later. The good news is, should we go that route, it will be trivial to extend `as[T]` to be usable with those, too. However, this is beyond the scope of this PR galore, as we'd also need to enhance the other side, `TurnIntoHttp[T]` which is also currently entirely string based.

Should there be interest in making working with binary protocols more convenient, I'd suggest to create an issue and describe the use cases.

core/src/main/scala/io/finch/request/package.scala
jenshalm: I have been thinking about the name `mapFuture` again. I think it is a bad name and needs to go, in particular as it really is a `flatMapFuture` and we should not use the type name in the method name, so that we could use the same name for other embedded monads.

The names `embed` or `hoist` are nice and short, however, I feel they don't capture the similarity to `map` and `flatMap` in any way (and as a non-native speaker I did not even know what hoist means). If we could live with a longer name, I think `embedFlatMap` is a fairly precise description of what that method does. What do you think?

vkostyukov: `embedFlatMap` sounds good for me. Per cat's example I mentioned, we may also consider `andThen` for this, which is pretty reasonable for me. We read a value and then pipe it into a function. I'm fine with both names.

jenshalm: One of the `andThen` implementations in cat's example is used as a synonym for a plain `flatMap`, whereas we have a special kind of `flatMap`. So while `andThen` clearly is the nicer name, I think `embedFlatMap` is more precise in its distinction from the plain `flatMap` on the reader.

vkostyukov: Agreed. Let's go with `embedFlatMap`. I will take care of `Router`.

jenshalm:
PR 1 is already merged now. 

PR 2 is now active and under review: https://github.com/finagle/finch/pull/165

vkostyukov:
PR2 #165 and PR3 #168 are merged.

vkostyukov:
PR4 #170 is merged

jenshalm:
I'm closing this PR now, as it was only meant to exist for review and discussion.

4 PRs have been merged now, the 5th is active for review now here: https://github.com/finagle/finch/pull/173.

The last PR deals with deprecating APIs, so it would be nice to get feedback from those who have an existing code base using Finch.

PR #163, opened 2015-02-02
Add support for reusable and combinable validation rules - fixes #26

### LINKED ISSUE #26: Composable ValidationRule's

jenshalm:
This is the first of five PRs extracted from https://github.com/finagle/finch/pull/162.

* made all the requested changes in scaladoc
* added three more tests at the end of `RequestReaderValidationSpec`

## PR 1: Add support for reusable and combinable validation rules

This corresponds to the feature set discussed here: https://github.com/finagle/finch/issues/26#issuecomment-72353887, so it should hopefully not be controversial.

This PR leaves everything in place that Ryan already added for validation logic. He recently introduced the option to declare validation rules inline:

```scala
RequiredIntParam("foo").should("be positive")(_ > 0)
```

This PR adds to this work the following:

Negation:
```scala
RequiredIntParam("foo").shouldNot("be positive")(_ > 0)  
```

Reusable Rules:
```scala
val bePositive = ValidationRule[Int]("be positive")(_ > 0)

RequiredIntParam("foo").should(bePositive)
```

Combinable Rules:
```scala
val bePositive = ValidationRule[Int]("be positive")(_ > 0)
def beSmallerThan(n: Int) = ValidationRule[Int]("be smaller than $n")(_ < n)

RequiredIntParam("foo").should(bePositive and beSmallerThan(5))
```

The old `ValidationRule` reader has been removed. Due to the name conflict there
was no option to deprecate it first.

This PR also includes a new section in `docs.md` for this feature.

Please ignore everything around error handling in this PR, this is covered by PR 4.

Breaking API changes: Only the removal of the old `ValidationRule`.


vkostyukov:
Very nice! I would love to merge this as soon as possible to no block the upcoming PRs. If there are no concerns about this I'm merging it today around noon (PST).

core/src/main/scala/io/finch/request/package.scala
```diff
@@ -34,6 +34,55 @@ import scala.reflect.ClassTag
 package object request {
 
   /**
+   * A reusable validation rule that can be applied to any ''RequestReader'' with a matching type.
+   */
+  class ValidationRule[T] private[request] (val description: String, predicate: T => Boolean) extends (T => Boolean) { self =>
```
vkostyukov: What do you think about implementing it as 
```scala
trait ValidationRule[A] { 
  def description: String
  def apply(value: A): Boolean
  ...
}
```

And then:

```scala
object ValidationRule {
  def apply[A](d: String)(p: A => Boolean): ValidationRule[A] = new ValidationRule[A] {
    def description = d
    def apply(value: A) = p(value)
  }
}
```

I know that this is just implementation details but this approach fits well in to the whole system, since other primitives (i.e., `RequestReader` uses it). Also could you please use `A` instead of `T`?

Sorry for being picky :) I just looked through the PR and found out that we may make it look like other Finch code.

jenshalm: I actually thought of this myself (I also prefer consistency as you hopefully see in my work). The only reason I went the other way was that I disliked the way you have to avoid having proper parameter names (your choice of `d` above), which makes scaladoc harder to read. If you think this is not a concern, I can change it.

vkostyukov: Yeah. I see. We might go with `desc` then :) I think it's pretty reasonable and short name. What do you think?

jenshalm: Yes, that should be fine. Do you want me to change that? I'd have to wait 12 hrs before I can get back to work on that. Alternatively, if you have time, you could merge and change it in a follow-up commit, so that I can prepare PR 2 in 12 hrs instead.

vkostyukov: Sure! I will do it.

vkostyukov:
Merging this. Thanks @jenshalm!

PR #165, opened 2015-02-03
improve DecodeRequest API to return Try instead of Option

jenshalm:
This is the 2nd of 5 PRs extracted from https://github.com/finagle/finch/pull/162.

* made all the changes requested in comments in https://github.com/finagle/finch/pull/162

## PR 2: Improve `DecodeRequest` API to return `Try[T]` instead of `Option[T]` for improved error handling

This is also very unlikely controversial. The existing API was insufficient for proper error handling.
Returning an `Option` is usually not a good idea in APIs where details about the failure may be helpful.

The new API is simply:
```scala
trait DecodeAnyRequest {
  def apply[A: ClassTag](req: String): Try[A]
}
```

This PR also makes all necessary adaptions for the various JSON integrations (Argonaut, Jawn, Jackson, Finch JSON).

Note that Finch JSON still does not report any error detail, as the underlying public API (Scala SDK JSON parser) is very limited. It should be possible to extract more detail through using a more low-level API here, but this is not yet covered by this PR.  

Breaking API changes: Only for those who manually implemented a `DecodeRequest`. Anyone using the various JSON integration modules should not be affected.

### DIFF
--- a/docs.md
+++ b/docs.md
@@ -297,13 +297,14 @@ An HTTP body may be fetched from the HTTP request using the following readers:
 Finch supports pluggable request decoders. In fact, any type `A` my be read from the request body using either:
 `io.finch.request.RequiredBody[A]` or `io.finch.request.OptionalBody[A]` if there is an implicit value of type 
 `DecodeRequest[A]` available in the scope. The `DecodeRequest[A]` abstraction may be described as function 
-`String => Option[A]`. Thus, any decoders may be easily defined to use the functionality of body readers. Note, that 
+`String => Try[A]`. Thus, any decoders may be easily defined to use the functionality of body readers. Note, that 
 the body type (i.e., `Double`) should be always explicitly defined for both `RequiredBody` and `OptionalBody`.
 
 ```
 implicit val decodeDouble = new DecodeRequest[Double] {
-  def apply(s: String): Option[Double] =
-    try { Some(s.toDouble) } catch { case _: NumberFormatException => None }
+  
+  def apply(s: String): Try[Double] = Try { s.toDouble }
+  
 }
 val req: HttpRequest = ???
 val readDouble: RequestReader[Double] = RequiredBody[Double]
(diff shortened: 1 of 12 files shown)


docs.md
```diff
@@ -297,13 +297,14 @@ An HTTP body may be fetched from the HTTP request using the following readers:
...
-`String => Option[A]`. Thus, any decoders may be easily defined to use the functionality of body readers. Note, that 
+`String => Try[A]`. Thus, any decoders may be easily defined to use the functionality of body readers. Note, that 
 the body type (i.e., `Double`) should be always explicitly defined for both `RequiredBody` and `OptionalBody`.
 
 ```
 implicit val decodeDouble = new DecodeRequest[Double] {
-  def apply(s: String): Option[Double] =
-    try { Some(s.toDouble) } catch { case _: NumberFormatException => None }
+  
+  def apply(s: String): Try[Double] = Try { Some(s.toDouble) }
```
vkostyukov: Should probably be `Try(s.toDouble)`.

vkostyukov:
Very nice @jenshalm! The plan is following. As usual, I'm going to merge it in around non (PST) and fix tiny things in the upcoming commit, so you can focus on the next PR.

jenshalm:
Ok, I made the little fix in the docs.

It's funny that the coverage goes down even though I barely add new code while I did add 3 tests. If possible I'd like to address coverage with or after PR 4 which is a big refactoring anyway.

Regarding your cleanup, tiny one line changes are fine. For anything bigger it would be nice if you could wait until after PR 4, as I otherwise need to fix conflicts for each PR I prepare.

vkostyukov:
Sure! I can wait, it's not a problem at all. Regarding the coverage report. I agree the reporting numbers are weird. I can barely believe we have 92% coverage right now. So, don't worry. I think we're totally fine while the number remains the same +-1%.

vkostyukov:
Merging this.

PR #168, opened 2015-02-03
Add an implicit as[A] method to all string-based RequestReaders

jenshalm:
This is the 3rd of 5 PRs extracted from https://github.com/finagle/finch/pull/162.

* fixed the compilation issue for Scala 2.10 which was actually caused by a compiler bug, through removing `extends AnyVal` from the implicit classes (as long as we support 2.10)
* fixed issues around `DecodeAnyRequest` and implicit resolution through making the magnet based on `DecodeAnyRequest` lower priority and added a test in `DecodeSpec` to prove this actually works

Note that two files show up as completely changed as someone checked them in with Windows line breaks. I'm not sure what the project's policy is regarding `autocrlf` settings.

## PR 3: Make `DecodeRequest` more broadly useful through adding an implicit `as[A]` method to all string based readers.

This has not yet been discussed anywhere, but is hopefully very obviously attractive. The use of the `DecodeRequest` API has previously been limited to body parsers. However, this limitation feels arbitrary. It is sometimes just as useful to have for parameters or for headers. As an example, in our current project almost each route contains two parameters which need to be converted to a Joda `DateTime`. Composing this outside of a reader would be cumbersome and feel like the wrong abstraction.

The new `as[A]` method makes it easy to do this together with other parameter validations and conversions:

```scala
val user: RequestReader[TimeSeries] = for {
  groupBy <- RequiredParam("groupBy")
  startDate <- RequiredParam("startDate").as[DateTime]
  endDate <- RequiredParam("startDate").as[DateTime]
} yield TimeSeries(groupBy, startDate, endDate)
```

The above will work as long as an implicit `DecodeRequest[DateTime]` is in scope. 

For headers this feature may come in handy, too. In particular for those where you write custom parsers based on RFCs, like the `Accept` header for example.

The beauty of this feature comes from the fact that the `as[A]` method is not hardcoded anywhere, but implicitly available for any string based reader. The `RequiredCookie` reader for example cannot be directly used with `as[A]` as it is of type `RequestReader[Cookie]`, but the following will work without problems:

```scala
RequiredCookie("visits").map(_.value).as[Int]
```

The `as[A]` method is available for the following reader types:
```scala
RequestReader[String]         --> RequestReader[A]
RequestReader[Option[String]] --> RequestReader[Option[A]]
RequestReader[Seq[String]]    --> RequestReader[Seq[A]]
```

Breaking API changes: Just one (hopefully minor): RequiredParams is now a `RequestReader[Seq[String]]` instead of `List` to make it more generic and matching to the implicit `as` for readers producing a `Seq[String]`.

Otherwise this PR is non-destructive, it only adds to the feature set. In my eyes the old `RequiredBody[A]` should at some point be deprecated in favor of `RequiredBody.as[A]` to make it all feel more unified, but this is not covered by this PR yet.

### DIFF
--- a/core/src/main/scala/io/finch/request/package.scala
+++ b/core/src/main/scala/io/finch/request/package.scala
@@ -31,7 +31,33 @@ import com.twitter.util.{Future, Return, Throw, Try}
 
 import scala.reflect.ClassTag
 
-package object request {
+package request {
+  
+  /**
+   * Trait with low-priority implicits to avoid conflicts that would arise from adding
+   * implicits that would work with any type in the same scope as implicits for concrete
+   * types.
+   * 
+   * Implicits defined in super-types have lower priority than those defined in a sub-type.
+   * Therefore we define low-priority implicits here and mix this trait into the package
+   * object.
+   */
+  trait LowPriorityImplicits {
+
+    /**
+     * Creates a ''DecodeMagnet'' from ''DecodeAnyRequest''.
+     */
+    implicit def magnetFromAnyDecode[A](implicit d: DecodeAnyRequest, tag: ClassTag[A]): DecodeMagnet[A] =
+      new DecodeMagnet[A] {
+        def apply(): DecodeRequest[A] = new DecodeRequest[A] {
+          def apply(req: String): Try[A] = d(req)(tag)
+        }
+      }
+    }
+  
+}
+
+package object request extends LowPriorityImplicits {
 
   /**
    * A reusable validation rule that can be applied to any [[RequestReader]] with a matching type.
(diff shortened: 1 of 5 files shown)


jenshalm:
Please note the two changes I mention in the beginning of the description!

The corresponding code changes are:
* https://github.com/finagle/finch/pull/168/files#diff-b558504bf15d7ec9ffdee43ebecd45b4R36
* https://github.com/finagle/finch/pull/168/files#diff-b558504bf15d7ec9ffdee43ebecd45b4R231
* https://github.com/finagle/finch/pull/168/files#diff-69d6a958e4f67713778173ffd22738acR105

core/src/main/scala/io/finch/request/package.scala
```diff
@@ -31,7 +31,33 @@ import com.twitter.util.{Future, Return, Throw, Try}
...
+   * Implicits defined in super-types have lower priority than those defined in a sub-type.
+   * Therefore we define low-priority implicits here and mix this trait into the package
+   * object.
+   */
+  trait LowPriorityImplicits {
+
+  /**
+   * Creates a ''DecodeMagnet'' from ''DecodeAnyRequest''.
+   */
+  implicit def magnetFromAnyDecode[A](implicit d: DecodeAnyRequest, tag: ClassTag[A]): DecodeMagnet[A] =
```
vkostyukov: There is something wrong with indentation here. 2 spaces missing?

vkostyukov:
Breaking news! Some of the Finch contributors is using Windows!

Nice job on resolving the issue with `DecodeAnyRequest`. PR looks very nice to me.

@rpless, @BenWhitehead, @benjumanji, @rodrigopr could you please review?

core/src/main/scala/io/finch/request/package.scala
jenshalm: Indentation is fixed. I'm afk now for 8 hrs.

rpless:
I like it and I agree with Jens' point about deprecating `RequiredBody[A]`. Might be good to do that a release or two done the road.

benjumanji:
:+1:

vkostyukov:
Thanks guys! Merging this.

core/src/main/scala/io/finch/request/package.scala
```diff
@@ -192,7 +218,76 @@ package object request {
...
+
+    /* IMPLEMENTATION NOTE: The three implicit classes for string based readers should
+     * all extend AnyVal to avoid instance creation for each invocation of the extension
+     * method. However, this let's us run into a compiler bug when we compile for Scala 2.10:
+     * https://issues.scala-lang.org/browse/SI-8018. The bug is caused by the combination of
+     * three things: 1) an implicit class, 2) extending AnyVal, 3) wrapping a class with type
+     * parameters. 2) is the only thing we can remove here, otherwise we'd need to move the
+     * body of the methods somewhere else. Once we drop support for Scala 2.10, these classes
+     * can safely extends AnyVal. 
+     */
```
BenWhitehead: Great explanation. We probably want to create an issue to track this `AnyVal` is added back to these implicit class when scala 2.10 support is dropped.

BenWhitehead:
I realise I'm late to the review, but things look good.

Thanks for the great work @jenshalm

vkostyukov:
@BenWhitehead, yeah. Sorry for merging this (and previous PRs) in such a short review-period. I just didn't want to block Jens' work with the upcoming PRs.

BenWhitehead:
@vkostyukov no worries, I've had a busy couple days and probably will the rest of the week. I don't want to hold things up and I feel like I got a good gist of what's going on from the first mega PR.

PR #170, opened 2015-02-04
Unify error handling and simplify implementation of request package - fixes #157

### LINKED ISSUE #157: Semantics for optional number parameters feel wrong
I just stumbled across the implementation for optional number values. I think the behaviour is almost never what a user would expect. If an API has optional parameters, the usual semantics are that they can be left out or otherwise, if present, must be valid. Treating a missing optional value the same as an invalid value can easily lead to silent failure and unexpected behaviour. I would expect:

```
Some(int)       // if the param is present and valid
None            // if the param is missing
ValidationError // if the param is present but invalid
```

I can do a PR should there be agreement on this.

jenshalm:
This is the 4th of 5 PRs extracted from #162.

* Changed the name of the `mapFuture` method to `embedFlatMap` as discussed in the other PR
* Adjusted syntax of new test methods to be aligned with the new contribution guide
* Added more detail for the description for 2) below to make the motivation for this big refactoring clearer
* Added back `extends AnyVal` in two of the three implicit classes for `as[A]` as the simplification of the implementation did no longer trigger the compiler bug there

## PR 4: Unify error handling and simplify the implementation of the `request` package

This is really two PRs in one, but since both require a bigger refactoring it would have been too cumbersome to split this work

#### 1) Unify Error Handling

The error handling has already been discussed here: https://github.com/finagle/finch/issues/26#issuecomment-72385723

It unifies all error types of Finch to just these 4:

```scala
case class RequestReaderErrors(errors: Seq[Throwable]) // when multiple request item were invalid or missing

case class NotFound(item: RequestItem) // when a required request item (header, param, cookie, body) was missing

case class NotParsed(item: RequestItem, targetType: ClassTag[_], cause: Throwable) // when type conversion failed

case class NotValid(item: RequestItem, rule: String) // when a validation rule did not pass for a request item
```

where RequestItem is this ADT:

```scala
  val description = kind + nameOption.fold("")(" '"+_+"'")
}
case class ParamItem(name: String) extends RequestItem("param", Some(name))
case class HeaderItem(name: String) extends RequestItem("header", Some(name))
case class CookieItem(name: String) extends RequestItem("cookie", Some(name))
case object BodyItem extends RequestItem("body")
```

This makes the life for both Finch users and Finch maintainers easier, as it is very consistent and intuitive.

#### 2) Simplify the Implementation of the `request` Package

Apart from the change in error types, this refactoring does not change any public APIs, but it greatly simplifies the implementation of the request package, which should become immediately obvious when skimming over the diff. The simplification was made possible through:

* Introducing a `RequestReader` companion object with a convenient factory method to create new readers with less plumbing
* Introducing an `embedFlatMap` in `RequestReader` which made internal use of `flatMap` obsolete, which usually led to a much more verbose implementation and would not allow us to keep track of the `RequestItem`
* Introducing an implicit `failIfEmpty` on any `RequestReader[Option[A]]` that allows for easy reuse of the same base reader for the required and optional variants
* Using the `should` and `shouldNot` method for internal validation, too

Some of the existing reader implementations have been moved to a new `optional.scala` file as they are candidates for deprecation, but this will be discussed in PR 5.

Breaking API changes: Everything around `Exception` types

### DIFF
--- a/core/src/main/scala/io/finch/request/package.scala
+++ b/core/src/main/scala/io/finch/request/package.scala
@@ -59,6 +59,18 @@ package request {
 
 package object request extends LowPriorityImplicits {
 
+  
+  implicit val decodeInt: DecodeRequest[Int] = DecodeRequest { s => Try(s.toInt) }
+  
+  implicit val decodeLong: DecodeRequest[Long] = DecodeRequest { s => Try(s.toLong) }
+  
+  implicit val decodeFloat: DecodeRequest[Float] = DecodeRequest { s => Try(s.toFloat) }
+  
+  implicit val decodeDouble: DecodeRequest[Double] = DecodeRequest { s => Try(s.toDouble) }
+  
+  implicit val decodeBoolean: DecodeRequest[Boolean] = DecodeRequest { s => Try(s.toBoolean) }
+  
+  
   /**
    * A reusable validation rule that can be applied to any [[RequestReader]] with a matching type.
    */
@@ -117,12 +129,31 @@ package object request extends LowPriorityImplicits {
   }
   
   /**
+   * Representations for the various request item types
+   * that can be processed with ''RequestReaders''.
+   */
+  object items {
+    sealed abstract class RequestItem(val kind: String, val nameOption:Option[String] = None) {
+      val description = kind + nameOption.fold("")(" '"+_+"'")
+    }
+    case class ParamItem(name: String) extends RequestItem("param", Some(name))
+    case class HeaderItem(name: String) extends RequestItem("header", Some(name))
+    case class CookieItem(name: String) extends RequestItem("cookie", Some(name))
+    case object BodyItem extends RequestItem("body")
+    case object MultipleItems extends RequestItem("request")
+  }
+  
+  import items._
+  
+  /**
    * A request reader (a Reader Monad) reads a ''Future'' of ''A'' from the ''HttpRequest''.
    *
    * @tparam A the result type
    */
   trait RequestReader[A] { self =>
 
+    def item: RequestItem
+    
     /**
      * Reads the data from given request ''req''.
      *
@@ -132,14 +163,22 @@ package object request extends LowPriorityImplicits {
     def apply[Req](req: Req)(implicit ev: Req => HttpRequest): Future[A]
 
     def flatMap[B](fn: A => RequestReader[B]) = new RequestReader[B] {
+      val item = MultipleItems
       def apply[Req](req: Req)(implicit ev: Req => HttpRequest) = self(req) flatMap { fn(_)(req) }
     }
 
     def map[B](fn: A => B) = new RequestReader[B] {
+      val item = self.item
       def apply[Req](req: Req)(implicit ev: Req => HttpRequest) = self(req) map fn
     }
     
+    def embedFlatMap[B](fn: A => Future[B]) = new RequestReader[B] {
+      val item = self.item
+      def apply[Req](req: Req)(implicit ev: Req => HttpRequest) = self(req) flatMap fn
+    }
+    
     def ~[B](that: RequestReader[B]): RequestReader[A ~ B] = new RequestReader[A ~ B] {
+      val item = MultipleItems
       def apply[Req] (req: Req)(implicit ev: Req => HttpRequest): Future[A ~ B] = 
         Future.join(self(req)(ev).liftToTry, that(req)(ev).liftToTry) flatMap {
           case (Return(a), Return(b)) => new ~(a, b).toFuture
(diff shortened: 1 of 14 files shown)


jenshalm:
Some additional notes:

* I noticed that the new error class `NotFound` clashes with the name of one of the classes in the response package. Should we live with that, or switch to less nice names like `ItemNotFound`, `ItemNotParsed`, etc?

* We discussed renaming `XxStringBody` back to `XxBody` and `XxArrayBody` to `XxBinaryBody`, but I did not tackle this in this PR yet, as the last one will lead to a discussion around deprecations anyway and it is best to discuss it then.

jenshalm:
Nice side-effect of the simplification: The request package object has 100% coverage now :-)

vkostyukov:
Nice job @jenshalm!

Regarding the name conflicts. Do you think we can keep the current style but rename `NotFound` to `NotExist` or `NotPresent` or `NotGiven`?

jenshalm:
`NotExist` sounds a bit like broken English. :-) The other two sound better, maybe `NotPresent`?

vkostyukov:
Right. I think `NotPresent` is a good option. Let's go with it.

jenshalm:
Just pushed the name change.

core/src/main/scala/io/finch/request/package.scala
```diff
@@ -309,7 +358,9 @@ package object request extends LowPriorityImplicits {
    *
    * @param message the message
    */
-  class RequestReaderError(val message: String) extends Exception(message)
+  class RequestReaderError(val message: String, cause: Throwable) extends Exception(message, cause) {
```
vkostyukov: Can we wrap `cause` in `Option`? I'm worried about `null`. It's not a big deal but this is probably the first occurrence of `null` in the entire codebase :)

core/src/main/scala/io/finch/request/package.scala
```diff
@@ -309,7 +358,9 @@ package object request extends LowPriorityImplicits {
...
+    def this(message: String) = this(message, null)
```
vkostyukov: You can also avoid an additional constructor by just declaring the default value for `cause = None`.

core/src/main/scala/io/finch/request/package.scala
```diff
@@ -320,105 +371,33 @@ package object request extends LowPriorityImplicits {
...
-      def apply[Req](req: Req)(implicit ev: Req => HttpRequest) = const
-    }
-  }
-
-  private[this] object StringToValueOrFail {
-    def apply[A](param: String, rule: String)(number: => A) = new RequestReader[A] {
-      def apply[Req](req: Req)(implicit ev: Req => HttpRequest) =
-        try number.toFuture
-        catch { case _: IllegalArgumentException => ValidationFailed(param, rule).toFutureException }
+  case class NotParsed(item: RequestItem, targetType: ClassTag[_], cause: Throwable)
```
vkostyukov: How about to define an extra `upapply` method that just matches `item` to allow users write:

```scala
x match {
  case NotParsed(what) => ???
}
```

core/src/main/scala/io/finch/request/package.scala
jenshalm: I did not like this either, but this ugliness is caused by interfacing with Java. The `null` is only used when calling a Java constructor, and the base class is not a case class, so it is not accessible from Scala code. And in Java Exception getCause can return `null`: http://docs.oracle.com/javase/7/docs/api/java/lang/Throwable.html#getCause()

So as ugly as it is, I'm afraid it is the right thing to do in this case.

core/src/main/scala/io/finch/request/package.scala
jenshalm: The problem is that you cannot overload `unapply`.

core/src/main/scala/io/finch/request/package.scala
jenshalm: We can add a comment, "interfacing with Java here", so that Scala devs do not get a heart attack when they see the null. :-)

jenshalm: I think there is actually no way to get rid of the null. Even if we'd use Option, at the point we pass it to the Java constructor, we'd need to call `cause.getOrElse(null)`, as Java does not know about Options. And you want to pass the cause up somehow, to stick to the contract of the Java Exception, and to have the cause included when you call `getStacktrace`.

vkostyukov: We can write `cause.orNull` which hides `null` from us :)

core/src/main/scala/io/finch/request/package.scala
vkostyukov: Right. I was surprised by it. Never did this before but was pretty sure it's possible. Nevermind then.

core/src/main/scala/io/finch/request/package.scala
jenshalm: That's a pretty imperfect disguise, I'd say. :-D

I don't see that as an actual improvement, but I also totally would not mind doing that. I'd recommend to keep `cause` non-optional in `NotParsed` though, as we always have a cause there, so it would only be wrapped in a `Some` for calling the super constructor.

vkostyukov: Yeah. Let's keep it with null. I can add a note in the upcoming commits so you probably don't need to update it.

vkostyukov:
Looks good to me :+1: 

Just one question. Do you think we can come up with a shorter name for `RequestReaderErrors`? Just noticed that it becomes more popular, i.e. `XParams` throw it. To fit the current naming style we might thing about:

```scala
NotPresent
NotValid
NotParsed
MultipleNot // :-D
```

What do you think @jenshalm?

jenshalm:
I'm not married to that name. :)

I would not mind changing it if someone has a good idea. `MultipleNot` is just a bit too comical. :)

jenshalm:
Don't think we get `Not` in there in a sensible way. Maybe just `RequestErrors`? But then I'd also shorten the base class.

vkostyukov:
Sounds good: `RequestError` and `RequestErrors`. We can postpone this renaming to the next PR that deprecates the stuff.

vkostyukov:
I'm merging this to allow Jens prepare his 5th PR. Thanks again @jenshalm. Amazing job, as usual!

Issue #164, opened 2015-02-02
Add predefined validation rules for common cases


vkostyukov:
The following list looks like a good start:

```scala
bePositive
beNegative
beLongerThan // for strings
beShorterThan // for string
beLessThan // for numbers
beGreaterThan // for number
```

vkostyukov:
Also it would be nice to allow users write simple rules on optional params:

```scala
val a = OptionalIntParam("a").should("be equal 100") { _ == 100 }
```

jenshalm:
Wouldn't it read better with lowercase: `should(beEven)`. I think it will be more lightweight to implement these rules with `val`s and `def`s, not with `object`s like the readers.

The convenience for optional values is already covered for reusable rules, through an implicit conversion. It would be nice to also cover it for inline rules, but I did not have a good idea yet. I think it would require at least an inline type annotation, to trigger the implicit conversion from `A => Boolean` to `Option[A] => Boolean`.

I don't have time to try this right now, but my gut feeling is you'd end up writing this:
```scala
val a = OptionalParam("a").as[Int].should("be equal 100") { i: Int => i == 100 }
```

vkostyukov:
Agreed. Lower case sounds good for me. I will update the issue description. Thanks!

jenshalm:
Is `beZero` a common use case? And `beEmpty` is a tricky one, people might expect it to work with `Option` and `Seq`. Maybe we can use structural typing here and allow it to be used with anything that has an `isEmpty` method?

vkostyukov:
I was thinking about adding `beEmpty` because of readers that allow empty strings. So, users would explicitly specify the requirement of not being empty. It's no longer actual for now, since we decided to handle it as separate ticket #171. But the idea of structural typing sounds cool! Do you think we can start with simple rules and then extend the set in the next releases?

Regarding the `beZero`. I was thinking of using it as `r shouldNot beZero`. I don't know, it's probably not as useful as I think, since we can use `bePositive` instead., which is the most common use case.

jenshalm:
I see. But yes, I think if you want to add `beEmpty`, using structural types might be more intuitive for the user. Structural typing is based on runtime reflection, so not insanely efficient, but for this little use case it's probably acceptable.

And yes, I think there are more use cases around `should bePositive` and `shouldNot beNegative`. I guess `shouldNot(beZero)` is only relevant for a mathematical app that allows to enter a divisor. :-)

vkostyukov:
[...]

PR #173, opened 2015-02-05
Deprecate APIs which are now supported in a more generic way or have been renamed

jenshalm:
This is the last of 5 PRs extracted from #162.

* Renamed `RequestReaderErrors` and `RequestReaderError` to `RequestErrors` and `RequestError`
* Deprecated `Required/OptionalArrayBody` in favour of `Required/OptionalBinaryBody`
* Rewrote the description below to capture all the discussions that happened in the meantime

## PR 5: Deprecate APIs which are now supported in a more generic way or have been renamed

The recent series of PRs has made the request package more feature-rich and consistent. This would be a great opportunity to deprecate some of the older APIs which are now covered in a more generic way.
This would help keeping the public API simple and minimal and avoid confusion caused by having multiple ways to do the same thing.

In particular this PR suggests the following deprecations:

* Deprecate all hard-coded readers that did type conversions to numbers and booleans in favour of the new `as[A]` method
  (e.g. deprecate `RequiredIntParam("foo")` in favour of `RequiredParam("foo").as[Int]`
* Deprecate `Required/OptionalArrayBody` in favour of `Required/OptionalBinaryBody`
* Deprecate `Required/OptionalStringBody` in favour of `Required/OptionalBody`
* Deprecate `ConstReader` and `EmptyReader` in favour of new factory methods `const`, `value` and `exception` in `RequestReader`

The above deprecations would introduce a grace period for everyone using those APIs. I think in particular for the heavily used `XxIntParam` readers it is nicer to give users some time. Seeing deprecation warnings that explicitly tell you what to use instead also makes it quite easy to modify your code base.

On top of the soft deprecations, this PR also introduces two breaking changes:

* Renamed `RequestReaderErrors` and `RequestReaderError` to `RequestErrors` and `RequestError` (but we have breaking changes in all other error types anyway)
* Removed `XxBody[A]` implementations in favour of `XxBody.as[A]`. Deprecation would have been difficult here due to the name clash.

In general it would be nice to get feedback from people having a code base that uses these Finch APIs.

This PR also updates `docs.md` to use the new API instead of the deprecated one.

Each of the recent PRs added some documentation for the new features, but I feel like I ruined the structure in the process as it makes the docs quite patchy. If there is interest I'd volunteer to do a 6th PR for rewriting the section for the request package over the weekend. I'd open an issue first that describes my ideas for the new structure. The APIs in the request package are now quite simple and consistent and it would be nice if the docs would capture this. It would also be nice to have more of the sample code written in applicative style.

### DIFF
--- a/docs.md
+++ b/docs.md
@@ -123,8 +123,6 @@ transformations will be performed. Reader Monad is sort of famous abstraction th
 ### Query String Params
 
 The following readers are available in Finch:
-* `io.finch.request.EmptyReader` - throws an exception instead of reading
-* `io.finch.request.ConstReader` - fetches a const value from the request
 * `io.finch.request.RequiredParam` - fetches required params within specified type
 * `io.finch.request.OptionalParam` - fetches optional params within specified type
 * `io.finch.request.RequiredParams` - fetches required multi-value params into the list
@@ -136,7 +134,7 @@ case class User(name: String, age: Int, city: String)
 // Define a new request reader composed from provided out-of-the-box readers.
 val user: RequestReader[User] = for {
   name <- RequiredParam("name")
-  age <- RequiredIntParam("age")
+  age <- RequiredParam("age").as[Int]
   city <- OptionalParam("city")
 } yield User(name, age, city.getOrElse("Novosibirsk"))
 
@@ -155,12 +153,12 @@ val user: Json = service(req) handle {
 }
 ```
 
-Optional request readers such as `OptionalIntParam` are quite often used for fetching pagination details.
+Optional request readers such as `OptionalParam` are quite often used for fetching pagination details.
 
 ```scala
 val pagination: RequestReader[(Int, Int)] = for {
-  offset <- OptionalIntParam("offset")
-  limit <- OptionalIntParam("limit")
+  offset <- OptionalParam("offset").as[Int]
+  limit <- OptionalParam("limit").as[Int]
 } yield (offset.getOrElse(0), math.min(limit.getOrElse(50), 50))
 
 val service = new Service[HttpRequest, HttpResponse] {
@@ -171,15 +169,15 @@ val service = new Service[HttpRequest, HttpResponse] {
 ```
 
 #### A `io.finch.request.RequiredParam` reader makes sure that
-* param is presented in the request (otherwise it throws `ParamNoFound(param)` exception)
-* param is not empty (otherwise it throws `ValidationFailed(param, rule)` exception)
-* param may be converted to a requested type `RequiredIntParam`, `RequiredLongParam` or `RequiredBooleanParam`
-(otherwise it throws `ValidationFailed(param, rule)` exception).
+* param is present in the request (otherwise it throws `NotPresent` exception)
+* param is not empty (otherwise it throws `NotParsed` exception)
+* param may be converted to a requested type through calling `as[Int]`, `as[Float]`, etc.
+(otherwise it throws a `NotParsed` exception).
 
-#### An `io.finch.request.OptionalParam` returns
-* `Future[Some[A]]` if param is presented in the request and may be converted to a requested type `OptionalIntParam`,
-`OptionalLongParam` or `OptionalBooleanParam`
-* `Future.None` otherwise.
+#### An `io.finch.request.OptionalParam.as[A]` returns
+* `Future[Some[A]]` if param is present in the request and can be converted to the requested type
+* `Future.None` if the param is missing.
+* `NotParsed` exception if the param is present, but cannot be converted to the requested type
 
 
 ### Improved Error Handling with Applicative Syntax
@@ -192,7 +190,7 @@ case class User(name: String, age: Int, city: String)
 
 val user: RequestReader[User] = 
   (RequiredParam("name") ~
-  RequiredIntParam("age") ~
+  RequiredParam("age").as[Int] ~
   OptionalParam("city")) map {
     case name ~ age ~ city => 
       User(name, age, city.getOrElse("Novosibirsk"))
@@ -227,23 +225,23 @@ case class User(name: String, age: Int)
 // monadic syntax
 val adult: RequestReader[User] = for {
   name <- RequiredParam("name")
-  age <- RequiredIntParam("age").should("be greater than 18"){ _ > 18 }
+  age <- RequiredParam("age").as[Int].should("be greater than 18"){ _ > 18 }
 } yield User(name, age)
 
 // applicative syntax  
 val adult2: RequestReader[User] = 
   (RequiredParam("name") ~
-  RequiredIntParam("age").should("be greater than 18"){ _ > 18 }) map {
+  RequiredParam("age").as[Int].should("be greater than 18"){ _ > 18 }) map {
     case name ~ age => User(name, age)
 }
   
 // reusable validators
-val beNonNegative = ValidationRule[Int]("be non-negative") { _ >= 0 }
-def beSmallerThan(value: Int) = ValidationRule[Int](s"be smaller than $value") { _ < value }
+val bePositive = ValidationRule[Int]("be non-negative") { _ > 0 }
+def beLessThan(value: Int) = ValidationRule[Int](s"be less than $value") { _ < value }
   
 val child: RequestReader[User] = 
   (RequiredParam("name") ~
-  RequiredIntParam("age").should(beNonNegative and beSmallerThan(18))) map {
+  RequiredParam("age").as[Int].should(bePositive and beLessThan(18))) map {
     case name ~ age => User(name, age)
 }
 ```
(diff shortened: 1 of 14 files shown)


vkostyukov:
Thanks @jenshalm! Awesome writeup!

I think we should go with breaking the `XBody` API. The 0.5.0 release is already huge. I personally believe that it's on a right track to 1.0.0. So, in the upcoming releases we may just introduce new features and bug fixes. Thus, I think we should try to make this release solid and break the things. Although, I would also be happy to see a feedback from users.

I heavily used `ConstReader` in the very first Finch project, the project when Finch were born :) It's not a big deal, since it still on 0.1.0, but I think it's a great idea to have `const` and `empty` methods on companion object.

Jens, please go ahead with docs changes. Do you think by the way, we should split it to several files? It's actually was splitted some time ago but I decided to merge everything in order to make it easy for users to read.

jenshalm:
Ok, I'm fine with breaking body APIs, let's just wait if someone complains.

I'm also fine with beefing up the `RequestReader` companion object. I could do that in a separate PR. Do you agree with trying to align names with the `Future` companion object? That would give us `const`, `value`, `exception`. We could add `empty` as a hardcoded variant of `exception`, and maybe `async(Req => Future[A])`, as the `apply` we usually need/use is `Req => A`.

But nevertheless I'm curious about your use cases. Am I right you needed `ConstReader` mainly for `flatMap`, because that need would be gone then. And for `EmptyReader` I lack the phantasy what you'd do with it. :-) Do you have some short sample?

Will do an issue for docs changes tomorrow, and then the actual writing on the weekend.

jenshalm:
Regarding splitting files, that might make sense. I just don't know what kind of navigation GitHub pages offers out of the box. Ideally there'd be some fixed navigation that does not scroll with the content.

vkostyukov:
I'm fine with `Future`-like names. I think we can go with just `const` for now, since it's a base thing for everything else, including `value` and `exception`. What do you think?

The use case was pretty simple :) The answer is in the base class I used, i.e., `Select` that covers all the select-like request to sql db:

```scala
abstract class Select(
    in: RequestReader[Seq[Any]] = ConstReader(Seq.empty[Any]),
    out: RowReader[JSONObject],
    sql: String
  ) extends Service[OAuth2Request[OAuthUser], Seq[JSONObject]] {

    def apply(req: OAuth2Request[OAuthUser]) = Db { client =>
      for {
        params <- in(req)
        query <- client.prepare(sql)
        rep <- query.select(params :_*) { out(_) }
      } yield rep
    }
  }
```

So, if you don't have any query-string params but path params, for example `GET /users/:id`, you would end up:

```scala
case class GetUsers(userId: Long) extends Select(
  in = ConstReader(Seq(userId)),
  out = UserNotFound,
  sql = "select ..."
)
```

This worked pretty well for us, I had base classes for insert, etc.

jenshalm:
Not sure, just having `const` means you'd need to write `RequestReader(Return(5))` and know about `Try` API. I think `value` and `exception` would round it off. What do you think?

vkostyukov:
I was thinking about `const[A](f: Future[A]): RequestReader[A]`, so we can say: `const(1.toFuture)` or `const(UserNotFound.toFutureException)`. How does it sound?

jenshalm:
Ah, I misunderstood that, I thought you meant `const(t: Try[A])` like in Future companion. It'll work, but feels a bit cumbersome, I think I'd expect to have `value(1)` and `exception(UserNotFound)`. I'd mostly use `const` if I obtained a `Future` from somewhere else, not when I specify the result inline. I think a fully comprehensive companion would look more something like this:

```scala
apply(item: RequestItem)(f: Req => A)
async(item: RequestItem)(f: Req => Future[A])
const(value: Future[A], item: RequestItem = Multiple)
value(value: A, item: RequestItem = Multiple)
exception[A](exception: Throwable, item: RequestItem = Multiple)
```

Or do you have an implicit in scope per default for `A => Future[A]`?

All but one of them would be trivial one-line implementations, and give enough convenience for most use cases. Not sure about the name `async`, but since `RequestReader` is essentially a `Req => Future[A]` it feels a bit strange when there is no way to wrap such a function.

docs.md
```diff
@@ -173,13 +173,13 @@ val service = new Service[HttpRequest, HttpResponse] {
...
+(otherwise it throws a `NotParsed` exception).
 
-#### An `io.finch.request.OptionalParam` returns
-* `Future[Some[A]]` if param is presented in the request and may be converted to a requested type `OptionalIntParam`,
-`OptionalLongParam` or `OptionalBooleanParam`
-* `Future.None` otherwise.
+#### An `io.finch.request.OptionalParam.as[A]` returns
+* `Future[Some[A]]` if param is presented in the request and can be converted to the requested type
+* `Future.None` if the param is missing.
+* `NotParsed` exception if the param is presented, but cannot be converted to the requested type
```
BenWhitehead: Possible typo: should be `... exception if the param is present, but cannot be converted to the requested type`

vkostyukov:
Yeah. This looks great to me! Do you think we really need `async`? We can go with `embed`, since there is already `embedFlatMap`.

I posted a question about renaming/deprecating body on Gitter: https://gitter.im/finagle/finch. BTW, feel free to join the room and share your experience with others.

docs.md
```diff
@@ -243,7 +243,7 @@ def beSmallerThan(value: Int) = ValidationRule[Int](s"be smaller than $value") {
   
 val child: RequestReader[User] = 
   (RequiredParam("name") ~
-  RequiredIntParam("age").should(beNonNegative and beSmallerThan(18))) map {
+  RequiredParam("age").as[Int].should(beNonNegative and beSmallerThan(18))) map {
```
BenWhitehead: I think we should change `beSmallerThan` to `beLessThan` or `beLessThanOrEqualTo` so that we're consistent with the operator names these functions are representing.

docs.md
jenshalm: @vkostyukov is this a typo? I just picked that up from the existing docs, so I'm not sure.

vkostyukov: Yeah. Looks like a typo.

docs.md
jenshalm: Yes, `beLessThan` is also the proposed name for the predefined rule in #164. I'll change it.

jenshalm:
I'm undecided regarding `async`. It just feels a bit odd that `RequestReader` is a `Req => Future[A]`, but does not allow to easily wrap an existing function. Not sure about use cases. Maybe it might be useful when you want to deal with large request bodies in non-blocking fashion? `embed` would be a matching name, yes.

vkostyukov:
Yes. I think it's enough for now just to have `apply`, `const`, `value` and `exception`.

jenshalm:
Ok, will do this in a separate PR.

For this one, it looks like there are no complaints yet regarding a breaking change with `XxBody[A]`. You know your users better. :-) I'll just wait until you tell me that it's ok to make the change.

jenshalm:
Actually I think the factory methods do belong into this PR, as they are the replacements for the deprecated `ConstReader`. So I can add them here as well. I don't mind either way.

vkostyukov:
Let's go with renaming `Body`. Also, please go ahead with adding factory methods.

docs.md
BenWhitehead: Great

BenWhitehead:
Looks good, thanks for all the hard work @jenshalm

jenshalm:
Ok, I did the following changes:

* Applied all changes requested in comments
* Deprecated `XxStringBody` in favour of `XxBody`
* Removed `XxBody[A]` implementations
* Introduced factory methods `const`, `value` and `exception` in `RequestReader`
* Updated the description for the PR to reflect those changes

Ironically, for implementing the factory methods, I needed an `embed` method to avoid duplication. Since we did not want to add it now, I made it private. Should users ever ask for it, just make it public.

From my side, I think it is good to be merged now. Should you find something small, feel free to merge and fix it afterwards, as I'm afk now for 8hrs. It's sleeping time in London... :-)

vkostyukov:
Cool! But I can't really see new factory methods in the change. You've probably missed the `git add` phase, which also quite offen happens to me.

jenshalm:
The build failed with `java.io.IOException: Server returned HTTP response code: 403 for URL: http://cobertura.sourceforge.net/xml/coverage-04.dtd`

Looks like something outside of my control. 

Yes, I missed the git add, normally doesn't happen.

vkostyukov:
Thanks! Awesome job @jenshalm!

jenshalm:
One final thing before I forget: I deliberately left the use of the deprecated readers in the tests. Since the deprecated readers all delegate to the new ones, we test all of them. Changing the tests now will decrease coverage. We can change them when we remove the deprecated readers.

vkostyukov:
Right. Sounds good to me. Pretty good idea to test the deprecated readers now.

Issue #174, opened 2015-02-06
Scalastyle sbt plugin?


vkostyukov:
Do we need https://github.com/scalastyle/scalastyle-sbt-plugin? It would be nice to enable it in Travis CI, so we can easily check the incoming PRs style.

Any thoughts?

vkostyukov:
This rule http://www.scalastyle.org/rules-0.6.0.html#org_scalastyle_scalariform_NullChecker is from nightmare I recently had after we merged the patch with `null` :) //c @jenshalm

pasviegas:
There is also https://github.com/puffnfresh/wartremover. Which is more about linting than style.

:)

vkostyukov:
Sounds cool! Thanks @pasviegas! Do you think we can use them both?

vkostyukov:
The `scalastyle` plugin is enabled. But I still would love to see `wartremover` enabled. Moving ticket to the next milestone.

jenshalm:
I wish these engines would allow for annotations on methods to ignore. For a rule like a null check it would be best to have it active per default and then explicitly allow yourself the use per annotation in just one place.

pasviegas:
Yeah, I use both normally, but I miss the annotations, sometimes is necessary to break some rules :)

vkostyukov:
We can also revise the Scalastyle rules and enable as much as we want: http://www.scalastyle.org/rules-0.6.0.html.

vkostyukov:
There is also https://github.com/scala/scala-abide.

vkostyukov:
I've enabled WartRemover and it produces around 100 warnings :) See #200. I think it's enough having just two tools enabled WartRemover + Scalastyle. I just want to make sure that we use all the rules we need from Scalastyle.

vkostyukov:
@jenshalm just found out that we [can suppress some scalastyle warnings](http://stackoverflow.com/questions/21931431/how-can-i-suppress-scalastyle-warning) by comment
```scala
// scalastyle:off null
```

We should probably enable `null`-rule.

jenshalm:
Yes, if they support this kind of annotation, our use of `null` is a perfect use case for that, as we only need it due to exceptional circumstances (java interop in this case). We should enable the rule then.

pasviegas:
Just found out this: https://github.com/scala/scala-abide

Might be good to check it out. :)

vkostyukov:
[...]

PR #176, opened 2015-02-06
Fix 174: Add scalastyle plugin

vkostyukov:
I took the `scalastyle-config.xml` file form the cats project and made some changes:

* Allowed to use `*`, `~` and `/` as object/class name
* Allowed to define more than 70 types in the same file
* Allowed `null`!


vkostyukov:
Merging this. But keep the ticket #174 open.

PR #177, opened 2015-02-06
Add API docs URL to POM

travisbrown:
This is a nice little feature of SBT that adds the base URL of the API docs to the metadata in the POM so that projects that depend on Finch can have their own documentation automatically linked to Finch's.

See the [SBT documentation](http://www.scala-sbt.org/0.13/docs/Howto-Scaladoc.html#Define+the+location+of+API+documentation+for+a+library) for more detail.


vkostyukov:
That's super nice @travisbrown! Thanks a lot!

PR #180, opened 2015-02-07
Split package objects into files

vkostyukov:
Since the codebase is growing, I decided to split the package objects into several files. I used the following rules:

* Each abstraction should has its own file, i.e., `RequestReader`, `Router`, `ValidationRule`
* Each concrete implementation should stay in the package object

This PR is also introduces predefined validation rules as discussed per #164 and allows implicitly convert inlined validation rules for optional param.

Please note the following:
* I realised that it's not possible to have rules such as `bePositive`, `beNegative` since we would have to specify the type exactly for them: `x should bePositive[Int]`, which is annoying. And I found `beGreaterThan(0)` much nicer (and shorter in case for `Long`) for this.
* To allow implicitly convert inlined rules, we have specify the value type explicitly, like this: `x should("bla") { y: Int => ... }`

Let me know if you have any concerns. Going to merge this today.


vkostyukov:
I'm pretty happy with the state of this PR, so I'm going to merge this.

PR #182, opened 2015-02-08
Fix #153: Cleanup scaladoc

### LINKED ISSUE #153: Cleanup scaladoc
It turned out that scaladoc syntax `''X''` means italic _X_. We need to use monospace (just like in markdown) syntax instead.

vkostyukov:


core/src/main/scala/io/finch/request/package.scala
```diff
@@ -198,14 +202,15 @@ package object request extends LowPriorityImplicits {
   object RequiredParam {
 
     /**
-     * Creates a ''RequestReader'' that reads a required string ''param''
-     * from the request or raises an exception when the param is missing or empty.
+     * Creates a [[RequestReader]] that reads a required string ''param'' from the request or raises a [[NotPresent]]
```
jenshalm: Does this syntax now work in recent Scala versions? Older versions required the use of FQN with the bracket syntax.

vkostyukov: It definitely works for 2.11 (just checked it). So, we're Ok with the web-site. Although, it doesn't work inside Intellij, so I will fix the docs to make them work everywhere. Thanks for pointing this!

PR #183, opened 2015-02-08
Restructure documentation for request package - fixes #178

### LINKED ISSUE #178: Rewrite section about requests in the documentation
The PRs for 0.5.0 have changed some areas of the request package in fairly significant ways. For the features I contributed I added documentation mostly in the form of "patches", leading to a format that is no longer that well organized. 

Here I'll propose a new structure for this section and would also be able to work on this over the weekend, so that we have it ready for the 0.5.0 release.

* __Requests__
    * __Custom Request Types__ -> keep this unchanged
    * __Request Reader__ -> mini intro - text from the old docs 
        * __Overview__ -> a new section with short code example, a short bullet list, less than what fits on a laptop screen, summarizing all features of the reader API with links to their sections
        * __API__ -> roughly like in the old introduction, but adding relevant new methods
    * __Base Readers__ -> Summary, maybe in table form, of the now only 12 base readers
        * __Required and Optional Readers__ -> Details about their behaviour
        * __Multi-Value Parameters__ -> Probably the only base reader that needs additional explanation, mostly content from the existing docs
        * __Custom Readers__ -> Mention factory methods for creating your own readers
    * __Combining and Reusing Readers__ -> Short example how one reader reads two params and is then combined with a reader for other request items
        * __Applicative Syntax__ -> Explain the advantages of error handling and how multiple errors are returned
        * __Monadic Syntax__ -> Mention how, when and why one might want to combine readers with monadic syntax instead of applicative syntax
    * __Type Conversion__ -> Explain `as[A]` method and the types of readers it is available for
        * __Built-in Decoders__ -> Short list of the available decoders for numbers and booleans
        * __Custom Decoders__ -> Short example for how to write your own decoder
        * __Integration with JSON Libraries__ -> Explain how JSON decoders are made available for use with `as[A]`
    * __Validation__ -> Very short introduction, describe behaviour in context of required and optional readers
        * __Inline Validation__ -> Short example/explanation
        * __Reusable Rules__ -> Short example/explanation
    
One particular thing is up for discussion. I would recommend to show most examples in applicative syntax now, due to the superior error handling. We would then only show the monadic syntax in the section dedicated to this feature. If you'd like to keep the current preference for monadic syntax in the manual, please let me know about your reasoning.

Also any suggestion about the structure and headlines is welcome.

jenshalm:
This is an almost complete rewrite of the section about request processing as outlined in this ticket: https://github.com/finagle/finch/issues/178.

This is probably difficult to review based on the diff, you can alternatively read the rendered version.

### DIFF
--- a/docs.md
+++ b/docs.md
@@ -4,13 +4,25 @@
 * [Endpoints](docs.md#endpoints)
 * [Requests](docs.md#requests)
   * [Custom Request Types](docs.md#custom-request-types)
-  * [Request Reader](docs.md#request-reader-monad)
-  * [Query String Params](docs.md#query-string-params)
-  * [Improved Error Handling with Applicative Syntax](docs.md#improved-error-handling-with-applicative-syntax)
+  * [Request Reader](docs.md#request-reader) 
+    * [Overview](docs.md#overview)
+    * [API](docs.md#api)
+  * [Base Readers](docs.md#base-readers)
+    * [Required and Optional Readers](docs.md#required-and-optional-readers)
+    * [Multi-Value Parameters](docs.md#multi-value-parameters)
+    * [Custom Readers](docs.md#custom-readers)
+    * [Error Handling](docs.md#error-handling)
+  * [Combining and Reusing Readers](docs.md#combining-and-reusing-readers)
+    * [Applicative Syntax](docs.md#applicative-syntax)
+    * [Monadic Syntax](docs.md#monadic-syntax)
+  * [Type Conversion](docs.md#type-conversion)
+    * [Built-in Decoders](docs.md#built-in-decoders)
+    * [Custom Decoders](docs.md#custom-decoders)
+    * [Integration with JSON Libraries](docs.md#integration-with-json-libraries)
   * [Validation](docs.md#validation)
-  * [Multiple-Value Params](docs.md#multiple-value-params)
-  * [HTTP Headers](docs.md#http-headers)
-  * [HTTP Bodies](docs.md#http-bodies)
+    * [Inline Validation](docs.md#inline-validation)
+    * [Reusable Rules](docs.md#reusable-validators)
+    * [Built-in Rules](docs.md#built-in-rules)
 * [Responses](docs.md#responses)
   * [Response Builder](docs.md#response-builder)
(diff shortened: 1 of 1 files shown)


docs.md
```diff
@@ -98,249 +110,443 @@ val s = RequiredParam("foo")(req)
...
+A typical `RequestReader` might look like this:
+
+```scala
+import io.finch.request._
+
+case class User(name: String, age: Int, city: String)
+    
+val user: RequestReader[User] = (
+  RequiredParam("name") ~
+  RequiredParam("age").as[Int].shouldNot("be less than 18"){ _ < 18 } ~
```
vkostyukov: We can use predefined rule `beLessThan(18)` here. Do you mind also rewriting it in DSL-style:  

```scala
RequiredParam("age").as[Int] shouldNot beLessThan(18) ~
```

docs.md
```diff
@@ -98,249 +110,443 @@ val s = RequiredParam("foo")(req)
...
+and combine them to build new readers:
 
 ```scala
-case class User(name: String, age: Int, city: String)
+case class Address(street: String, city: String, postCode: String)
+
+val address: RequestReader[Address] = (
+  RequiredParam("street") ~
+  RequiredParam("city") ~
+  RequiredParam("postCode").shouldNot("be longer than 5 characters"){ _ > 5 }
```
vkostyukov: We can use predefined rule here: `shouldNot beLongerThan(5)`.

vkostyukov:
This is amazing! Thanks a lot @jenshalm for this hard work! I will merge this once you're happy with the state of this PR.

docs.md
jenshalm: I'l switch to the predefined rule. But the DSL-style does not compile without adding more parentheses, so I think this syntax is clearer.

jenshalm:
I switched to use built-in rules in the two example you commented. I also added a small section mentioning these built-in rules. The only downside is that our examples for custom rules now still mostly show logic you can do with built-in rules. I did not have a quick idea how to rewrite the samples with something not supported out of the box without making it very contrived. So in my view this is good to be merged now and we can rethink those samples later.

vkostyukov:
Super cool! I'm merging this.

PR #184, opened 2015-02-09
Allow a RequestReader to be used as a Finagle Service

jenshalm:
In this case I might be missing something obvious which makes this PR obsolete. But I was wondering about the ways to combine a reader with Finagle Services and found examples like this in the documentation:

```scala
val user: RequestReader[User] = for {
  name <- RequiredParam("name")
  age <- RequiredParam("age").as[Int]
  city <- OptionalParam("city")
} yield User(name, age, city.getOrElse("Novosibirsk"))

val service = new Service[HttpRequest, Json] {
  def apply(req: HttpRequest) = for {
    u <- user(req)
  } yield Json.obj(
    "name" -> u.name,
    "age" -> u.age,
    "city" -> u.city
  )
}
```

This shows the use of a reader **inside** a service implementation. This is convenient for small and simple services, but for a bigger application I might prefer to have the reader itself as a pluggable service:

```scala
case class MyRequest(request: HttpRequest)

val userReader: RequestReader[User] = ???

val userService: Service[MyRequest, User] = userReader.asService[MyRequest]

val profileService: Service[User, UserProfile] = ???

userService ! profileService ! TurnIntoHttp[UserProfile]
```

The reader becomes usable as a Finagle service through "fixing" the request type it can handle.
This way, apart from using the reader inside a service implementation, it can also be composed
with existing services that only deal with domain objects and not with anything HTTP.

This PR implements the `asService[Req]` method shown above.

Is this a bad idea? Can the same be achieved by other means?

If this PR is seen as useful it would be cool to squeeze it into 0.5.0.

### DIFF
--- a/core/src/test/scala/io/finch/request/ReaderAsServiceSpec.scala
+++ b/core/src/test/scala/io/finch/request/ReaderAsServiceSpec.scala
@@ -1,0 +1,46 @@
+/*
+ * Copyright 2014, by Vladimir Kostyukov and Contributors.
+ *
+ * This file is a part of a Finch library that may be found at
+ *
+ *      https://github.com/finagle/finch
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * You may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * Contributor(s):
+ * Jens Halm
+ */
+package io.finch.request
+
+import org.scalatest.{Matchers, FlatSpec}
+import com.twitter.util.{Future, Await}
+import com.twitter.finagle.httpx.Request
+import com.twitter.finagle.Service
+import io.finch.HttpRequest
+
+class ReaderAsServiceSpec extends FlatSpec with Matchers {
+
+  
+  "A RequestReader" should "be usable as a Finagle Service" in {
+    case class MyRequest(request: HttpRequest)
+    
+    implicit def extractRequest(req: MyRequest): HttpRequest = req.request
+    
+    val request: MyRequest = MyRequest(Request(("foo", "5")))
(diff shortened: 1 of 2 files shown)


core/src/test/scala/io/finch/request/ReaderAsServiceSpec.scala
```diff
@@ -0,0 +1,46 @@
...
+
+  
+  "A RequestReader" should "be usable as a Finagle Service" in {
+    case class MyRequest(request: HttpRequest)
+    
+    implicit def extractRequest(req: MyRequest): HttpRequest = req.request
+    
+    val request: MyRequest = MyRequest(Request(("foo", "5")))
+    val reader: RequestReader[Int] = RequiredParam("foo").as[Int]
+    val service: Service[MyRequest, Int] = reader.asService[MyRequest]
```
vkostyukov: Do you think it's possible to use an implicit conversion here instead of calling the `aService` method explicitly?

jenshalm: Where/how would you specify the `MyRequest` type in this case?

(I am afk now for 8hrs).

vkostyukov: We have [a similar conversion|https://github.com/finagle/finch/blob/master/core/src/main/scala/io/finch/package.scala#L203] for `Future[Rep] -> Service[Req, Rep]`. So, the compile will figure out the `Req` type in most of the cases. In our case, we have an explicit type `Service[MyRequest, Int]`. In case of service composition via `!` it would definitely work when `reader` is not the first service in the chain. We should probably experiment with it in tests.

vkostyukov:
I generally like the idea. I think we can merge this into 0.5.0. I was just wondering if it's possible to use an implicit conversion here.

jenshalm:
Your question about implicits made me think about it. I'm not sure the idea is mature enough, so I would not mind if you want to keep it open for 0.6.0. The reason is that, unless I'm overlooking something, I think `asService[Req]` currently only makes sense if `Req` is not a custom type. In which case it is often the first service, so that we cannot easily use an implicit.

But I would not know how to use `asService` with a custom request type. Because then the reader would potentially spit out some domain object and the additional information of the custom request is lost for the subsequent service. And while it is possible to use custom request objects with readers, there currently is no easy way to access any information from the custom request inside the reader. 

This might be something where we need to wait for the most elegant usage patterns to emerge.

vkostyukov:
Yeah. I see. Let's postpone this for the next release then.

jenshalm:
Closing this as these ideas are superseded by the new direction.

PR #189, opened 2015-02-13
Fix code samples in documentation that do not compile

jenshalm:
While it might look nice with spaces, it really does not compile.


vkostyukov:
My bad. Thanks @jenshalm!

PR #198, opened 2015-02-17
Add missing Finagle import to example

haakonn:
Very minor thing, but would help Finch beginners (like me) get started: Httpx isn't imported in the README.md example. I needed to find this class on my own before the example would work. This PR adds an import for it.


vkostyukov:
Nice! Thanks @haakonn!

Issue #197, opened 2015-02-17
Multipart/form-data support for file uploading


jenshaase:
Hi,

If I'm correct Finagle currently doesn't support `multipart/form-data` out of the box. However, Finatra (http://finatra.info/docs/index.html#uploads) has build in support for it.

It would be nice if there is the possibility to upload a file within finch.

Cheers,
Jens

jenshaase:
This code may be a solution (adopted from the finatra code):

~~~
  import com.twitter.finagle.httpx.netty.Bijections._
  import scala.collection.JavaConversions._
  import org.jboss.netty.handler.codec.http.multipart.{HttpPostRequestDecoder, MixedAttribute, MixedFileUpload}

  object OptionalMultipartParam {
    def apply(param: String): RequestReader[Option[Array[Byte]]] =
      RequestReader(items.ParamItem(param)) { req =>
        val dec = new HttpPostRequestDecoder(from(req))
        dec.getBodyHttpDatas.find(_.getName == param).flatMap{
          case data: MixedFileUpload => Some(data.get)
          case data: MixedAttribute => Some(data.get)
          case _ => None
        }
      }
  }

  object RequiredMultipartParam {
    def apply(param: String) =
      OptionalMultipartParam(param).failIfEmpty.shouldNot("be empty")(_.isEmpty)
  }
~~~

vkostyukov:
Interesting. Thanks @jenshaase for failing a ticket. Do you mind to create a pull-request? It would be fair if you implement this since it's your idea/code. This is an amazing feature.

jenshaase:
Ok. I can submit a pull request in the next days.

vkostyukov:
Cool! Thanks again!

vkostyukov:
Fixed by #201.

PR #201, opened 2015-02-20
Added support for multipart/from-data requests

jenshaase:
see #197


vkostyukov:
Amazing job @jenshaase!  Just one idea from my side. Sorry for not bringing it up earlier. What to you think about making `XMiltipartParam` an additional case of just `Param`? `XParam` reader is already making some magic: it searches for the param in both query-string (GET params) and body (url-encoded POST params) - this done by Finagle. We can add a third case here - `miltipart` content type and make magic on the Finch side. So, users can use param readers no matter how params are actually passed. In this case I would suggest to rename `XMultipartFile` to just `XUpload` or `XFileUpload`. Thus, the following reader should successfully read the test case:

```scala
val r: RequestReader[String ~ FileUpload] = RequiredParam("type") ~ RequiredUpload("groups")
```

What do you think? //c @jenshalm @rpless

jenshaase:
Good idea. I can change this.

jenshalm:
Yes, agree, nice simplification.

vkostyukov:
Just one thing. I don't really like depending on Netty type `FileUpload` here. I see the following ways to decouple it:

* Create a case class `io.finch.request.FileUpload` with similar functionality
* Just read `Array[Body]` instead
* Have several readers for this: `Upload => Array[Byte]`, `UploadContentType => UploadName => String`

Not sure which solution is better.

jenshalm:
Solution 2) reduces info, 3) is not convenient. If Netty API really has to be avoided, I'd be in favour of wrapping it then.

jenshaase:
I would also recommend to write a wrapper. Returning an Byte-Array is very problematic, because it can result in an `OutOfMemoryError` if the file is too large.

```
class FileUpload(underlying: org.jboss.netty.handler.codec.http.multipart.FileUpload) {
  val contentType = underlying.getContentType
  val filename = underlying.getFilename
  lazy val bytes = underlying.get
  def renameTo(f: File) = underlying.renameTo(f)

  // other methods and values: http://netty.io/4.0/api/io/netty/handler/codec/http/multipart/FileUpload.html
}
```

vkostyukov:
Let's go with wrapper. @jenshaase do you think we can make it a case class with just 3 properties: 

```scala
case class FileUpload(contentType: String, fileName: String, bytes: Array[Byte])
```

I mean, we probably don't need to inherit all the Netty methods here.

jenshalm:
Like the other Jens already mentioned, we might want to avoid forcing to materialize the bytes. His class looks better to me.

jenshaase:
Some background information on Netty. Netty has basically two implemented classes of `FileUpload`. `MemoryFileUpload` for small files that can be stored in Memory and `DiskFileUpload` for larger files that can not be stored in memory. Therefore I also would like to have a method `renameTo`.

vkostyukov:
Alright. In this case let's just use raw Netty type for now, since otherwise we're going to repeat it's functionality in our wrapper. Let's just make a type alias for now in `io.finch.request`:

```
type FileUpload = org.jboss.netty.handler.codec.http.multipart.FileUpload
```

jenshaase:
Ok nice. I will update this pull request in the next days.

vkostyukov:
Thanks a lot @jenshaase! Appreciate your time on this!

jenshaase:
Updated the pull request, however it currently has conflicts. Should I merge the code or will you do it?

core/src/main/scala/io/finch/request/package.scala
```diff
@@ -261,7 +271,20 @@ package object request extends LowPriorityImplicits {
      *
      * @return an `Option` that contains a param value or `None` if the param is empty
      */
-    def apply(param: String): RequestReader[Option[String]] = RequestReader(ParamItem(param))(_.params.get(param))
+    def apply(param: String): RequestReader[Option[String]] =
+      RequestReader(ParamItem(param)){ req =>
+        // Finagle doesn't handle multipart parameter by default, so
+        // we need to add it here
+        if (req.method == Method.Post && HttpPostRequestDecoder.isMultipart(from(req))) {
```
vkostyukov: What do you think about making query-string params a default case here? In 99% cases we will end up fetching the query-string/url-ecoded param but with this implementation we always check whether or not it's multipart. I would simplify this:

```scala
RequestReader(ParamItem(param)) { req =>
  req.params.get(param).orElse(
    val decoder = new HttpPostRequestDecoder(from(req)) 
    decoder.getBodyHttpDatas.asScala.find(_.getName == param) flatMap {
      case attr: Attribute => Some(attr.getValue)
      case _ => None
    }
  )
}
```

I can make these changes while rebasing the PR if you don't mind.

jenshaase: Looks good to me. Feel free to change this while rebasing the PR.

vkostyukov:
Thanks @jenshaase! Just merged it.

Issue #204, opened 2015-02-23
Finch in Action


vkostyukov:
This is a big ticket that related to #190, #172 and PR #184. For now we have amazing and highly-composable basic blocks but it's still not really clear how to use them together. One and simple example is shown in the demo project, although I was wondering if we could do better.

There is nothing wrong with using Finch in a straightforward way (via `Service[HttpRequest, HttpResponse]`). But we also should think about using it in a more convenient manner. The PR #184 is a good example of this effort. My hight-level idea is to provide users a way to write their microservices as functions:

```scala
def getUser(userId: Long): Future[User] = ???
def getUsers(fromId: Int, toId: Int): Future[Seq[User]] = ???
```

and then use Finch basic blocks to serve those functions on Finagle. It might look as follows:

```scala
val e1: Endpoint[User] = Get / "users" / int /> getUser
val e2: Endpoint[Seq[User]] Get / "users" /> { fetchFromAndTo map getUsers }
```

We don't really need to convert request reader into service (per #184) or create an explicit service instance. We do have everything is needed in `RequestReader` so we can just map it:

```scala
val pagination: RequestReader[Int ~ Int] = ???
val users: RequestReader[Seq[User]] = pagination map { case from ~ to => getUsers(from, to) }
```

Thus, we may define an `Endpoint` as `type Endpoint[+A] = Router[RequestReader[A]]`. Therefore, we're replacing `Service` with `RequestReader`, which sounds reasonable for me:

* They look similar: `Req => Future[A]`
* `RequestReader`s are composable as hell (we did a great job), `Service`s are not (`!` is an ugly hack).

vkostyukov:
Would be nice to hear others opinion on this direction. /cc @BenWhitehead, @jenshalm, @rpless, @rodrigopr, @benjumanji.

benjumanji:
This seems like a nice idea. I was just writing some code (not yet finch'd) that basically involved a bunch of filters to turn `Service[Request, Respose]` to `Service[C,D]`. This kind of thing would make that pretty pleasant, so once we have an impl, I have some code I can test this on.

vkostyukov:
Although, it's not clear yet how to fetch the details from a custom request type:

```scala
case class MyReq(http: HttpRequest, currentUserId: Long)
```

vkostyukov:
[...]

PR #206, opened 2015-02-26
Finch in Action

vkostyukov:
**Read a Gist about it**: https://gist.github.com/vkostyukov/e0e952c28b87563b2383

This PR is an attempt to address #204.

The work is in progress but I feel that direction is correct. The hight-level idea is to:
* Allow users write their services using custom domain types, i.e., `(Int, String) => User` instead of `HttpRequest => HttpResponse`.
* Hide explicit `HttpRequest` and `HttpResponse` types. Ideally users should not deal with them directly.
* Use `RequestReader` (`Micro`) instead of `Service`.
* Enable implicit conversion from `A` to `HttpResponse` 

I just realized that `RequestReader` is actually a `Service` or `Microservice` (`Micro`) but more powerful. Like a service it takes a request and transforms it into something. For example, request reader (or microservice) `RequiredParam("a")` takes request and transforms it into a string value. So, we can consider `RequestReader` as a building block for users services: we use bunch of predefined micorservices to build powerful Finagle services. For that propose I added a magic `~>` compositor that works as follows:

```scala
def magic(a: String, b: String): Future[String] = ???
val a: Micro[String] = RequiredParam("a") ~ RequiredParam("b") ~> magic
```

This  idea is originally came from Jens.

I was also trying to enable an implicit conversion from `A` to `HttpResponse` if there is an implicit `EncodeResponse` available in the scope.

This is a long-standing PR that will probably lead to some refactoring but I want to have it published so I can get some feedback on this direction. I'm personally super excited about treating the request reader as microservice. So the Finch application is just a bunch of composed microservices (request readers).

The resulting API looks as follows (`Micro` is a `RequestReader`):

```scala
def echo(what: String): Micro[String] = Micro.value(what)
val e: Endpoint[String] = Get / "echo" / string /> echo
Httpx.serve("8081", e) // string will be implictly converted into plain-text HTTP response
```


vkostyukov:
For now I'm really happy with the following:

* Magic of both `~>` and `/>` compositors
* Using `RequestReader` instead of `Service`, as a result hiding the `HttpRequest` type
* Downgrade the `Endpoint` to carry just one type - output type, which actually matters

But, I don't like the way I solved the last problem: hiding the `HttpResponse` type and composing the endpoints of different types. Given the following endpoints:

```scala
val e1: Endpoint[A] = ???
val e2: Endpoint[B] = ???
val e3: Endpoint[C] = ???
```

The problem is how to compose them together. For now, I solved it in an ugly way, using the type `type Endopints = Seq[Endpoint[HttpResponse]]` that explicitly enables implicit conversions:

```scala
val e: Endpoints = Seq(e1, e2, e3)
```

I was wondering if there is another way to solve it and found out that I've already seen a similar problem. And more importantly, I've seen an elegant solution. The scodec library uses Shapeless `HList` to convert list of codecs (i.e., `Codec[A] :: Codec[B] :: Codec[C]`) into a single `Codec[X]`. This is exactly what I'm trying to do. Thus, the idea is to have an `HList` of endpoints and then magically convert it into an `Endpoint[HttpResponse]` (which is actually a `Service[HttpRequest, HttpResponse]`) if there is an implicit view `X => HttpResponse` available for each type `A`, `B`, etc.

This sounds like a solid approach. And I'm totally fine with introducing dependency to Shapeless, since it will bring us to an awesome solution.

That said, this is my plan for weekend.

vkostyukov:
Was playing around `HList` and realized that we might better use it instead of `~` and `/` types to handle composite entities. In case of endpoint, it doesn't really make sense to curry all the types of endpoints. I decided to solve it in a different and much simpler way (via `!` compositor). 

Still looking for a good solution and would be happy to hear feedback.

jenshalm:
This looks like a good direction. I'm happy to see that pattern matching gets avoided through implicits now and not a new `Micro` trait. I analyzed the impl and the code samples and I don't have any concerns.  

Somehow I'm still not in love with the name `Micro`, being a type alias for `RequestReader` even though the two somehow sound like two distinct things. But this is just semantic nit-pickyness and I don't have a counter-proposal either.

I'd still have two questions:

* What is this new `!` compositor for routes about?

* What would be the most elegant way to apply a certain reader (like for auth) for each route? Is there anything apart from the repetition of using an `auth` reader in each `Micro` then? Apart from the plumbing it would introduce the risk of forgetting it for just one route (e.g. if you don't need the `User` object, but still require the request to be authenticated).

vkostyukov:
Super cool questions, as usual.

Probably, `Micro` isn't a best name. But it's definitely better than `RequestReader`, since it's no longer a `RequestReader` but `RequestHandler` or `RequestProcessor`. The only thing I like about `Micro` is that it's a short name :) I would be happy to change it into something more reasonable.

Here is the problem I'm trying to solve with `!` combinator. Let's say we have two endpoints:

```scala
val i: Endpoint[Int] = Get / "i" /> Micro.value(10)
val s: Endpoint[String] = Get / "s" /> Micro.value("10")
```

The question here is how to compose them together and what's the result type should be. I'm still thinking how to handle this and I'm not sure I'm happy with `!` compositor here, which is a _forced or_ it converts `Int` and `String` into `HttpResponse` if there are both `EncodeResonse[A]`s (for string and int) available in the scope.

Regarding the `auth` request reader. I can't really see a way to do it w/o filters. So, since the `RequestReader` still takes a request of any type we can think about having a _classic_ `Auth` filter that changes a request type to `AuthReq` and then having a factory method like this:

```scala
object RequestReader {
  def apply[Req, A](f: Req => A)(implicit ev: Req => HttpRequest): RequestReader[A] = ???
}
```

We may define an `auth` request reader like this:

```scala
case class AuthReq(userId: Int, http: HttpRequest)
val auth: RequestReader[Int] = RequestReader[AuthReq, Int](_.userId)
```

So combine both, an `Auth` filter that actually performs authorization and `auth` request reader that fetches the auth details from the custom request type. This will guarantee us that all the services will be authorized by Finagle filter.

@jenshalm, how does it sound?

vkostyukov:
Just realized that it's actually not possible to define this factory method. Hopefully, there is another way to  make a reader that reads from a custom request type.

vkostyukov:
Thus, I still have to solve to problems:

1. How to compose `Endpoint[A]` and `Endpoint[B]` together (and what is a result type).
2. How to apply auth filter globally.

ImLiar:
@vkostyukov 
Are u sure about authentication in front of routes?
We successfully use lazy auth after routing part done it's work. It allows to avoid request url checks in filter and start authentication process only when you need it (in specific endpoints)

vkostyukov:
Yeah. I also think that having a `auth` done after routing is a nice way to control the things. Although, it also may cause a bug when you simply forget to put `auth` before service that doesn't requires any authorization information (i.e., `userId`).

BenWhitehead:
@ImLiar @vkostyukov I think post routing auth can work in some scenarios, but for us our API is all auth controlled so it makes a lot more sense for use to auth pre-route.

vkostyukov:
Right. I'm thinking about supporting both ways somehow.

BenWhitehead:
Great! Thank you.

vkostyukov:
Simplified the implementation quite heavily based on a idea that `Endpoint` is `Router[RequestReader[HttpResponse]]`. This solves the first problem I've been working on. The second problem (pre routing auth) still in progress.

vkostyukov:
I think this is ready to go alive. I spend a month trying to approach this problem and ended up solving it in the most straightforward way. I'm pretty happy with the current solution. It still requires some stylistic changes and docs but for now this code might be merged in.

Anyway. I'm totally happy to revisit this problem later.

PR #210, opened 2015-03-14
Added first test to demo

suls:
Adding a first simple test to the `demo` app.

To achieve this I had to refactor `App` slightly. It is now broken up in to run-the-thing stuff in `App` and wire-the-things-together in `Demo`. This was necessary due to the `DelayedInit` mixin from `scala.App`.

I also got rid of one wartremover warning by assigning `Await.ready(..` thus returning `Unit` as needed.

### DIFF
--- a/demo/src/test/scala/io/finch/demo/DemoSpec.scala
+++ b/demo/src/test/scala/io/finch/demo/DemoSpec.scala
@@ -1,0 +1,112 @@
+package io.finch.demo
+
+import java.net.InetSocketAddress
+import java.nio.charset.Charset
+
+import com.twitter.finagle.Service
+import com.twitter.finagle.httpx.Request
+import com.twitter.util.Await
+import io.finch._
+import org.jboss.netty.buffer.ChannelBuffers
+import org.jboss.netty.handler.codec.http.{DefaultHttpRequest, HttpMethod, HttpVersion}
+import org.scalatest._
+
+class DemoSpec extends FlatSpec with Matchers {
+
+  val await = mkAwait(Demo.backend)
+
+  "The demo" should "deny unauthorized requests" in {
+    await(Request()).status shouldBe io.finch.response.Unauthorized().status
+  }
+
+  it should "allow authorized requests for /users" in {
+    val req = GET("/users")
+
+    await(req).status shouldBe io.finch.response.Ok().status
+  }
+
+  it should "allow posting authorized requests" in {
+    val req = POST("/users?name=foobar")
+
+    await(req).status shouldBe io.finch.response.Ok().status
+  }
+
+  it should "allow posting authorized ticket requests" in {
+    val req = POST("/users/0/tickets", "{\"label\": \"foobar\"}")
+
+    await(req).status shouldBe io.finch.response.Ok().status
+  }
+
+  it should "trigger BadRequest when user not found" in {
+    val req = GET("/users/9879")
+
+    await(req).status shouldBe io.finch.response.BadRequest.status
+  }
+
+  it should "raise NotFound for routes not found" in {
+    val req = GET("/foo/bar")
+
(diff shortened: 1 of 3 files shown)


suls:
Apologies for not following up earlier. In my last change I have removed the `*demo*` exclusion rule from the test-coverage - thus the 10% drop. I am currently working on increasing it again - and so far it is looking pretty good.. Please have some more patience ;)

suls:
I am having difficulties trying to hit https://coveralls.io/builds/2173234/source?filename=demo%2Fsrc%2Fmain%2Fscala%2Fio%2Ffinch%2Fdemo%2FMain.scala#L76 and https://coveralls.io/builds/2173234/source?filename=demo%2Fsrc%2Fmain%2Fscala%2Fio%2Ffinch%2Fdemo%2FMain.scala#L84 .. any hints are more than welcome!

rpless:
You should be able to get a `NotParsed` by violating the ValidationRule here: 
https://github.com/finagle/finch/blob/master/demo/src/main/scala/io/finch/demo/reader.scala#L35

I think you can get a `NotValid` by passing invalid Json here:
https://github.com/finagle/finch/blob/master/demo/src/main/scala/io/finch/demo/reader.scala#L42

vkostyukov:
Nice. It's green!

I wanted to wait for #204 (demo will be rewritten as well as it's tests) but I think it's a good idea to merge it know and revisit latter.

demo/src/test/scala/io/finch/demo/DemoSpec.scala
```diff
@@ -0,0 +1,111 @@
...
+
+    await(req).status shouldBe io.finch.response.BadRequest().status
+  }
+
+  def mkAwait(service: Service[HttpRequest, HttpResponse]) : Request => HttpResponse =
+    (req) => Await.result(service(req))
+
+
+  def POST(path: String, body: Option[String]) : Request = {
+    val r = new DefaultHttpRequest(HttpVersion.HTTP_1_1, HttpMethod.POST, path)
```
vkostyukov: Are you sure we can't use `httpx.Request` companion object to make a request?

demo/src/test/scala/io/finch/demo/DemoSpec.scala
```diff
@@ -0,0 +1,111 @@
...
+      r.headers().add("content-length", buf.readableBytes())
+    }
+
+    new Request {
+      val httpRequest = r
+      lazy val remoteSocketAddress = new InetSocketAddress(0)
+    }
+  }
+
+  def POST(path: String, body: String) : Request = POST(path, Option(body))
```
vkostyukov: The space before `: Request` is redundant. Actually, I surprised scalastyle didn't notice it.

demo/src/test/scala/io/finch/demo/DemoSpec.scala
suls: No, we can't as it sadly doesn't support adding headers .. see https://github.com/twitter/finagle/blob/master/finagle-httpx/src/main/scala/com/twitter/finagle/httpx/Request.scala#L185-L199

Maybe you have an efficient way of adopting the API in that project? :+1:

demo/src/test/scala/io/finch/demo/DemoSpec.scala
suls: Noted:

- [ ] fix formatting

demo/src/test/scala/io/finch/demo/DemoSpec.scala
vkostyukov: Yeah. That's what we're trying to do in the "next" HTTP implementation. Hopefully, it will get better soon.

vkostyukov:
I think this is ready to go live. Thanks @suls!

PR #211, opened 2015-03-22
Partially Fix #200: fixes 'Inferred type containing Nothing' warnings in RequestReader

### LINKED ISSUE #200: Fix warnings from Wartremover
We enabled Wartremoved in #199. For now it produces around 100 warnings. It would be nice if we can fix them and change the level from warning to error to prevent new things.

This is a perfect issue for newcomers. It's not necessary to fix all the warnings in one PR. The best idea would be to fix them file-by-file.

nicmart:


vkostyukov:
This is cool! Thanks @nicmart!

PR #212, opened 2015-03-26
Fix #187: Reasonable RequestReaders names

### LINKED ISSUE #187: Unify request reader names
I was thinking about simplifying/unifying the request reader names. We can make readers optional by default and name them as `param`, `header`, etc. And then provide a way to make reader required. There is already a method `failIfEmpy`, which does exactly what is needed. But we can come up with a better name. I was thinking about just function `required`:

```scala
val a: RequestReader[String] = required param("a")
val b: RequestReader[Option[String]] = header("b")
```

vkostyukov:
This change involves the following:
* `RequiredX` is now `x`, eg. `param("a")`
* `OptionalX` is now `xOption`, eg. `headerOption("X-Secret")`

This naming convention is heavily inspired by Scala standard library, i.e., `head` and `headOption`.

The only one thing doesn't fit well into this approach: `OptionalParams` that return `Seq[String]`. For now, I called it `params0()`. Would be happy to hear feedback and ideas on how can we name optional params.

/cc @jenshalm, @benjumanji, @rpless, @BenWhitehead

### DIFF
--- a/core/src/main/scala/io/finch/request/package.scala
+++ b/core/src/main/scala/io/finch/request/package.scala
@@ -51,10 +51,8 @@ import scala.reflect.ClassTag
  * {{{
  *   case class Complex(r: Double, i: Double)
  *   val complex: RequestReader[Complex] =
- *     RequiredParam("real").as[Double] ~
- *     OptionalParam("imaginary").as[Double] map {
- *       case r ~ i => Complex(r, i.getOrElse(0.0))
- *     }
+ *     param("real").as[Double] ~
+ *     paramOption("imaginary").as[Double].withDefault(0.0) ~> Complex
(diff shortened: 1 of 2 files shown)


BenWhitehead:
Overall I think it's a reasonable change and the actual code change seems minimal.

Not crazy about `params0()` but I also can't come up with something better right now. I'll be sure to add a comment here if I come up with anything.

vkostyukov:
For I decided to go with:
* `RequiredParams` -> `paramsNonEmpty`
* `OptionalParams` -> `params`

vkostyukov:
Merging this for now. I think the current naming schema is pretty simple and fits well into a Finch infrastructure.

Issue #186, opened 2015-02-11
Default value for optional readers


vkostyukov:
I was wondering if it might be useful. The idea is to instead of

```scala
val p: RequestReader[String] = OptionalParam("p") map { _.getOrElse("n/a") }
```

to allow users write something like this

```scala
val p: RequestReader[String] = OptionalParam("p") withDefault "n/a"
```

We can also use `orElse` instead of `withDefault`.

vkostyukov:
I'm putting this into the next milestone.

jenshalm:
I think that would be quite nice to have. If the value passed to the function is a simple value, `withDefault` works better as a name, as the expected signature for an `orElse` method would have an `Option[T]` param instead. For maximum flexibility you could have both:

```scala
def withDefault[B >: A](default: ⇒ B): RequestReader[Option[B]]
def orElse[B >: A](alternative: ⇒ Option[B]): RequestReader[Option[B]]
```

vkostyukov:
Nice! Let's plan this for the 0.6.0 release.

vkostyukov:
Fixed in #186.

PR #214, opened 2015-03-26
Ench #186: Default value for optional readers

ImLiar:
Simple implementation based on scala `Option` methods


ImLiar:
Oh wow... Some bugs with coveralls

vkostyukov:
Yep. Sometimes this happens. Still don't have time to dig in and find out why this happens /cc: @rpless.

The PR looks cool! Thanks a lot!

vkostyukov:
BTW, @ImLiar do you use Finch in production? If so do you mind adding you company into the adopters lists?

ImLiar:
@vkostyukov we are still at very early development stage, but yeah, I remember about adopters list :smile:

PR #216, opened 2015-03-29
Update SBT version to 0.13.8 and clean up settings syntax

travisbrown:
This is a trivial change but it's kind of cool that you don't need `: _*` for settings now.


vkostyukov:
Thanks @travisbrown! This is awesome!

PR #225, opened 2015-04-06
Add Spright to adopters list

keeth:
Hello, we use finch in production, in several backend microservices - we love it.  We just launched a couple weeks ago.  Feel free to add us to your adopters list.  Thanks!!

### DIFF
--- a/README.md
+++ b/README.md
@@ -72,6 +72,7 @@ Adopters
 * [Konfettin](http://konfettin.ru)
 * [JusBrasil](http://www.jusbrasil.com.br)
 * [Sabre Labs](http://sabrelabs.com)
+* [Spright](http://spright.com)
 * *Submit a pull-request to include your company/project into the list*
 
 Contributing


vkostyukov:
This is awesome @keeth! 

Can you share a bit of details in wich configuration you use Finch (version, modules, etc)? This would be very helpful for me and other contributors in order to understand our priorities.

keeth:
Hi @vkostyukov, sure we are on 0.5.0 and use the core and argonaut modules.  Finch was the deciding factor for using Finagle at all - really like the routing syntax and JSON handling.

vkostyukov:
Nice! Thanks!

There is a nice integration library between Argonaut and Shapeless: https://github.com/alexarchambault/argonaut-shapeless. It will automatically derive any instances of `EncodeJson`/`DecodeJson` for any class. The following code illustrates this magic. 

```scala
import argonaut._, Argonaut._, Shapeless._
import io.finch.argonaut._
import io.finch.response._
import io.finch.request._

case class User(id: Int, name: String)

val rep = Ok(User(1, "Bob"))
val req = body.as[User]
```

Again, thanks a lot for this!

keeth:
Cool, I have heard good things about Shapeless but have not tried it yet, thanks for the tip!

travisbrown:
@keeth We'd love to have a PR adding Spright to [the Finagle adopters list](https://github.com/twitter/finagle/blob/develop/ADOPTERS.md) as well!

PR #231, opened 2015-04-13
updating outdated doc

lderks:
The doc is now updated with the more simple route building feature that is available since 0.6.0


vkostyukov:
Great! Thanks @lderks!

Issue #233, opened 2015-04-15
Remove finch-json library


rpless:
Following up on a twitter conversation from last week, I propose we cease support of the finch-json library. While the API is nice, it has been causing us a lot of issues recently and it depends on deprecated scala APIs. There are plenty of other nice json libraries out there (some of which we support out of the box). I don't think we should be spending our cycles supporting a hand-rolled json library that we tell people not to use in production.

As such I think that we should:
- Deprecate finch-json in 0.7.0
- Delete finch-json in 0.8.0 (1.0 at the latest if we wanted to hold off on this)
- Transition all documentation, examples, and demos to argonaut

penland365:
:+1: to this idea - there is plenty of work to go around focusing on `finch` proper. I'll also throw a :+1: to the idea of using [Argonaut](http://argonaut.io) for the default json library.

travisbrown:
:+1:—I like the idea of a lightweight, dependency-free option, but agree that it's not worth the cost, especially if it's not a viable choice for actual use.

vkostyukov:
:+1: 

Let's deprecate `finch-json` in this release and make a note in the docs regarding the preferred JSON implementation. @rpless do you mind to do that?

rpless:
@vkostyukov sure thing, I should be able to do it this weekend.

BenWhitehead:
:+1: We haven't been using finch-json since 0.1.0 as it was too slow for our needs.

travisbrown:
Now we'll have [automatic API doc links to Argonaut](https://github.com/argonaut-io/argonaut/pull/175) after the next release!

vkostyukov:
@travisbrown This is great! :+1:

vkostyukov:
Fixed by #309.

PR #234, opened 2015-04-18
Deprecate Finch Json.

rpless:
Added deprecation warnings and noted the deprecation in documentation. A start for #233. (should be noted that this adds a lot of warnings to our build).

Do we want to include updates to the demo package in this PR or do we want to do that when we remove the library entirely?

### DIFF
--- a/demo/src/main/scala/io/finch/demo/model.scala
+++ b/demo/src/main/scala/io/finch/demo/model.scala
@@ -22,11 +22,11 @@
 
 package io.finch.demo
 
+import argonaut._, Argonaut._
 import com.twitter.util.Future
 import com.twitter.finagle.Service
 
 import io.finch._
-import io.finch.json._
 
 object model {
 
@@ -35,12 +35,16 @@ object model {
 
   // A ticket object with two fields: `id` and `label`.
   case class Ticket(id: Long, label: String) extends ToJson {
-    override def toJson = Json.obj("id" -> id, "label" -> label)
+    override def toJson = Json.obj("id" := id, "label" := label)
   }
 
   // A user object with three fields: `id`, `name` and a collection of `tickets`.
-  case class User(id: Long, name: String, tickets: Seq[Ticket]) extends ToJson {
-    override def toJson = Json.obj("id" -> id, "name" -> name, "tickets" -> Json.arr(tickets.map(_.toJson) :_*))
+  case class User(id: Long, name: String, tickets: List[Ticket]) extends ToJson {
+    override def toJson = Json(
+      "id" := id,
+      "name" := name,
+      "tickets" := jArray(tickets.map(_.toJson))
+    )
   }
 
   // A helper service that turns a model object into JSON.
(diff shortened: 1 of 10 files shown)


docs/json.md
```diff
@@ -8,6 +8,8 @@
 
 ### Finch-JSON
 
+This library is deprecated in favor of other json libraries.
```
travisbrown: One bit of nitpicking—elsewhere "JSON" is capitalized in prose (also below).

vkostyukov:
Yes. I think we should also update the demo package to use Argonaut JSON. @rpless do you mind to do that in this PR?

travisbrown:
Looks good to me, but can you provide explicit arguments for all `@deprecation` annotations? Using named arguments for these annotations on 2.10 causes weird failures during test compilation (which you can see in the Travis CI results).

docs/json.md
rpless: Good catch. We should keep it consistent.

rpless:
I'll make the updates to demo and fix the 2.10 compiler fails.

rpless:
I've updated the documentation and the failing build. I'm still working on the updates. Its proving more than a little annoying because of package name collisions.

rpless:
I've updated the demo. I tried to change as little as possible, so I used the argonaut json API instead of case class codecs. Thoughts?

demo/src/main/scala/io/finch/demo/model.scala
```diff
@@ -35,12 +35,16 @@ object model {
 
   // A ticket object with two fields: `id` and `label`.
   case class Ticket(id: Long, label: String) extends ToJson {
-    override def toJson = Json.obj("id" -> id, "label" -> label)
+    override def toJson = Json.obj("id" -> jNumber(id), "label" -> jString(label))
```
travisbrown: `Json("id" := id, "label" := label)` would be a better sell.

demo/src/main/scala/io/finch/demo/model.scala
```diff
@@ -35,12 +35,16 @@ object model {
...
   }
 
   // A user object with three fields: `id`, `name` and a collection of `tickets`.
   case class User(id: Long, name: String, tickets: Seq[Ticket]) extends ToJson {
-    override def toJson = Json.obj("id" -> id, "name" -> name, "tickets" -> Json.arr(tickets.map(_.toJson) :_*))
+    override def toJson = Json.obj(
```
travisbrown: Similarly here:

``` scala
    override def toJson = Json(
      "id" := id,
      "name" := name,
      "tickets" := tickets.map(_.toJson).toList
    )
```

Could also change the `Seq` to `List` and skip the `toList`.

demo/src/main/scala/io/finch/demo/model.scala
```diff
@@ -57,7 +61,7 @@ object model {
    */
   implicit class SeqToJson[A](s: Service[A, Seq[ToJson]]) extends Service[A, ToJson] {
     private[demo] def seqToJson(seq: Seq[ToJson]) = new ToJson {
-      def toJson: Json = Json.arr(seq.map { _.toJson }: _*)
+      def toJson: Json = Json.array(seq.map { _.toJson }: _*)
```
travisbrown: I'd probably use `jArray` here but that's more a matter of taste.

travisbrown:
@rpless Thanks! This is looking good to me, but using the `Json("foo" := thingWithACodec)` syntax wherever possible would simplify it a bit further. I've noted a few places above, and would be happy to do a pass in a separate PR.

rpless:
@travisbrown Thanks for the feedback. Making the updates now.

travisbrown:
:+1: from me, thanks!

vkostyukov:
This looks great! Thanks, Ryan, Travis!

PR #236, opened 2015-04-22
Make DecodeRequest invariant and remove DecodeMagnet

travisbrown:
This is a discussion PR—I don't expect it to be a candidate for merging, and before it becomes one it would need some changes to the docs. This may be a discussion that's been had before, and if so we can just close this.

As I've been working on other stuff in Finch (that I promised to finish up last night :frowning:) I've been finding the `DecodeMagnet` stuff a little noisy, and it's not hard to get rid of if you're willing to make `DecodeRequest` invariant (so that you don't get the compiler trying to serve up instances for `Nothing` in weird places).

I'm personally of the opinion that type classes should be invariant by default, and that anything else should take some justification, in part because of the inference and implicit resolution problems we're seeing here.

Making `DecodeRequest` only requires one change to one test, and it's pretty minor. We'd have to explain the decision to users, but I don't think that would be too difficult or undesirable (especially since this change would let us remove the handwaving about `DecodeMagnet`).

Anyone strongly in favor of covariance here?

### DIFF
--- a/core/src/main/scala/io/finch/request/LowPriorityRequestReaderImplicits.scala
+++ b/core/src/main/scala/io/finch/request/LowPriorityRequestReaderImplicits.scala
@@ -17,12 +17,11 @@ trait LowPriorityRequestReaderImplicits {
    * Creates a [[io.finch.request.DecodeMagnet DecodeMagnet]] from
    * [[io.finch.request.DecodeAnyRequest DecodeAnyRequest]].
    */
-  implicit def magnetFromAnyDecode[A](implicit d: DecodeAnyRequest, tag: ClassTag[A]): DecodeMagnet[A] =
-    new DecodeMagnet[A] {
-      def apply(): DecodeRequest[A] = new DecodeRequest[A] {
-        def apply(req: String): Try[A] = d(req)(tag)
-      }
-    }
+  implicit def decodeRequestFromAnyDecode[A](
+    implicit d: DecodeAnyRequest, tag: ClassTag[A]
+  ): DecodeRequest[A] = new DecodeRequest[A] {
+    def apply(req: String): Try[A] = d(req)(tag)
+  }
 
   /**
    * Adds a `~>` and `~~>` compositors to `RequestReader` to compose it with function of one argument.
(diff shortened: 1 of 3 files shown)


rodrigopr:
I agree, removing `DecodeMagnet` will definitely simplify things.
also Argonaut changed to invariant too, which make the last few versions incompatible with finch.

travisbrown:
@rodrigopr I'm pretty sure we can make this decision independently of what Argonaut does—I see that tests are failing if we switch to 6.1-M4, but we should be able to work around that (see for example [my question here](http://stackoverflow.com/q/25770507/334519)—the context is different but the problem is pretty similar, if not identical).

rodrigopr:
Thanks for the link @travisbrown, didn't knew it was possible. :)
I wasn't suggesting to just follow Argonaut though, just point out they made a similar decision.

travisbrown:
@rodrigopr Agreed!—the motivations are basically the same, I think.

vkostyukov:
This is a great simplification! Thanks a lot @travisbrown! I would be happy to get this merged. 

Re the docs, we probably might want to keep them as is for now since they match to the latest available verision 0.6.0. I think it's ok if we address it later.

Re the variance. I'm fine with makiing the `DecodeRequest` invariant. I'm probably not in the cotext, but is there any problems we may face with the Argonaut support having this patch merged?

travisbrown:
@vkostyukov The `argonaut` and `demo` tests pass with this patch, on both the current Argonaut 6.0.4 release and the 6.1-M4 milestone. I'd be curious to hear from someone with a larger `finch-argonaut` codebase, but I don't think there should be problems (except in cases where someone tries to write `body.as[Animal]` and only has a `DecodeRequest[Cat]`, but that's not Argonaut-specific).

Upgrading to 6.1 without this patch would be a little more tricky, but should be possible. The workaround in the response to [my question here](http://stackoverflow.com/q/25770507/334519) seems to fix things.

vkostyukov:
It's clear now. Thanks @travisbrown!

I believe this can be merged then.

PR #237, opened 2015-04-22
Fix the media type of plain text responses

charithe:
The built-in implicit `EncodeResponse` for strings is incorrectly defining the media type to be `plain/text`. This causes some browsers to download the response as a file instead of rendering it on screen. HAProxy is very picky about the media type as well and fails any health check that outputs this value.

This PR fixes the media type to the correct value of `text/plain`.


travisbrown:
I haven't had a chance to look at the content, but the one Travis CI failure is spurious and I've restarted that job.

vkostyukov:
Great! Thanks @charithe! Merging this.

BenWhitehead:
:+1:

PR #240, opened 2015-04-24
Update versions, including Finagle (to 6.25.0)

travisbrown:
Now we just need to publish 0.7.0 so we can update the API docs so that we get linking to Finagle's.


vkostyukov:
Nice! Thanks a lot @travisbrown!

PR #241, opened 2015-04-24
Clean up pom and make root project more useful

travisbrown:
This commit does two or three different things but they're minor enough that I thought I'd just submit them together.

First, the SBT console currently isn't very useful because every command results in "Unused import" warnings. This turns off those warnings in the console by defining the console's `scalacOptions` separately and omitting that warning.

To make the console more useful I've also made the root project depend on `core` and `argonaut` and have added a few initial imports. This means that a user who checks out the repository and types `sbt console` will immediately have a REPL where they can do interesting stuff. Adding these projects as dependencies for the root project doesn't have any other impact, since the root project is really only for aggregation. Since it's only for aggregation, I've also turned off publishing for it—it's an empty artifact and there's not really any point in having it on Maven Central.

Lastly I went ahead and used SBT's `scmInfo` settings key to set the SCM values in the pom, instead of doing these manually as XML. I've also changed the values a bit to reflect what seems to be current convention—the `url` value is an HTTP URL, and the connection value is `scm:git:<SSH URL>` (I don't think `scm:<GIT URL>` is even a valid value—I believe it'd need to be `scm:git:git://...`—but I could be wrong).

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -8,7 +8,7 @@ lazy val buildSettings = Seq(
   crossScalaVersions := Seq("2.10.5", "2.11.6")
 )
 
-lazy val compilerOptions = scalacOptions ++= Seq(
+lazy val compilerOptions = Seq(
   "-deprecation",
   "-encoding", "UTF-8",
   "-feature",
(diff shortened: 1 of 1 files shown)


vkostyukov:
@travisbrown, I always amazed by the magic you do with SBT!

Do you think it also makes sense to not publish both playground and demo projects?

travisbrown:
Thanks! I probably wouldn't publish them, but it's not a big deal. If you don't mind I could go ahead and `noPublish` them in this PR?

vkostyukov:
Yes. I think it's a good idea to exclude them. In fact, I should have done it in the very beginning. I just wasn't sure it's possible.

travisbrown:
Cool, I've updated the commit.

vkostyukov:
Great! Thanks again, this is brilliant!

vkostyukov:
Merging this!

Issue #242, opened 2015-04-26
Replace RequestReader's composite type ~ with HList


vkostyukov:
I'm sold to add a dependency to Shapeless and replace our hack types `~` and `/` with `HList`. This should be relatively easy to do right now, since in 0.6.0 release I tried to hide types `~` and `/` from users API, so they could use implicit conversion and deal with types `(A, B, ..., Z)` instead of `A ~ B ~ ... ~ Z`.

How about to make something similar to scodec combinators API:

```scala
case class User(id: Int, name: String)
val user: RequestReader[User] = (param("id").as[Int] :: param("name")).as[User]
```

I'm not sure if this even possible, but I like the API a lot. @travisbrown I don't know how does it fit into the concept you've been working on so we might want to rethink this according to your understanding.

vkostyukov:
Fixed in #246.

PR #246, opened 2015-04-27
Use Shapeless's HList for request reader composition

travisbrown:
I'm still working on the heterogeneous router, but I wanted to go ahead and post the `RequestReader` changes since they're more straightforward, and since #242 is now broken out into its own issue.

A few points for discussion:

* I'm using Shapeless 2.2.0 even though it's still a release candidate, since the release is likely to happen before 0.7.0 is ready otherwise. The only change that would be necessary (I believe) to go with 2.1 is in the `build.sbt`—before 2.2 plain old `%%` didn't work because `_2.10.4` had to be specified for 2.10.
* I don't see any way to get rid of the boilerplate for `~>` and `~~>` without requiring the types of their arguments to be fully specified, which is annoying. I've provided commented-out generic implementations, but I've left the (modified) boilerplate-y versions for now.
* I'm tempted to put the implicit classes that can't be made value classes in 2.10 because of [SI-8011](https://issues.scala-lang.org/browse/SI-8011) in version specific source trees, now that this is super easy in SBT 0.13.8. I haven't yet, though.
* One small annoyance: `a ~ b ~> f` now needs parentheses: `(a :: b) ~> f`.
* Otherwise I've only changed one test, to use a new `.tupled` method instead of explicitly mapping a tuple out of the `~`.
* I went ahead and made it so that a "bigger" request for `a` is fine in e.g. `a :: param("foo")` since this seems to fit common usage.

The syntax in #242 is supported verbatim:

``` scala
case class User(id: Int, name: String)
val user: RequestReader[User] = (param("id").as[Int] :: param("name")).as[User]
```

Thanks to Shapeless's `Generic`.

### DIFF
--- a/core/src/main/scala/io/finch/request/package.scala
+++ b/core/src/main/scala/io/finch/request/package.scala
@@ -28,13 +28,17 @@ package io.finch
 
 import com.twitter.io.Buf
 import com.twitter.finagle.httpx.{Cookie, Method}
-import com.twitter.util.{Future, Throw, Try}
+import com.twitter.util.{Future, Return, Throw, Try}
 
 import org.jboss.netty.handler.codec.http.multipart.{HttpPostRequestDecoder, Attribute}
 
 import scala.annotation.implicitNotFound
 import scala.collection.JavaConverters._
 import scala.reflect.ClassTag
+
+import shapeless._
+import shapeless.ops.function.FnToProduct
+import shapeless.ops.hlist.Tupler
 
 /**
  * This package introduces types and functions that enable _request processing_ in Finch. The [[io.finch.request]]
@@ -44,15 +48,16 @@ import scala.reflect.ClassTag
  *
  * The cornerstone abstraction of this package is a `RequestReader`, which is responsible for reading any amount of data
  * from the HTTP request. `RequestReader`s might be composed with each other using either monadic API (`flatMap` method)
- * or applicative API (`~` method). Regardless the API used for `RequestReader`s composition, the main idea behind it is
- * to use primitive readers (i.e., `param`, `paramOption`) in order to _compose_ them together and _map_ to
+ * or applicative API (`::` method). Regardless the API used for `RequestReader`s composition, the main idea behind it
+ * is to use primitive readers (i.e., `param`, `paramOption`) in order to _compose_ them together and _map_ to
  * the application domain data.
  *
  * {{{
  *   case class Complex(r: Double, i: Double)
- *   val complex: RequestReader[Complex] =
- *     param("real").as[Double] ~
- *     paramOption("imaginary").as[Double].withDefault(0.0) ~> Complex
+ *   val complex: RequestReader[Complex] = (
+ *     param("real").as[Double] ::
+ *     paramOption("imaginary").as[Double].withDefault(0.0)
+ *   ).as[Complex]
  * }}}
  *
  * A `ValidationRule` enables a reusable way of defining a validation rules in the application domain. It might be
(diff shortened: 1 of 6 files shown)


vkostyukov:
Omg. @travisbrown, this is amazing! 

Quick question. Do you think we can make the following work:

```scala
case class User(name: String)
val rr: RequestReader[User] = param("name").as[User]
```

Also, don't you think we should deprecate the `~` firstly?

vkostyukov:
My point is that we probably don't need both `~>` and `~~>` if we allow the users to use `as` on any reader and convert its underlying type into something meaningful so they could use  `map`. 

```scala
case class User(name: String)

// This this
val rr: RequestReader[User] = param("name").as[User].map(u => u.copy(name = u.name + " Bob"))

// Instead of this.
val rr: RequestReader[User] = param("name").map(_ + " Bob" ) ~> User
```

travisbrown:
Makes sense: `~` is back and the following now works:

``` scala
case class User(name: String)
val rr: RequestReader[User] = param("name").as[User]
```

Or even:

``` scala
case class Foo(i: Int)
val rr: RequestReader[Foo] = param("name").as[Int].as[Foo]
```

Do you think it's worth coming up with separate names for the `DecodeRequest` and `Generic` versions of `as`? I don't think collisions are likely in normal usage, but it seems inelegant. Alternatively we could unify them by having `Generic` provide `DecodeRequest` instances, but that seems like extra indirection for no terribly good reason.

I think `~>` and `~~>` are kind of handy, but as a compromise I've only used the boilerplate-free generic versions for `HList` request readers. So the syntax isn't quite as convenient, but they're still there if you need them.

vkostyukov:
Re name collisions. I agree it feels weird to use the same name for both cobinators. How about to rename the `Generic` version to something like `to` or `in`:

```scala
case class User(id: Int, name: String)
val a: RequestReader[User] = (param("id").as[Int] :: param("name")).to[User]
val b: RequestReader[User] = (param("id").as[Int] :: param("name")).in[User]
```

Re `~>` and `~~>` cobinators. What would be syntax look like for the `HList` request reader?

travisbrown:
(For example, just to put something on the table: `as` for parsing and `to` for `HList` to case class conversions. I'd probably prefer `parse` for parsing and `as` for generic conversions, but that breaks existing code.)

vkostyukov:
So it would be `param("id").parse[Int]` vs. `param("id").as[Int]`. I kind of like the second version more. Maybe we should keep `as` for both of the cases for now?

core/src/main/scala/io/finch/request/package.scala
```diff
@@ -530,6 +534,91 @@ package object request extends LowPriorityRequestReaderImplicits {
...
+    def as[A](implicit gen: Generic.Aux[A, L]): PRequestReader[R, A] = self.map(gen.from)
+
+    /**
+     * Converts this request reader to one that returns a tuple with the same types as this
+     * [[shapeless.HList]].
+     *
+     * Note that this will fail at compile time if this this [[shapeless.HList]] contains more than
+     * 22 elements.
+     */
+    def tupled(implicit tupler: Tupler[L]): PRequestReader[R, tupler.Out] = self.map(tupler(_))
```
vkostyukov: How about to rename it to `asTuple` to unify the conversion API:

```scala
val a = (param("a") :: param("b")).as[AB]
val a = (param("a") :: param("b")).asTuple
```

travisbrown:
Yeah, I think I agree now—I'm about to post a slight reworking that makes the distinction less important.

vkostyukov:
Cool! Thanks a lot Travis!

core/src/main/scala/io/finch/request/package.scala
travisbrown: `tupled` is inspired by the one on Scalaz's `ApplicativeBuilder`s, but I agree that `asTuple` has a nice symmetry.

vkostyukov:
Btw, feel free to deprecate the `~` syntax in this PR so we could remove it in 0.7.0-SNAPSHOT.

travisbrown:
Okay, I've deprecated `~`. I've also just added some tests for the new stuff, and I added a ScalaCheck dependency because that seemed to make more sense than writing a bunch of cases by hand.

vkostyukov:
:+1: Nice!

I think it's ready to be merged. Just one thing re `tupled`. Do you think we should keep it as `tupled`?

travisbrown:
Oh, right—I'll change it, add a test for it, and rebase all these commits.

vkostyukov:
Merging this! Thanks @travisbrown!

PR #248, opened 2015-04-28
Clean up Travis CI config

travisbrown:
Minor changes: we don't need to specify the SBT version here, and it probably makes sense to make Coveralls happen separately from the build itself.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,4 +7,5 @@ jdk:
   - openjdk7
   - oraclejdk7
   - oraclejdk8
-script: "sbt -sbt-version 0.13.8 ++$TRAVIS_SCALA_VERSION clean coverage test coveralls scalastyle"
+script: sbt ++$TRAVIS_SCALA_VERSION clean coverage test scalastyle
+after_success: sbt ++$TRAVIS_SCALA_VERSION coveralls
--- a/project/plugins.sbt
+++ b/project/plugins.sbt
@@ -6,7 +6,7 @@ resolvers ++= Seq(
 
 addSbtPlugin("com.github.mpeltonen" % "sbt-idea" % "1.5.2")
 addSbtPlugin("org.scoverage" % "sbt-scoverage" % "1.0.4")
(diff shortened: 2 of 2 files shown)


travisbrown:
As a side note, this also has the advantage of making passing builds in forks actually pass.

rpless:
It looks like sbt-coveralls released 1.0.0 a couple days ago. They have some fixes for multi-project builds, so we may just want to update to that. But :+1: to stopping the building from randomly failing. 

1.0 usage looks like this: https://github.com/scoverage/sbt-coveralls#travis-ci-integration

travisbrown:
Thanks for pointing that out, @rpless! I haven't added `coverageAggregate` because I don't know how that would change the reporting (since it's been around and we've never used it before), but I updated the version.

rpless:
Yeah that's a good call. `coverageAggregate` always worked fine for generating offline reports, but didn't work for Coveralls. I'll play with it later this week. My hope is that their fix will let us remove all the `disablePlugins(CoverallsPlugin)` we have in our build.

travisbrown:
Thanks, @rpless—I'll be curious to hear if you get it to work, since we've got some other projects where that bit of clean-up would be useful.

vkostyukov:
Just an quick question: Does anyone know why coveralls stopped reporting coverage status via Github comments to every PR?

travisbrown:
@vkostyukov No, I noticed that yesterday (or maybe last week), and figured you'd changed some setting.

vkostyukov:
Well, this is strange. We haven't changed anything and it's accidentaly broken.

travisbrown:
I don't know if it's too strange, considering that it's coveralls. :disappointed:

travisbrown:
The "Leave comments?" box is definitely checked in the PR alerts section of the settings.

vkostyukov:
I removed the threshold values (on Coveralls) and restarted the build. Hopefully, it will help.

vkostyukov:
Still no comments from @coveralls :(

Let's merge this than.

PR #250, opened 2015-04-28
Fix test, add some tests, and turn off coverage for playground

travisbrown:


vkostyukov:
Thanks! Merging this.

PR #251, opened 2015-04-28
Fix #244 & Fix #245

### LINKED ISSUE #244: EncodeResponse for byte arrays
It would be very helpful to have implicit response encoder for byte arrays and use it with TurnIntoHttp[Array[Byte]].

### LINKED ISSUE #245: doc is broken: when will be available new doc?
Hi, I'm interested in use finch but seems than the doc is broken:

Endpoints

```mkd
WARNING: io.finch.Endpoint is going to be deprecated in the 0.6.0 release. So, it's gently recommended to use the route combinators instead.

https://github.com/finagle/finch/blob/master/docs/routes.md <--- ***this DOC doesnt exist

```
endpoints seems to be an important part in the old versions and for me it's not clear how replace them, please let me know when new doc will be available, thanks!...

vkostyukov:
Fixes both issues #244 and #245.

Re #244: I hardcoded the content type to "application/octet-stream". Not sure if it's a right way to do that.

### DIFF
--- a/core/src/main/scala/io/finch/response/ResponseBuilder.scala
+++ b/core/src/main/scala/io/finch/response/ResponseBuilder.scala
@@ -37,7 +37,8 @@ import com.twitter.finagle.httpx.{Response, Cookie, Status}
 case class ResponseBuilder(
   status: Status,
   headers: Map[String, String] = Map.empty[String, String],
-  cookies: Seq[Cookie] = Seq.empty[Cookie]
+  cookies: Seq[Cookie] = Seq.empty[Cookie],
+  contentType: Option[String] = None
 ) {
 
   /**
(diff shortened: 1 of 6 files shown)


travisbrown:
Looks good to me—we may want to go [this route](https://github.com/finagle/finagle-serial/pull/39/files) (`update` with retries before the build script) since these downloading and resolution failures seem increasingly common.

vkostyukov:
Agreed! I'm tired of clicking the "restart" button.

vkostyukov:
@BenWhitehead could you please review it?

I added a new method on a `ResponseBuilder` - `withContetntType` that overrides the content type provided by `EncodeResponse`:

```scala
val rep = Ok.withConentType(Some("applicatoin/json"))("invalid json")
```

BenWhitehead:
@vkostyukov I think this is a good approach. Thanks!

vkostyukov:
Great! I'm merging this.

PR #252, opened 2015-04-29
Use shields.io for slightly more consistent badges

travisbrown:
This is probably too trivial for words, but when coveralls is down (which is often), our badges look like this:

screen shot 2015-04-29 at 6 02 13 am

If we switch to [shields.io](http://shields.io/) we could have this:

screen shot 2015-04-29 at 6 02 39 am

The fact that the gitter badge is more consistent in size and style with the others is a nice bonus.

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,10 +8,10 @@ close as possible to the bare metal Finagle API.
 
 Badges
 ------
-[![Build Status](https://travis-ci.org/finagle/finch.svg?branch=master)](https://travis-ci.org/finagle/finch)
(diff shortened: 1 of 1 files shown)


travisbrown:
And yeah, the fact that the spacing still isn't consistent bothers me, but I suspect it'll bother me less when it's not six in the morning.

penland365:
Github needs some kind of Award system in place to honor PR's of such importance and magnitude.

:shipit:

vkostyukov:
Amazing job @travisbrown! As usual!

PR #253, opened 2015-04-29
Add fully-derived request readers for case classes

travisbrown:
This isn't a candidate for merging at the moment, just a demo for tonight's meetup that we may want to consider including someday (any feedback is welcome, of course).

### DIFF
--- a/core/src/main/scala/io/finch/request/RequestReader.scala
+++ b/core/src/main/scala/io/finch/request/RequestReader.scala
@@ -26,7 +26,7 @@
 
 package io.finch.request
 
-import com.twitter.util.{Throw, Return, Future}
+import com.twitter.util.{Future, Return, Throw, Try}
 import io.finch._
 import io.finch.request.items._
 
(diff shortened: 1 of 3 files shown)


core/src/main/scala/io/finch/request/RequestReader.scala
```diff
@@ -205,4 +205,36 @@ object RequestReader {
...
+      ct: ClassTag[HV],
+      key: Witness.Aux[HK],
+      fpt: FromParams[T]
+    ): FromParams[FieldType[HK, HV] :: T] = new FromParams[FieldType[HK, HV] :: T] {
+      def reader: RequestReader[FieldType[HK, HV] :: T] =
+        param(key.value.name).as(dh, ct).map(field[HK](_)) :: fpt.reader
+    }
+  }
+
+  def to[A]: GenericDerivation[A] = new GenericDerivation[A]
```
vkostyukov: Don't you think `for[User]` feels more suitable here? I remember it was `for` in the discussion but I'm not sure which name is better.

core/src/main/scala/io/finch/request/RequestReader.scala
```diff
@@ -205,4 +205,36 @@ object RequestReader {
...
+      gen: LabelledGeneric.Aux[A, Repr],
+      fp: FromParams[Repr]
+    ): RequestReader[A] = fp.reader.map(gen.from)
+  }
+
+  trait FromParams[L <: HList] {
+    def reader: RequestReader[L]
+  }
+
+  object FromParams {
```
vkostyukov: I like it! Do you think we should do the same trick for headers (is it useful?) and bodies. Like:

```scala
val a = RequestReader.for[User].fromParams
val b = RequestReader.for[User].fromHeaders
val c = RequestReader.for[User].fromBody
```

core/src/main/scala/io/finch/request/RequestReader.scala
travisbrown: Yeah, but `for` is a keyword, which I'd forgotten during that conversation, apparently. :smile:

vkostyukov:
This is amazing! Thanks, Travis!

I like the API really much. And I think if we extend it to headers and bodies it will be very intuitive.

core/src/main/scala/io/finch/request/RequestReader.scala
vkostyukov: Oh my :) I completely forgot about it :) I should use it more to remember!

vkostyukov: How about `RequestReader.of[User]`?

vkostyukov: Or even: 

`RequestReader.materialize[User].fromParams`
`RequestReader.derive[User].fromParams`

travisbrown:
Now included in #266.

PR #255, opened 2015-04-30
Aggregate coverage reports for coveralls

rpless:
It looks like we haven't had coverage reports on Coveralls since eab7aa0247d01b4b937010b13637d9ea2cb58e26. sbt-coveralls has been failing on our build because the reports haven't been aggregated, but its been failing silently because it now runs as an `after_script` on TravisCI. This PR aggregates the reports before trying to ship them to Coveralls.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,5 +7,5 @@ jdk:
   - openjdk7
   - oraclejdk7
   - oraclejdk8
-script: sbt ++$TRAVIS_SCALA_VERSION clean coverage test scalastyle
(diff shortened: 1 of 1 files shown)


vkostyukov:
Sorry, I'm probably missing something. Should this PR enable @coveralls reporting for Github PRs? Or it fixes something else?

rpless:
I don't know if it will, but I hope it does. That's not what this PR solves, though. This is happening in our build right now: https://travis-ci.org/finagle/finch/jobs/61056215#L1983 (you might have to expand the output line to see the error). This means coveralls is failing and we aren't getting reports at all.

vkostyukov:
Right. I see. Thanks Ryan! I'm merging this.

Issue #254, opened 2015-04-29
Single Abstraction for everything: Endpoint


vkostyukov:
This is a discussion issue - there is no milestone for this yet.

I'm still thinking about "Your REST API as a Monad" approach and it seems to me that we still can do that given there is a Heterogenous Router is on the way.

The idea is to have just a single abstraction (`Endpoint`) that combines both `Router` and `RequestReader` together, so we can organize the code in separate endpoints. We can say that FInch program is a bunch of endpoints served with Finagle HTTP.

The API in my head looks something like this:

```scala
val getUser: Endpoint[User] = 
  (Get / "users" / int("id") ? param("name")) { (id, name) =>
    User(10)
  }

val getTickets: Endpoint[Seq[Ticket]] = 
  (Get / "users" / int("userId") / "tickets") { id =>
    Seq(Ticket(...))
  }

val checkHealth: Endpoint[HttpResponse] = * / "health" /> Ok("I'm Alive!")

val api: Endpoint[User :: Seq[Ticket] :: HttpResponse :: HNil] = 
  getUser :: getTickets :: checkHealth

Httpx.serve(":8081", api)
```

Also, it would be nice to be able lift an endpoint to function. Like this:

```scala
val getUser: Endpoint[User] = 
  (Get / "users" / int("id") ? param("name")) { (id, name) =>
    User(10)
  }
val f: (Int, String) => User = getUser.lift
```

This would allow to erase the difference between service-level and controller-level, which are tipically used in a traditional web-frameworks. `Endpoint`s should cover them both.

And the last idea for today. We can also replace all the routers with just one `path` and apply the same type-class based decoding approach there. I like how it plays together:

```scala
val a: Endpoint[Int :: String :: Int :: HNil] = 
  Get / path.as[Int] / path.as[String] :: param("age").as[Int]
val b: Endpoint[User] = a.as[User]
```

I'm not quite sure yet if it's even possible, but it looks pretty solid to me. 

Feedback is very welcome!

ImLiar:
As far as I understand your idea its pretty simple to make monadic trait `Endpoint[Route[_]]`, with two applies: one for "saving callback", and the second one (abstract) for applying callback on request. Just like it works in `Route`.
And all you need is an implicit conversion from Route[_] to Endpoint[Route[_]]

vkostyukov:
As mentioned in #256 by @ImLiar, the suggested API is less intuitive than the old (current) one. My understanding is following. There are to possible directions for the future Finch:

[...]

PR #256, opened 2015-05-01
Use DecodeRequest in Router

vkostyukov:
The idea is to reuse the `DecodeRequest` type in `Router`. It's inspired by the idea that we will be able to compose request readers and routers together. So the new API looks much more consistent:

```scala
val a = Get / path.as[Int] ? param("id").as[Int]
// vs
val b = Get / int ? param("id").as[Int]
```

### DIFF
--- a/core/src/main/scala/io/finch/route/package.scala
+++ b/core/src/main/scala/io/finch/route/package.scala
@@ -22,9 +22,13 @@
 
 package io.finch
 
+import io.finch.request._
+
 import com.twitter.finagle.Service
 import com.twitter.finagle.httpx.Method
 import com.twitter.util.Future
+
+import scala.reflect.ClassTag
 
 /**
  * This package contains various of functions and types that enable _router combinators_ in Finch. A Finch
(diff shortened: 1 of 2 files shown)


ImLiar:
Don't you think its kinda noisy?
I mean it became unclear and less readable as for me.

How about some alias/proxy for `path.as`? I've made PR

vkostyukov:
That was basically my goal - to unify the names of both request readers and routers to use the same schema. 

This is WIP PR and I'm not sure it will make to 0.7.0 since it "depends" on #254. My ultimate goal is to "Use single abstraction `Endpoint` for everything". So, this PR will make the transition more smooth.

At the end of the day, we will probably have an `Endpoint` with `::` compositor. So the API would look as follows:

```scala
Get :: "users" :: int :: param("name")
```

It's not really clear what exactly "int" means here. But we could make it better with `path`:

```scala
Get :: "users" :: path.as[Int] :: param("name")
```

I'm probably not going to merge this. It's just an experiment so we can use `DecodeRequest` in `Routers`.

ImLiar:
By the way: `::` is not so URI-looking as it was supposed to be at first.

vkostyukov:
Dropping this for now.

PR #258, opened 2015-05-04
Updated the Demo project to handle Json Data in the Argonaut style

penland365:
I decided to go ahead and push this - I was trying to figure out a clean way to handle the the `ToJson` type marker, but given how in flux everything is I figured it was better to just punt that to a later date and show an Argonaut style Json encoding of the data types.

I'm pretty sure there's a way to provide a default implementation of 
```scala
  trait ToJson { self => 
    def toJson: Json = self.asJson
  }
```
but I'd rather kick this up and get quick feed back.  The biggest thing is that we can can lean on Argonaut to auto-magically encode `Seq[A]`.


travisbrown:
Thanks, @penland365. Here's [an alternative proposal](https://github.com/travisbrown/finch/commit/80a2bd143200a5375fbbacca65e3faa2c66beade) that eliminates `ToJson` (it seems best to avoid mixing inheritance and type class-based serialization entirely), puts the type class instances in the case class companion objects, and uses a version of the service-wrapping approach outlined [here](http://stackoverflow.com/a/29991909/334519).

Once the coproduct router is in, we'll be able to eliminate the `JsonEndpoint` implicit class and the calls to `toJson`, but otherwise this is pretty much what usage will look like.

(Note that I've rebased against my version updates to show that this works with Argonaut 6.1.)

penland365:
I couldn't get the Endpoints to compose the way I wanted when I was playing w/ this over the weekend Travis - I was trying to define it as a service we exported in argonaut-json, but this is clearly better than mine.

:shipit:

penland365:
@travisbrown Do you think it would appropriate to place 
```scala
implicit class JsonEndpoint[Req, Rep](endpoint: Endpoint[Req, Rep]) {
  def toJson(implicit encoding: EncodeJson[Rep]): Endpoint[Req, Json] = endpoint.map { service =>
      new Service[Req, Json] {
        def apply(req: Req): Future[Json] = service(req).map(encoding(_))
     } 
  }   
}
```
inside of ``finch-argonaut`?  I've immediately duplicated that code in two projects today - adding that service seems like something we might want to export.

travisbrown:
@penland365 This will change a bit with the coproduct tooling for routers, which is coming soon, so probably best to wait?

penland365:
I'm totally fine w/ that.  :+1:

vkostyukov:
Right. With coproduct routers, we should be able to get rid of that implicit class and serve domain types directly.

travisbrown:
This is merged in #276.

Issue #259, opened 2015-05-05
Make argonaut-json depend on argonaut 6.1


lderks:
finch-argonaut still depends on 6.0.4

https://github.com/finagle/finch/blob/master/build.sbt#L147

Would be better to update this to 6.1, which is also available:

http://mvnrepository.com/artifact/io.argonaut/argonaut_2.11/6.1

vkostyukov:
Fixed in #260.

PR #260, opened 2015-05-05
Update versions for Shapeless, Argonaut, and Jackson

travisbrown:
Addresses #259 (thanks, @lderks—I missed the news that the release was out last week). Now that `DecodeRequest` is invariant the upgrade is easy.

I've also updated Shapeless and Jackson to the latest versions.


vkostyukov:
Thanks @travisbrown!

PR #261, opened 2015-05-05
added JsonItem

lderks:
Is it a good idea to add a JsonItem?  
This can be thrown when for example required json items are missing  

i.e. throw new NotPresent(JsonItem(label))

See the updated demo

### DIFF
--- a/core/src/main/scala/io/finch/request/package.scala
+++ b/core/src/main/scala/io/finch/request/package.scala
@@ -144,6 +144,7 @@ package object request extends LowPriorityRequestReaderImplicits {
     case class ParamItem(name: String) extends RequestItem("param", Some(name))
     case class HeaderItem(name: String) extends RequestItem("header", Some(name))
     case class CookieItem(name: String) extends RequestItem("cookie", Some(name))
+    case class JsonItem(name: String) extends RequestItem("json", Some(name))
     case object BodyItem extends RequestItem("body")
     case object MultipleItems extends RequestItem("request")
   }
--- a/demo/src/main/scala/io/finch/demo/Main.scala
+++ b/demo/src/main/scala/io/finch/demo/Main.scala
@@ -74,6 +74,10 @@ object Demo {
 
     case NotPresent(BodyItem) => BadRequest(Json("error" := "body_not_present"))
 
+    case NotPresent(JsonItem(j)) => BadRequest(
+      Json("error" := "json_not_present", "item" := j)
+    )
+
     case NotParsed(ParamItem(p), _, _) => BadRequest(
       Json("error" := "param_not_parsed", "param" := p)
     )
--- a/demo/src/main/scala/io/finch/demo/reader.scala
+++ b/demo/src/main/scala/io/finch/demo/reader.scala
@@ -25,6 +25,7 @@ package io.finch.demo
 import argonaut._, Argonaut._
 import io.finch.request._
 import io.finch.argonaut._
+import io.finch.request.items.JsonItem
 
 object reader {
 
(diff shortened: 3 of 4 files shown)


vkostyukov:
Folks, I'm a bit skeptical with this change. I (probably) understand the problem it tries to solve, but I'm pretty sure we need a bit different solution here. Here are my thoughts. As far as I understand this change intent to _structure_ errors from parsing the JSON body.

For example with Argonaut there two possible ways to work with JSON:

1) `body.as[Json]` (we can do it because Argonaut provides and identity `DecodeJson[Json]`)
2) `body.as[User]` if there is an instance of `DecodeJson[User]` availalve in the scope

The first usage doesn't throw any exceptions about missing JSON properties since it simply doesn't know what properties are required. So `JsonItem` feels pretty much as end-user business logic: one can create a custom exception (i.e., `JsonPropertyNotFound(property: String)`) and throw it.

With the second usage (which is preferred, by my understanding), a `NotParsed` exception will be thrown automatically by the `body` request reader. So far so good, but the information inside the exception is not structured - it's just a string that says what is missing:

```scala
(body.as[Json])(request).handle {
  case NotParsed(BodyItem, targetType, cause) => ...
}
```

So we can grab the `cause` throwable and parse its message to understand what exactly is missing to make the life of frontend developers much easier.

Thus, I believe this PR raises a very question, but the solution should probably be different. In order to understand if we could do better here, we have to check if the underlying JSON libraries provide the structured information in their exceptions so we can grab it and move inside `case class BodyItem(name: String)`, and handle it from the Finch side, not the user side. This seems like a good solution for me since it's very generic.

travisbrown:
I think that this is an incremental improvement over the `"N/A"`-returning implementation (at least in the context of a demo project), but after a closer look I agree that the structured error with `JsonItem` isn't necessary or desirable.

If you just used `body.as[Ticket]` with a `DecodeRequest[Ticket]` in scope (based on whatever kind of codec your JSON library provides), in the case of failure you'd get an exception that would be specific to the JSON library. It probably makes sense to have the same thing happen here. One approach would be to define a new type to represent a "partial" ticket (i.e. one without an `id`):

``` scala
case class NewTicket(label: String) {
  def toTicket: Ticket = Ticket(Id(), label)
}

object NewTicket {
  implicit val newTicketDecodeJson: DecodeJson[NewTicket] =
    jdecode1L(NewTicket.apply)("label")
}

val ticket: RequestReader[Ticket] = body.as[NewTicket].map(_.toTicket)
```

Alternatively you could pass in the instance directly and skip the case class definition:

``` scala
val ticket: RequestReader[Ticket] = body.as(
  decodeArgonaut(jdecode1L((label: String) => Ticket(Id(), label))("label")),
  implicitly[scala.reflect.ClassTag[Ticket]]
)
```

This works, but passing in the `ClassTag` by hand isn't ideal.

lderks:
Thanks for your great comments.

The idea behind the JsonItem was to give a more friendly feedback to the frontend developers. 

At the moment, missing json for `body.as[Ticket]` results in a `case NotParsed(BodyItem, _, cause)` error situation. The detailed exception message contains: `Attempt to decode value on failed cursor.: [*.--\\(username)]` which is to technical and argonaut specific. It would be good to have a more simple message being returned to the frontend developers.

Grabbing the cause throwable and parsing its message from the finch side into a more informative message is a better solution.

travisbrown:
You could modify the error message when defining the Argonaut codec. That's not terribly convenient, I know, but trying to represent JSON decoding errors in a library-independent way in Finch seems like a slippery slope.

If you're using Argonaut elsewhere in your application, having clear error messages for your codecs will be independently valuable, and I'd guess the Argonaut team would be interested in patches that improve error message customization.

vkostyukov:
I've created an issue #264 about it. Do you mind @lderks if I close this in favour of that issue?

lderks:
Thanks Vladimir, 

Yes lets close this pull request.

vkostyukov:
Great! Thanks again @lderks! You identified a very good problem here!

PR #262, opened 2015-05-05
Json4s encode/decode module implementation

ImLiar:
We are using json4s in our project since it provides powerful functional API to operate with AST and some neat extensions (mongo objectid, joda date time (de)serialization).

I suppose it would be useful to have this module like an optional one.

### DIFF
--- a/docs/json.md
+++ b/docs/json.md
@@ -49,7 +49,21 @@ val ok: HttpResponse = Ok(Foo(10, "foo")) // will be encoded as JSON
 val foo: RequestReader[Foo] = body.as[Foo] // a request reader that reads Foo
 ```
 
+### Json4s
+
+The `finch-json4s` module provides the support for the [JSON4S][7] library. The library usage looks as follows:
+
+```scala
+import io.finch.json4s._
+implicit val formats = DefaultFormats ++ JodaTimeSerializers.all
+case class Bar(x: Int, y: Boolean)
+
+val ok: HttpResponse = Ok(Bar(1, true)) // will be encoded as JSON
+val foo: RequestReader[Bar] = body.as[Bar] // a request reader that reads Bar
+```
+
 [3]: https://github.com/finagle/finch/blob/master/json/src/main/scala/io/finch/json/Json.scala
 [4]: http://argonaut.io
 [5]: https://github.com/non/jawn
 [6]: http://jackson.codehaus.org/
+[7]: http://json4s.org/
--- a/json4s/src/main/scala/io/finch/json4s/package.scala
+++ b/json4s/src/main/scala/io/finch/json4s/package.scala
@@ -1,0 +1,30 @@
+package io.finch
+
+import com.twitter.util.Try
+import io.finch.request.DecodeRequest
+import io.finch.response.EncodeResponse
+import org.json4s._
+import org.json4s.jackson.JsonMethods
+import org.json4s.jackson.Serialization._
+
+package object json4s {
+
+  /**
+   * @param formats json4s `Formats` to use for decoding
+   * @tparam A the type of data to decode into
+   */
+  //@TODO get rid of Manifest as soon as json4s migrates to new reflection API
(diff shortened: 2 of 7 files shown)


json4s/src/main/scala/io/finch/json4s/package.scala
```diff
@@ -0,0 +1,29 @@
...
+import org.json4s.jackson.JsonMethods
+import org.json4s.jackson.Serialization._
+
+package object json4s {
+
+  /**
+   * @param formats json4s `Formats` to use for decoding
+   * @tparam A the type of data to decode into
+   */
+  implicit def decodeJson[A : Manifest](implicit formats: Formats): DecodeRequest[A] = DecodeRequest(
```
vkostyukov: I believe it should extend `DecodeAnyRequest` by analogy with Jackson. And can we use `ClassTag` here instead of `Manifest` since they are deprecated (please correct me if I'm wrong) in 2.11.

vkostyukov:
That's great @ImLiar! We should definitely merge this in. Just a couple of nitpicks from my side.

json4s/src/main/scala/io/finch/json4s/package.scala
ImLiar: Yea, I know about manifest, but json4s still depends on it :disappointed:

vkostyukov: Argh. Let's for with `Manifest` for now then. But please, add a TODO note that it has to be uodated.

json4s/src/main/scala/io/finch/json4s/package.scala
```diff
@@ -0,0 +1,29 @@
...
+    (input: String) => Try(JsonMethods.parse(input).extract[A])
```
vkostyukov: I'm pretty sure you don't need to specify the type of `input` jere.

vkostyukov:
Please, also add a new package in the list in README.md file.

docs/json.md
```diff
@@ -53,3 +53,22 @@ val foo: RequestReader[Foo] = body.as[Foo] // a request reader that reads Foo
...
+
+The `finch-json4s` module provides the support for the [JSON4S][7] library. The library usage looks as follows:
+
+```scala
+import io.finch.json4s._
+implicit val formats = DefaultFormats ++ JodaTimeSerializers.all
+case class Bar(x: Int, y: Boolean)
+
+val ok: HttpResponse = Ok(Bar(1, true)) // will be encoded as JSON
+val foo: RequestReader[Bar] = body.as[Bar] // a request reader that reads Foo
```
vkostyukov: that reads "Bar".

ImLiar:
Well, since `DecodeAnyRequest` doesnt allow to pass `Manifest[A]` I had to deal with `DecodeRequest[A]`. Nothing to do here before **json4s** fix

docs/json.md
```diff
@@ -53,3 +53,22 @@ val foo: RequestReader[Foo] = body.as[Foo] // a request reader that reads Foo
...
+```
+
+[3]: https://github.com/finagle/finch/blob/master/json/src/main/scala/io/finch/json/Json.scala
```
vkostyukov: Somehow this is duplicated: 3-6 are defined before ### Json4s.

vkostyukov:
Right. `DecodeAnyX` uses `ClassTags`. Alright then. Let's use `DecodeX`.

docs/json.md
```diff
@@ -53,3 +53,22 @@ val foo: RequestReader[Foo] = body.as[Foo] // a request reader that reads Foo
 [4]: http://argonaut.io
```
vkostyukov: These already duplicated bellow.

ImLiar:
I'm missing some nice markdown editor :cry:

vkostyukov:
I think it's ready to go alive. @ImLiar could you please squash the commits so I can merge this in?

vkostyukov:
This is great! Thanks a lot @ImLiar!

PR #265, opened 2015-05-08
Aggregate json4s, fix one style issue, and change rule on another

travisbrown:
I just noticed that the new `json4s` wasn't being aggregated, which is the only reason it wasn't crashing Scalastyle. It seems reasonable to allow `json4s` as a package name, so I changed that rule. Also fixed a misspelled file name.


vkostyukov:
Thanks @travisbrown! Good catch!

Issue #121, opened 2014-12-18
Add benchmark


vkostyukov:
We can add a sub-project `finch-benchmark` with JMH-based benchmark to test core basic blocks. I would pay a special attention for `RequestReader` abstraction, which is the cornerstone of the Finch applications. We can also test `finch-json` project in the same benchmark.

vkostyukov:
We can also test `Router`.

vkostyukov:
As discussed in the Gitter channel, we might want to test different JSON libraries supported in Finch.

travisbrown:
I've just been putting together some initial benchmarks, since I want to keep an eye on the performance impact of the API additions and changes I'm making. I'll have a pull request ready this morning, but in the meantime I thought people might find this interesting:

```
[info] i.f.b.s.argonaut.ArgonautBenchmark.userOperations  avgt   20  181.007 ± 5.217  ms/op
[info] i.f.b.s.finagle.FinagleBenchmark.userOperations    avgt   20  157.384 ± 5.480  ms/op
[info] i.f.b.s.jackson.JacksonBenchmark.userOperations    avgt   20  157.506 ± 4.963  ms/op
[info] i.f.b.s.json4s.Json4sBenchmark.userOperations      avgt   20  180.603 ± 4.645  ms/op
```

There's no noticeable difference between using Finch with Jackson over the vanilla Finagle HTTP API (also with Jackson), and while you pay a little more for Argonaut and Json4s, it's not much (~14%).

travisbrown:
We can close this now that #266 is merged, no?

vkostyukov:
Sure. Let's close this.

PR #266, opened 2015-05-09
Add some initial benchmarks for request readers and services

travisbrown:
I've added some JMH benchmarks to make it easier to keep track of the performance impact of the new API additions and changes for 0.7.0. This partially addresses #121—I think we can mark that complete when the router benchmarks (which I'll push with the new coproduct stuff, probably later today) go in.

The documentation is light and the more integration-y service benchmark doesn't have any tests yet, but I wanted to go ahead and get it out as a starting point.

Note that this includes my #253 (rebased), since I wanted to be able to compare the performance of a fully-derived request reader.

The initial results are heartening. While the derived reader is a little slower than the others (about 75% of the throughput of the other applicative readers), there's no real difference between the new `HList`-based applicative readers (even with the `Generic` machinery) and the old `~` readers:

```
[info] Benchmark                                             Mode  Cnt       Score      Error  Units
[info] FailingRequestReaderBenchmark.derivedReader          thrpt   20   52650.452 ± 1605.621  ops/s
[info] FailingRequestReaderBenchmark.hlistApplyReader       thrpt   20   60900.097 ±  989.982  ops/s
[info] FailingRequestReaderBenchmark.hlistGenericReader     thrpt   20   60573.071 ± 1062.515  ops/s
[info] FailingRequestReaderBenchmark.monadicReader          thrpt   20  201983.780 ± 5474.721  ops/s
[info] FailingRequestReaderBenchmark.tildeReader            thrpt   20   61516.793 ±  550.878  ops/s
[info] SuccessfulRequestReaderBenchmark.derivedReader       thrpt   20  125559.994 ± 2311.396  ops/s
[info] SuccessfulRequestReaderBenchmark.hlistApplyReader    thrpt   20  163283.190 ± 2934.045  ops/s
[info] SuccessfulRequestReaderBenchmark.hlistGenericReader  thrpt   20  162149.855 ± 1955.893  ops/s
[info] SuccessfulRequestReaderBenchmark.monadicReader       thrpt   20  256000.476 ± 7205.922  ops/s
[info] SuccessfulRequestReaderBenchmark.tildeReader         thrpt   20  169802.934 ± 2864.328  ops/s
```

(Note that it's not really fair to compare the monadic one to the others, since it doesn't accumulate errors in the case of invalid requests.)

The service results also look pretty good (here lower is better):

```
[info] Benchmark                                          Mode  Cnt    Score   Error  Units
[info] i.f.b.s.argonaut.ArgonautBenchmark.userOperations  avgt   20  181.648 ± 6.394  ms/op
[info] i.f.b.s.finagle.FinagleBenchmark.userOperations    avgt   20  155.486 ± 4.754  ms/op
[info] i.f.b.s.jackson.JacksonBenchmark.userOperations    avgt   20  160.128 ± 4.651  ms/op
[info] i.f.b.s.json4s.Json4sBenchmark.userOperations      avgt   20  177.745 ± 5.174  ms/op
```

There's no real difference between using Finch with Jackson vs. the vanilla Finagle HTTP API (also with Jackson), and Argonaut and Json4s are only a bit slower.


vkostyukov:
Should we abandon the previous PR on derived RRs #253 then? I'm excited to merge this in! Let me know once you feel it's ready to go alive.

travisbrown:
We still need to decide on names for #253. I don't particularly like `RequestReader.derive[User]`, but maybe `RequestReader.deriveFor[User]`? Or `RequestReader.to[User]`, which I think would still be my first pick.

Do you think `fromHeaders` or `fromBody` would actually be useful? I'm not sure what `fromBody` would look like, since you'd use `body.as[User]` to get the contents decoded as a JSON representation of a user.

vkostyukov:
I picked [`derive`](https://github.com/argonaut-io/argonaut/blob/master/src/main/scala/argonaut/EncodeJson.scala#L47) from Argonaut. Is it possible to omit `for` by replacing it with apply like this: `RequestReader[User].fromParams`?

Yeah. It doesn't make sense to have `fromBody`. And I don't think `fromHeaders` would be useful. So, let's go with `fromParams` for now.

travisbrown:
Oh, hmm, I guess that's a reasonable precedent—it just sounds like you're deriving a `User`.

I don't know about `apply`—it'd work but I think it'd be better to indicate more explicitly that there's a lot of magic going on. I'll change it to `derive`.

travisbrown:
[Results with `-prof gc`](https://gist.github.com/travisbrown/6cd36967259ab331a595).

vkostyukov:
I'm merging this in. Amazing job @travisbrown!

PR #267, opened 2015-05-09
Begin to abstract JSON testing tools

travisbrown:
There's currently a lot of duplication and inconsistency across the tests for the `argonaut`, `json4s`, and `jackson` projects, and the tests that are in place now only validate a few hand-rolled cases. Testing against JSON strings is also fragile, since most libraries don't make guarantees about the order of fields in objects, and since our API doesn't make any guarantees about whitespace (at least I think it doesn't?).

This PR is a quick sketch of some abstractions that allow us to test most (or all) of the functionality of these subprojects in a few lines:

```
"The Argonaut codec provider" should "encode a case class instance as JSON" in encodeFoo
it should "decode a case class instance from JSON" in decodeFoo
it should "encode a list of case class instances as JSON" in encodeFooList
it should "decode a list of case class instances from JSON" in decodeFooList
```

I haven't removed any of the existing tests, although a lot of them are redundant now. If other people are interested in this approach I can do that clean-up, etc.

I've also done a bit of minor rearranging in the `argonaut` tests.

### DIFF
--- a/argonaut/src/test/scala/io/finch/argonaut/ArgonautSpec.scala
+++ b/argonaut/src/test/scala/io/finch/argonaut/ArgonautSpec.scala
@@ -1,69 +1,29 @@
 package io.finch.argonaut
 
-import argonaut.Argonaut._
 import argonaut._
-import com.twitter.finagle.Service
-import com.twitter.finagle.httpx.Request
-import com.twitter.util.{Await,Return}
-import io.finch._
-import io.finch.request._
-import io.finch.response.TurnIntoHttp
-import org.jboss.netty.handler.codec.http.HttpHeaders
+import argonaut.Argonaut._
+import io.finch.test.json._
+import org.scalatest.prop.Checkers
 import org.scalatest.{FlatSpec, Matchers}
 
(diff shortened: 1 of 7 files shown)


vkostyukov:
That looks perfect! @travisbrown please, go ahead with cleaning up all the redundant JSON tests we have now. I especially love PRs that remove more code that add while maintain the functionality. Well done!

travisbrown:
Okay, I've updated this. Still a net positive in terms of lines of code, but it's close, and a lot cleaner. Shows 100% code coverage for `argonaut`, `jackson`, and `json4s`.

Note that I've stopped testing request and response-specific stuff in the JSON projects' tests, since if we're testing `DecodeRequest` and `EncodeResponse` properly this shouldn't be necessary.

vkostyukov:
I'm merging this. Thanks @travisbrown!

vkostyukov:
Oh wait. I can't do it. @travisbrown could you please rebase it?

travisbrown:
Done, thanks!

vkostyukov:
Great! I'm merging this.

PR #268, opened 2015-05-10
Fix 249: Shapeless-powered Router

vkostyukov:
This work is in progress. I haven't deprecated the `/` type yet, but planning to do so. I was trying to follow the style/ideas @travisbrown used for request readers.

I added just one test for now.

### DIFF
--- a/core/src/main/scala/io/finch/package.scala
+++ b/core/src/main/scala/io/finch/package.scala
@@ -22,9 +22,6 @@
 
 package io
 
-import com.twitter.finagle.httpx.path.Path
-import io.finch.response.{Ok, EncodeResponse}
-
 import com.twitter.finagle.httpx
 import com.twitter.util.Future
 import com.twitter.finagle.Service
--- a/core/src/main/scala/io/finch/route/Router0.scala
+++ b/core/src/main/scala/io/finch/route/Router0.scala
@@ -121,3 +121,24 @@ object ** extends Router0 {
   def apply(route: Route): Option[Route] = Some(Nil)
   override def toString = "**"
 }
+
+object Router0 {
+
+  /**
+   * Implicit class that provides `::` on any [[Router0]] to support building HList routers.
+   */
+  final implicit class ValueRouter0Ops[B](self: B)(implicit ev: B => Router0) {
+
+    /**
+     * Composes this [[Router0]] with the given `that` [[RouterN]] in terms of logical **and**.
+     */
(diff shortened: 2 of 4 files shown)


travisbrown:
What do you think about keeping the `/` syntax? We could still use `HList` behind the scenes.

vkostyukov:
I was thinking about keeping the `/` API unchanged: I'm pretty sure some users rely on the `/` composite type and use pattern-patching over it. As far as I remember, @ImLiar was doing something like this with custom implicit conversions over the `/` type.

Although, it seems to me that it's quite possible to migrate from `/` to `::`. And the changes are quite mechanical. So, maybe we should just break the current API. My only wish is to introduce the `::` method as an alternative: for consistency.

vkostyukov:
Rebased against master but haven't touched the API yet.

vkostyukov:
I tried to reuse the `/` method name: it seems to be painfull. I would much prefer a name that ends with colon to simplify the implementation. So I would go with `::` for now.

ImLiar:
I was on vacation and skipped a lot with new router API. 
Whats the pain with `/` method name? Is it possible to make at least alias method `/` -> `::`?

vkostyukov:
I've rebased it against master.

@ImLiar answering your question. The reason for this is a special treatment of operators like "::" (ends with a colon). Using the right-associative operator `::` allows us build HList routers pretty easy since we're basically doing in the same manner we build lists: from right to left (starting with `Nil` and pretending heads one-by-one).

With left-associative operator `/` we would need to _append_ new element to HList. This is the one problem. The other problem is how to tell the difference between `Router[String]` and `Router[HList]` in terms of `def /[A](that: Router[A])`. This information is not available at runtime, but it's available at compile time, so we had `implicit class RouterOps[L <: HList](r: Router[L])`.

ImLiar:
Oh, I completely forgot about associativity in scalac. Well, from this point it sounds reasonable.

travisbrown:
It's possible to build up hlists with a left-associative operator in a way that's not too terrible (see [this](https://gist.github.com/travisbrown/826aea169b149931739d) for a similar coproduct example).

I'd personally prefer `/`, in part to distinguish it from the `::` for request readers, since what's happening here is very different—we're consuming tokens off a stack, not composing independent reading operations. I'll try to make a bit of time to look at this this afternoon, but might not get to it until tomorrow.

vkostyukov:
Closed in favour of #306.

PR #269, opened 2015-05-10
removed disablePlugin from all subprojects

rpless:
I forgot to do this when we initially bumped sbt-scoverage versions. We no longer need to explicitly disable the CoverallsPlugin on each sub project ( @travisbrown I believe you were looking for this for some other projects).

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -106,20 +106,17 @@ lazy val root = project.in(file("."))
 lazy val core = project
   .settings(moduleName := "finch-core")
   .settings(allSettings)
-  .disablePlugins(CoverallsPlugin)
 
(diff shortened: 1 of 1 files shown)


travisbrown:
Looks good to me—thanks, @rpless.

By the way, Coveralls support seems to be broken again in sbt-scoverage 1.1.0 with sbt-coveralls 1.0.0, but [should be fixed once sbt-coveralls 1.1.0 comes out](https://github.com/scoverage/scalac-scoverage-plugin/issues/104#issuecomment-99160357).

vkostyukov:
Great! Thank you @rpless and @travisbrown!

PR #270, opened 2015-05-10
Add sketch of coproduct endpoints

travisbrown:
I'm super unsatisfied with this, but I've been through so many iterations that I just want to get something up so I can continue iterating in public at least.

I've included the commits made and mentioned in @penland365's #258, since I wanted to show off the new syntax in the demo project. I've also updated the service benchmark to use `:|:`, and the results look about the same (I'm rerunning with `-prof gc` now).

As a quick overview of usage, you write something like this:

``` scala
val users: Endpoint[HttpRequest, HttpResponse] =
  Get    / "users" / long /> getUser :|:
  Get    / "users" /> allUsers       :|:
  Post   / "users" /> createUser     :|:
  Put    / "users" /> updateUser     :|:
  Delete / "users" /> deleteUsers
```

And if all of the composed endpoints return either `Response` or something with an `EncodeResponse` instance, you'll get a service-able endpoint. Notably the types of the individual endpoints don't have to all be the same.

### DIFF
--- a/benchmarks/src/main/scala/io/finch/benchmarks/service/model.scala
+++ b/benchmarks/src/main/scala/io/finch/benchmarks/service/model.scala
@@ -25,7 +25,7 @@ class UserDb {
 
   def all: Future[List[User]] = Future.value(users.synchronized(users.values.toList))
 
-  def delete: Future[Int] = Future.value(
+  def delete(): Future[Int] = Future.value(
     users.synchronized {
       val count = users.size
       users.clear()
--- a/benchmarks/src/main/scala/io/finch/benchmarks/service/service.scala
+++ b/benchmarks/src/main/scala/io/finch/benchmarks/service/service.scala
@@ -30,29 +30,26 @@ class FinchUserService(implicit
 
   def allUsers: Service[HttpRequest, List[User]] = Service.const(db.all)
 
-  val createUser: Service[HttpRequest, Long] = Service.mk { req =>
+  val createUser: Service[HttpRequest, HttpResponse] = Service.mk { req =>
     for {
       NewUserInfo(name, age) <- body.as[NewUserInfo].apply(req)
-      user <- db.add(name, age)
-    } yield user
+      id <- db.add(name, age)
+    } yield Created.withHeaders("Location" -> s"/users/$id")()
   }
 
-  val updateUser: Service[HttpRequest, Unit] = Service.mk { req =>
-    body.as[User].apply(req).flatMap(db.update)
+  val updateUser: Service[HttpRequest, HttpResponse] = Service.mk { req =>
+    body.as[User].apply(req).flatMap(db.update).map(_ => NoContent())
   }
 
-  def deleteUsers: Service[HttpRequest, Int] = Service.const(db.delete)
-
-  def toHttp[A](f: A => HttpResponse): Service[A, HttpResponse] = Service.mk(a => f(a).toFuture)
+  def deleteUsers: Service[HttpRequest, HttpResponse] =
+    Service.const(db.delete().map(count => Ok(s"$count users deleted")))
 
   val users: Endpoint[HttpRequest, HttpResponse] =
-    (Get / "users" / long /> getUser).map(_ ! toHttp(Ok(_))) |
-    (Get / "users" /> allUsers).map(_ ! toHttp(Ok(_))) |
-    (Post / "users" /> createUser).map(
-      _ ! toHttp(id => Created.withHeaders("Location" -> s"/users/$id")())
-    ) |
-    (Put / "users" /> updateUser).map(_ ! toHttp(_ => NoContent())) |
-    (Delete / "users" /> deleteUsers).map(_ ! toHttp(count => Ok(s"$count users deleted")))
+    Get    / "users" / long /> getUser :|:
+    Get    / "users" /> allUsers       :|:
+    Post   / "users" /> createUser     :|:
+    Put    / "users" /> updateUser     :|:
+    Delete / "users" /> deleteUsers
 
   val handleExceptions = new SimpleFilter[HttpRequest, HttpResponse] {
     def apply(req: HttpRequest, service: Service[HttpRequest, HttpResponse]): Future[HttpResponse] =
--- a/core/src/main/scala/io/finch/route/package.scala
+++ b/core/src/main/scala/io/finch/route/package.scala
@@ -25,6 +25,9 @@ package io.finch
 import com.twitter.finagle.Service
 import com.twitter.finagle.httpx.Method
 import com.twitter.util.Future
+import io.finch.response.{EncodeResponse, Ok}
+import shapeless._
+import shapeless.ops.coproduct.{Mapper, Unifier}
 
 /**
  * This package contains various of functions and types that enable _router combinators_ in Finch. A Finch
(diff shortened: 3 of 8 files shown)


travisbrown:
As a side note: this PR includes the first implicit conversions I've written in ages, and I still feel a little dirty (even though I wrote them days ago). I couldn't really get a syntax I liked without them, but we might want to revisit.

vkostyukov:
Thanks @travisbrown! This is a great start!

I was wondering if we could think of coproduct routers not coproduct endpoints. Don't you think it will be more general solution if we lift this logic to `Router`, so we could write:

```scala
val g: Router[HttpResponse] = Get / "hello" /> Ok("Hello, World!")
```

This works in my head if we 1) convert value `Router[A]` into `Router[A :+: CNil]` and then 2) apply polymorphic fold to convert it into `Router[HttpResponse]` that might be served. How does it sound?

travisbrown:
I started out with that, but it seemed to add complexity and a lot of fiddliness to support collecting coproducts in `RouterN` instead of `Endpoint`. For the conversion to an `Endpoint[A, HttpResponse]`, for example, would you want all of the following to work as coproduct elements in the `RouterN`?

* `A: EncodeResponse`
* `HttpResponse`
* `Service[A, B: EncodeResponse]`
* `Service[A, HttpResponse]`

This should be doable, and I had something kind of like that mostly working, but it didn't seem worth it. I'd be happy to put it back in place and fix up the details, though.

travisbrown:
To expand a bit: with the endpoint router the user just provides an `Endpoint[A, B :+: C :+: ... :+: CNil]`, which means any of our processing that works on the coproduct (to merge it to an `HttpResponse`, for example) doesn't need to know anything about `A`—it can be `HttpRequest` or a custom subtype of `HttpRequest` or anything else.

If we start with `RouterN[B :+: C :+: ... :+: CNil]`, we still need to end up in an `Endpoint[A, HttpResponse]`, but we (not the user) have to determine what `A` should be. Presumably it's just the most specific of the input types of the services in the coproduct (with a compile-time error if there isn't a unique one). We can manage that with a polymorphic fold over the coproduct, but that's a lot messier than the polymorphic map in the current implementation.

travisbrown:
One more footnote (or question): I've been working on the assumption that the coproduct stuff should work inside the framework of the current router / request reader / endpoint model for now. In the slightly longer term it should make it easier to unify them, but in the meantime it can help us out a lot while we figure out the new model.

vkostyukov:
Right. I understand your concerns. I was thinking about coproduct routers as about _new_ thing that is implemented in Finch to support some _new_ style. My original idea was to _fix_ the request type to `HttpRequest` and deal only with output type from `RouterN[A]`, so we should not worry about services (and endpotins) at all. Thus, at the end of the day, we will end up with just `RouterN[HttResponse]` that we can serve. We still can support services, but I believe dealing with `Service[HttpRequest, HttpResponse]` is sufficient for now.

So, I would go with the following list for now:
- `A: EncodeResponse`
- `Future[A: EncodeResponse]`
- `HttpResponse`

Travis, what do you think about it? My global idea is to move in the direction where the request type is fixed, so instead of filters (i.e., auth) users would use just request readers that fetch an authorized users from the request.

vkostyukov:
I remember, @jenshalm and I discussed the _right approach_ of doing authentication, etc (everything that might be done with filters). At that point, we decided to allow users to override the request type with filters, given there is an implicit view available. But it seems like a hack for me today. Finagle Httpx doesn't support inheritance over the `Request` type. I talked to Neuman, he said that Finagle Httpx moving to immutable request/response that are just case classes.

My point is that using filters within Finagle HTTP is kinda a bad idea, especially if there is `RequestReader` available in Finch. You can do the same thing with request reader but it will be 1) reasonable 2) type-safe 3) composable.

This is just my vision. It might be wrong though. So I'm totally open for discussion around this topic. So, before updating this PR we should probably agree that Future Finch has fixed request type but has a bunch of hand tools (request readers) that allow to implement the same behaviour but in a way it fits Finch ideology.

vkostyukov:
One more note. I'm not 100% sure that unifying the request type and introducing the coproduct routers is a right thing to do. I understand that it will require a reasonable migration path for current users who rely on custom request types and filters. But I certainly believe that consistent restrictions in the library/framework are usually a good thing that promotes a clean and unified coding style. Most of the succeed libraries nowadays use such restrictions in some way.

That said, I'm open for discussion about this. If we decide to not unify the request type, this PR fits pretty nicely into the current picture.

travisbrown:
I don't have a strong opinion here, but my feeling is that the coproduct stuff is mostly orthogonal to the questions of unifying the request type, etc. We could have either coproduct endpoints or coproduct routers in the current model (although the latter are a little less clean).

Just to summarize, the main problem coproducts solve is that there's currently no way to `|` two endpoints (or routers) that have different types. This means you have to take _every_ endpoint _all the way_ to `HttpResponse` (or some serialization type like `Json`) before you combine them, which is verbose and can really hurt readability. Coproducts give us a meaningful way to `|` endpoints with different types, which gives us an immediate payoff regardless of what we decide on the filter vs. only-request-readers issue.

My vote would be to polish up the approach here and publish 0.7.0. The `HList`-powered request readers and coproduct endpoints seem like significant enough additions / improvements to justify a release, and it'd be nice to get more feedback on how they work in the wild, in part because it might help inform our thinking on how to simplify the router / endpoint / request reader story.

vkostyukov:
From your perspective is it going to be a big deal to promote coproduct endpoints to coproduct routers in the next release? Sorry, I'm probably missing something due to lack of experience with Shapeless, but how is it different to implement coproduct routers vs. coproduct endpoints in terms of underlying Shapeless-specific routines?

travisbrown:
The part that collects the values into the coproduct is more or less identical for routers and endpoints (with [a bit more abstraction](https://twitter.com/travisbrown/status/598117265339490304) it could be literally identical). The part that's different is how to operate on the resulting coproduct. In the endpoint case we only have to work with `Response` and `A: EncodeResponse`, which means a polymorphic map with a simple `Poly1` is all we need. If we've got those things plus Finagle services in the coproduct we need a polymorphic fold instead with a more complex `Poly2`.

vkostyukov:
I see. In the very beginning, I was thinking about just 3 cases (for coproduct `Router`): 
- `A: EncodeResonse`
- `HttpResponse`
- `Future[A: EncodeResponse]`

As far as I understand, having just these three cases will result in almost identical implementation, right? As long as we're not mixing in Finagle service, we're good?

travisbrown:
Yeah, definitely—including services isn't even too bad if we fix the input type to `Request`.

Is the idea still that you'd go from a `RouterN[X :+: Y :+: CNil]` to a `RouterN[Service[Request, Response]]`?

vkostyukov:
Yeah, I think so, given that we have wrap `Future[A]` with something. Before we start doing anything, let me put a gist of our playroud project with possible Coproduct `Router`, so we can see the whole picture.

vkostyukov:
I'm probably thinking about something like this: https://gist.github.com/vkostyukov/8e437949a39951b153d3. But, it turned out we also need to allow `RequestReader` be in a coproduct, which might be a problem since it's identical to Finagle services.

travisbrown:
But `RequestReader` fixes the input type, so it shouldn't be too bad. I'll aim to put together an implementation supporting your example tonight.

travisbrown:
Here's a [quick implementation](https://github.com/finagle/finch/commit/2ebab816434ecc3b599d174ed251cd17b463ef48).

vkostyukov:
That looks great! Does it work for 2.10?

travisbrown:
Yep—I haven't updated the demo, though, and it still needs some tests.

travisbrown:
Superseded by #276.

PR #273, opened 2015-05-18
add Qubit to users

jibs:
Thanks for an awesome project :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -75,6 +75,7 @@ Adopters
 * [Sabre Labs](http://sabrelabs.com)
 * [Spright](http://spright.com)
 * [SoFi] (https://www.sofi.com/)
+* [Qubit] (http://www.qubit.com)
 * *Submit a pull-request to include your company/project into the list*
 
 Contributing


vkostyukov:
Hi @jibs! 

This is awesome! I'm very glad Finch is used to build such cool things like Qubit!

Do you mind to share a bit of detail (since which version do you use Finch, what modules do you use etc.) so I and other Finch developers could understand the priorities of our adopters?

vkostyukov:
I've merged it. Thanks @jibs!

jibs:
Sorry for the delay :)

Sure - we're using a bunch of different versions - vast majority is 0.6.0 at the moment. Primarily finch is augmenting a lot of the internal routing stuff we wrote at Qubit (we've been using finagle for a while so there is a lot of this code). There have been some tiny contributions back by us but we hope to make this more substantial as finch gets more traction here.

Thanks again!

vkostyukov:
Thanks @jibs! This is grreat!

PR #274, opened 2015-05-22
Consider using codecov.io instead of coveralls.io

rpless:
I've been becoming increasingly frustrated with coveralls. It has some uptime issues, doesn't always pick up reports, and the tooling is not great. codecov.io seems to have more features and a really slick browser plugin that integrates coverage data right onto github. It also seems to have taken some initiative on the tooling front and doesn't require you to conform to their api, they transform standard scoverage reports to work in their system (which means we have one less plugin to rely on and thus one less point of failure). I can't speak to the uptime yet because I haven't been using it very long.

I think we should consider moving in a version or two, especially if we keep having issues with coveralls. I'd encourage you to look at [this branch](https://codecov.io/github/rpless/finch?branch=codecov-test) on my fork and let me know what you think. Also definitely check out the browser plugin.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,4 +8,5 @@ jdk:
   - oraclejdk7
   - oraclejdk8
 script: sbt ++$TRAVIS_SCALA_VERSION clean coverage test scalastyle && sbt ++$TRAVIS_SCALA_VERSION coverageAggregate
-after_success: sbt ++$TRAVIS_SCALA_VERSION coveralls
+after_success:
+  - bash <(curl -s https://codecov.io/bash)
--- a/README.md
+++ b/README.md
@@ -9,7 +9,7 @@ close as possible to the bare metal Finagle API.
 Badges
 ------
 [![Build Status](https://img.shields.io/travis/finagle/finch/master.svg)](https://travis-ci.org/finagle/finch)
-[![Coverage Status](https://img.shields.io/coveralls/finagle/finch/master.svg)](https://coveralls.io/r/finagle/finch?branch=master)
+[![Coverage Status](https://img.shields.io/codecov/c/github/codecov/finagle/finch/master.svg)](https://codecov.io/github/finagle/finch)
 [![Gitter](https://img.shields.io/badge/gitter-join%20chat-green.svg)](https://gitter.im/finagle/finch?utm_source=badge&utm_medium=badge&utm_campaign=pr-badge&utm_content=badge)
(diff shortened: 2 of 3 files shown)


vkostyukov:
Looks cool! Thanks @rpless! Does it report code coverage statistic per PR? @travisbrown what do you think about switching from Coveralls?

rpless:
Yes and its configurable for the amount of information you want. Have a look at https://github.com/rpless/finch/pull/2#issuecomment-105517281. Also its opt-in so you have to explicitly tell codecov to do it.

rpless:
One annoying thing though is each time it gets a build it updates the comment. So for every build in our matrix we would get a notification. So this is my github newsfeed right now:

selection_001

I'll probably file an issue with them to see whats up with this, but I see it as a con.

vkostyukov:
We should probably also change the badge in the README file. I just checked out: http://shields.io has badges for codecov.

travisbrown:
I'm open to this move, given all the trouble we've had with Coveralls in this and other projects, but ideally we'd be consistent across Finagle org projects (and Twitter projects generally).

The one concern I have is that we'd be relying a little more heavily on a third party with this move. The details of the scoverage reports have been changing a lot (relative paths, etc.), which sometimes breaks Coveralls reporting, but at least we can fix the breakage via the open source sbt-coveralls plugin (or bother the maintainers to fix it). It's handy to be able to rely on Codecov to handle scoverage reports without a plugin, but it also means that if they decide they don't want to support something that scoverage starts doing, we're stuck writing our own plugin from scratch.

rpless:
I made the README updates.

@travisbrown I think that's a valid concern. Codecov not being open source is definitely a con for us. I'm still on the fence about switching to it, but I thought I'd at least bring it up and see how others felt about.

I talked to their support about not spamming notifications and it seems like its just a github behavior. Rather than repeatedly comment on a PR, codecov will update the comment for each build, each sends notifications.

vkostyukov:
@travisbrown maybe we should go with codecov fo now? It believe, it's pretty easy to switch back in case of any troubles. At least, codecov reports code coverage per PRs, which I'm missing very much.

travisbrown:
Sure, that's fine with me—it'll be a good experiment.

vkostyukov:
Great! I'm merging this then.

travisbrown:
I've just approved the application access request, so we should be ready to go.

rpless:
The badge currently reads "unknown coverage". Looks like I didn't properly form the shields link. The correct badge should be:
[Coverage Status](https://codecov.io/github/finagle/finch)

I can make a PR when I get home, unless one of you just wants to slip this into master within the next hour or so.

Issue #238, opened 2015-04-23
Revise all the implicit conversions


vkostyukov:
We might probably want to move them into the corresponding companion objects (if possible) to make the compilation faster for both Finch contributors and Finch users.

vanyacohen:
I'm happy to help make this happen, but I've encountered some trouble identifying offending code. It seemed like every implicit method I found was already in an object (or trait extended by an object). Would you be willing to help me understand what needs to be fixed?

vkostyukov:
@vanyacohen, Travis and I are addressed a lot of conversions in #278, but I'm pretty sure something is left. Otherwise, we can just close this issue.

travisbrown:
Hi @vanyacohen—one example I just noticed by grepping for `implicit def` is `toOptionalRule` in the `request` package object. Right now you have to import the contents of the package object to get it to apply, and then it's just another implicit conversion in scope that the Scala compiler has to consider all the time. If it were in the `ValidationRule` companion object (since it has `ValidationRule` as both source and target), it wouldn't need to be imported and the compiler would only consider it when it needs an implicit conversion to or from `ValidationRule`.

There are probably at least a few other conversions like this still around.

vkostyukov:
Thanks @travisbrown! `ValidationRule` is probably the lowest hanging fruit for this ticket. Once we fix it we can probably close the ticket.

vanyacohen:
Yes, thanks @travisbrown! I'll go ahead and move toOptionalRule to the ValidationRule companion object and look for similar conversions.

PR #276, opened 2015-05-29
Add coproduct routers

travisbrown:
This is an implementation of coproduct routers as discussed in #270, which this pull request supersedes.

I've also rebased and made some version updates (including to Shapeless 2.2.0 final). Still needs a little cleanup and work on documentation and tests, which I should be able to get to tomorrow (I'm on vacation today).

### DIFF
--- a/benchmarks/src/main/scala/io/finch/benchmarks/service/model.scala
+++ b/benchmarks/src/main/scala/io/finch/benchmarks/service/model.scala
@@ -25,7 +25,7 @@ class UserDb {
 
   def all: Future[List[User]] = Future.value(users.synchronized(users.values.toList))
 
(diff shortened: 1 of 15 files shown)


vkostyukov:
This looks awesome! Thank you, @travisbrown!

Please, don't worry about the docs. I have already started a branch so I will update all the docs at once. But we definitely need some tests.

Have a nice vacation!

travisbrown:
I've just moved some things around (in the spirit of #238) and added a test that confirms that you can put together a coproduct router with all of the applicable result types and convert it into a `Endpoint[HttpRequest, HttpResponse]`.

vkostyukov:
Superb! Thanks a lot @travisbrown! I think it's ready to go alive.

PR #277, opened 2015-05-31
Finch client Spike - Do Not Merge

penland365:
This is a quick spike around what I think a `finch-client` could look like, at least for my typical use cases.  It's heavily biased towards a `RESTFul` style.  A small syntax showing useage is in the `README` for the package.  Quick synopsis:

1. Create a Finch request w/ a host and an endpoint (finagle.Path)
2. Make the actual Http request w/ an Http `Verb` => `Get(finchRequest)`
3. `Verb` returns a `String Xor Resource` from `Cats` to handle failures better ( super shallow w/ this right now )
4. A `Resource` is simply and HttpResponse Status like `Ok` parameterized on the response content type => `Ok(twitter.io.Buf)` or a `Ok(String)` or `Ok(argonaut.Json)`
5. Right now I've got two implicit DecodeResource TypeClasses for `String` and `json.Argonaut`.

Thoughts on this style?  Love it?  Hate it?  It needs a lot of work, but this is much closer to what I envision that my partial baked gist I put in gitter and then yanked out on Friday.

### DIFF
--- a/client/README.md
+++ b/client/README.md
@@ -1,0 +1,14 @@
+The `finch-client` module provides an attempt at a finch styled rest-client.
+
+Useage
+------------
+```scala
+val connection = FinchRequest(Connection("api.github.com"), Path("/users/penland365"))
+val resource = for {
+    i <- Get(rReq) // Http Verb is the function to send the request
+    j <- Future.value(i.toList.head) // Get returns a String Xor Resource
+    k <- Future.value(j.as[Json]) // requires a DecodeResource TypeClass
+    l <- Future.value(k.getOrDefault(YourDefaultResourceHere))
+  } yield l
+val = Ok(headerMap, argonaut.Json)
+```
--- a/client/src/main/scala/io/finch/client/DecodeResource.scala
+++ b/client/src/main/scala/io/finch/client/DecodeResource.scala
@@ -1,0 +1,28 @@
+package io.finch
+package client
+
+import argonaut._, Argonaut._
+import com.twitter.io.Buf
+import com.twitter.io.Buf.Utf8
+
+trait DecodeResource[A] {
+  def apply(buf: Buf): A 
+}
+
+object DecodeResource {
+
+  implicit def decodeStringResource: DecodeResource[String] = new DecodeResource[String] {
+    def apply(buf: Buf): String = asString(buf)
+  }
+  
+  implicit def decodeJsonResource: DecodeResource[Json] = new DecodeResource[Json] {
+    def apply(buf: Buf): Json = asString(buf)
+      .parseOption
+      .getOrElse(jEmptyObject)
+  }
+  
+  def asString(buf: Buf): String = Utf8.unapply(buf) match {
+    case Some(x) => x
+    case None    => ""
+  }
+}
--- a/client/src/main/scala/io/finch/client/resources.scala
+++ b/client/src/main/scala/io/finch/client/resources.scala
@@ -1,0 +1,27 @@
+package io.finch
+package client
+
+import com.twitter.finagle.httpx.{HeaderMap, Response}
+import com.twitter.io.Buf
+
+sealed trait Resource[A] { 
+  val headers: HeaderMap
+  val content: A
+  def as[B](implicit ev: A => Buf, d: DecodeResource[B]): Resource[B] = 
+    Ok(headers, d(content))
+}
+
+object Resource {
+  def fromResponse(resp: Response): Resource[Buf] = resp.statusCode match {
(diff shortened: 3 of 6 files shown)


client/src/main/scala/io/finch/client/DecodeResource.scala
```diff
@@ -0,0 +1,28 @@
...
+import com.twitter.io.Buf
+import com.twitter.io.Buf.Utf8
+
+trait DecodeResource[A] {
+  def apply(buf: Buf): A 
+}
+
+object DecodeResource {
+
+  implicit def decodeStringResource: DecodeResource[String] = new DecodeResource[String] {
```
travisbrown: Might as well make these `val`s, no?

client/src/main/scala/io/finch/client/resources.scala
```diff
@@ -0,0 +1,27 @@
+package io.finch
+package client
+
+import com.twitter.finagle.httpx.{HeaderMap, Response}
+import com.twitter.io.Buf
+
+sealed trait Resource[A] { 
+  val headers: HeaderMap
+  val content: A
+  def as[B](implicit ev: A => Buf, d: DecodeResource[B]): Resource[B] =
```
travisbrown: Maybe I don't understand the intended usage yet, but implicit functions make me nervous. Would it be possible to provide this as a `def toBuf(a: A): Buf` on the trait?

client/README.md
```diff
@@ -0,0 +1,14 @@
+The `finch-client` module provides an attempt at a finch styled rest-client.
+
+Useage
```
travisbrown: "Usage" would be more standard.

travisbrown:
I'm just starting to look at this, but I'm not sure about introducing a Cats dependency just for a biased disjunction type. Any reason you don't just fail in the future with some type representing a response failure? Then you could use the standard future failure handling mechanisms.

client/src/main/scala/io/finch/client/DecodeResource.scala
```diff
@@ -0,0 +1,28 @@
...
+    def apply(buf: Buf): String = asString(buf)
+  }
+  
+  implicit def decodeJsonResource: DecodeResource[Json] = new DecodeResource[Json] {
+    def apply(buf: Buf): Json = asString(buf)
```
travisbrown: I think I'd pretty much always prefer an explicit failure (with information about the parse error) to an empty object in the case of bad input.

client/src/main/scala/io/finch/client/DecodeResource.scala
```diff
@@ -0,0 +1,28 @@
...
+      .parseOption
+      .getOrElse(jEmptyObject)
+  }
+  
+  def asString(buf: Buf): String = Utf8.unapply(buf) match {
```
travisbrown: `.getOrElse("")` would be a little clearer, but again if `Utf8.unapply` fails I'd generally want to know it.

penland365:
1. I have no issues removing the right biased disjunction and using a separate type to represent a failure - using `Xor` came from looking at Dispatch and realizing they return a `Future[Either[FailureType, ResultType]]` and trying to play with something similar. 

2. In continuation of 1, I should have been explicit in the failure information - honestly I just forgot about it while I was spiking this.

3. implicit functions make me nervous - the reasoning for them was an attempt at utilizing TypeClasses to handle decoding a [com.io.twitter.Buf](https://github.com/twitter/util/blob/master/util-core/src/main/scala/com/twitter/io/Buf.scala) to some `A`. In my head, that would allow us parameterize a Resource on the ContentType itself. I'm sure there are better ways to do this though - it's just that most implementations of TypeClasses I've seen use implicits.

client/src/main/scala/io/finch/client/DecodeResource.scala
penland365: Agreed

client/src/main/scala/io/finch/client/DecodeResource.scala
penland365: Agreed.

client/src/main/scala/io/finch/client/DecodeResource.scala
penland365: Yes, they should be `val`s.

travisbrown:
The problem with implicit functions is that they can get applied behind your back, just like implicit conversions defined as `def`s (this is why we've got things like the `ToRequest` type class in my last PR—it's isomorphic to a `A => Request`, but you can put one into implicit scope without `A`s being treated as `Request`s everywhere).

PR #278, opened 2015-05-31
Convert value router into an endpoint

vkostyukov:
This is WIP PR.

My goal is to allow converting value routers `RouterN[A]` to endpoints (see test case). The implicit conversion I wrote seems to do the trick, but it's not applied by a compiler by some reason. Still trying to understand what's missing.

@travisbrown, would totally appreciate you help here.


travisbrown:
Taking a look now—thanks for fixing my stupid mistakes!

travisbrown:
A slightly cleaner way to write this would be to use the case directly:

```scala
implicit def routerToEndpoint[R: ToRequest, A](
  router: RouterN[A]
)(implicit
  polyCase: EncodeAll.Case.Aux[A, Service[R, HttpResponse]]
): Endpoint[R, HttpResponse] = router.map(polyCase)
```

Both this and your version work if the router is typed as a `RouterN[String]`:

```scala
val router: Router[String] = Get / "foo" /> "bar"
val e: Endpoint[HttpRequest, HttpResponse] = router
```

But without the intermediate definition it'll try to force the router into the expected type while inferring its type. There are ways to work around this, but I'm in an airport right now, and I'm not sure I'm going to be able to get a solution together before I have to get on this plane.

vkostyukov:
Cool! Please don't worry, I will try to workaround this and will update the PR. Have a safe flight!

travisbrown:
Thanks!

Here's a quick minimized example of the kind of problem we're seeing here:

```scala
scala> implicit def foo(xs: => List[String]): List[Int] = xs.map(_.toInt)
foo: (xs: => List[String])List[Int]

scala> val strings = List("1", "2")
strings: List[String] = List(1, 2)

scala> val ints: List[Int] = strings
ints: List[Int] = List(1, 2)

scala> val ints: List[Int] = List("1", "2")
<console>:8: error: type mismatch;
 found   : String("1")
 required: Int
       val ints: List[Int] = List("1", "2")
                                  ^
<console>:8: error: type mismatch;
 found   : String("2")
 required: Int
       val ints: List[Int] = List("1", "2")
                                       ^
```

I actually can't remember now whether I've found a better solution to this kind of thing than "don't use implicit conversions".

vkostyukov:
I went ahead and refactored the coproduct routers to be implicitly convertible into services, not endpoints. I was hoping that we can reduce the API surface around endpoints since my ultimate goal is to replace them with endpoints discussed in #254. Somehow, it helped to workaround the issue with implicit conversion for value routers. So, this is ready to be reviewed.

travisbrown:
I don't have a strong opinion on the service vs. endpoint question, although I'm wary about more magic happening in implicit conversions. On the topic why this fixes the `valueRouterToX` issue, take the following modification of my example above:

```scala
scala> implicit def foo(xs: List[String]): Set[Int] = xs.map(_.toInt).toSet
foo: (xs: List[String])Set[Int]

scala> val strings = List("1", "2")
strings: List[String] = List(1, 2)

scala> val ints: Set[Int] = strings
ints: Set[Int] = Set(1, 2)

scala> val ints: Set[Int] = List("1", "2")
ints: Set[Int] = Set(1, 2)
```

The last line works (where `List[Int]` as target didn't) because the compiler knows it can't get a `Set` from a `List`, and so it allows `List("1", "2")` to have its type inferred correctly before looking at the implicit conversion. The same thing happens when you make the target `Service` instead of `Endpoint` (which is still just a `RouterN`).

vkostyukov:
Hm.. I see. That's interesting. Thanks for the explanation, Travis!

Re endpoint vs. service. I was thinking about getting the rid of that intermediate step in `Router[Coproduct] => Endpoint[HttpRequset, HttpResponse] => Service[HttpRequest, HttpResponse]` since I don't really see any reason to have that endpoint in the middle. It had some sense before coproduct routers, but now it seems redundant. Basically, the only possible operation (that makes sense) on an HTTP endpoint is to converting it into a service. So, I decided to just remove that step.

Although, I totally agree, that supporting that conversion (i.e., `Router[Coproduct] => Endpoint[HttpRequest, HttpResponse]`) fits very nicely into the current concept: it feels more compatible with the current style. But, I think we should start a new style with coproduct routers and I belive we can safely do that since it's a new feature.

@travisbrown what do you think about this direction?

travisbrown:
I agree that skipping the `Endpoint` step has good arguments in its favor.

One counter-proposal (to play devil's advocate): what about a `toService` method on `Router` that requires implicit evidence that the result can be converted to a Finagle service? We could also provide a `Finch.serve(":8080", api)` helper that would allow serving appropriately-typed routers directly.

vkostyukov:
Hm.. The `toService` method makes sense for me. I also like the idea about `Finch.serve()` but I'm not sure if we'll able to maintain this: as far as I understand we would need to support/duplicate all the `Httpx.withX` methods.

vkostyukov:
From my understanding `toService` API woul look as folllows:

```scala
val api: Service[HttpRequest, HttpResponse] = (
  postUser :+: postTicket :+: getUser :+: getTicket
).toService
```

@travisbrown is it what you had in mind?

travisbrown:
@vkostyukov Right, that's what I was thinking. It'd simplify a lot of the implementation and make the boundary between routers and services a little clearer, and the syntactic cost is pretty small.

vkostyukov:
I'm going to update this PR today to use new `toService` style.

vkostyukov:
I've added the `toService` method and decided to eliminate the `RouterNConversions` trait so everything is in `RouterN` companion object.

travisbrown:
Looks generally good to me, but see my #282 for a slightly different approach that I find a little cleaner (especially on the error message front).

vkostyukov:
@travisbrown do you think we can merge this?

travisbrown:
+1 from me!

PR #282, opened 2015-06-02
Add ToService type class for better error messages

travisbrown:
This is just a proposal, but moving the `toService` method from enrichment classes to `RouterN` itself via a new type class makes it easier to provide nice error messages. Now instead of this:

```scala
scala> case class X(i: Int)
defined class X

scala> (Get / "foo" /> X(100).toFuture).toService
<console>:28: error: could not find implicit value for parameter folder: shapeless.ops.coproduct.Folder.Aux[io.finch.route.RouterN.EncodeAll.type,shapeless.:+:[com.twitter.util.Future[X],shapeless.CNil],com.twitter.finagle.Service[R,io.finch.HttpResponse]]
       (Get / "foo" /> X(100).toFuture).toService
                                        ^
```

The user sees this:

```scala
scala> case class X(i: Int)
defined class X

scala> (Get / "foo" /> X(100).toFuture).toService
<console>:28: error: You can only convert a router into a Finagle service from R to an HttpResponse if R can be
converted into an HttpRequest, and if the result type of the router is one of the following:

  * An HttpResponse
  * A value of a type with an EncodeResponse instance
  * A future of an HttpResponse
  * A future of a value of a type with an EncodeResponse instance
  * A RequestReader that returns a value of a type with an EncodeResponse instance
  * A Finagle service that returns an HttpResponse
  * A Finagle service that returns a value of a type with an EncodeResponse instance
  * A coproduct made up of some combination of the above

com.twitter.util.Future[X] does not satisfy the requirement. You may need to provide an EncodeResponse instance for
com.twitter.util.Future[X] (or for some part of com.twitter.util.Future[X]).

       (Get / "foo" /> X(100).toFuture).toService
                                        ^
```

### DIFF
--- a/core/src/main/scala/io/finch/route/RouterN.scala
+++ b/core/src/main/scala/io/finch/route/RouterN.scala
@@ -28,7 +28,6 @@ import io.finch._
 import io.finch.request._
 import io.finch.response._
 import shapeless._
-import shapeless.ops.coproduct.Folder
 
 /**
  * A router that extracts some value of the type `A` from the given route.
--- a/core/src/main/scala/io/finch/route/RouterN.scala
+++ b/core/src/main/scala/io/finch/route/RouterN.scala
@@ -129,6 +129,11 @@ trait RouterN[+A] { self =>
 
   // A workaround for https://issues.scala-lang.org/browse/SI-1336
   def withFilter(p: A => Boolean): RouterN[A] = self
+
+  /**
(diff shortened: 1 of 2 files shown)


vkostyukov:
Omg. This looks fantastic! I like this approach very much. Thanks, Travis!

vkostyukov:
Merging this.

PR #283, opened 2015-06-02
Change EncodeResponse to treat response body as Array[Byte]

yanana:
Finch's current `EncodeResponse` implementation imposes HTTP response to be a String. This limitation causes binary content to be encoded as a String. Therefore bytes invalid as UTF-8 string might be silently replaced to other bytes (this behaviour is not specified by the language spec. see [here](http://docs.oracle.com/javase/8/docs/api/java/lang/String.html#String-byte:A-java.lang.String-)). To get rid of this unwanted behaviour, I made these changes:

1. Change `EncodeResponse#apply` to return `Array[Byte]` instead of `String`
2. Now set Content-Type header with `contentType=` instead of `setContentType` because the latter method decides the content to be a text and appends `;charset=utf-8` to the header.

**NOTE:** This is a breaking change since it may affect all the user-defined `EncodeResponse`s.

### DIFF
--- a/argonaut/src/main/scala/io/finch/argonaut/package.scala
+++ b/argonaut/src/main/scala/io/finch/argonaut/package.scala
@@ -51,5 +51,5 @@ package object argonaut {
    * @return Create a Finch ''EncodeJson'' from an argonaut ''EncodeJson''
    */
   implicit def encodeArgonaut[A](implicit encode: EncodeJson[A]): EncodeResponse[A] =
-    EncodeResponse("application/json")(encode.encode(_).nospaces)
+    EncodeResponse.fromString[A]("application/json")(encode.encode(_).nospaces)
 }
--- a/core/src/main/scala/io/finch/response/package.scala
+++ b/core/src/main/scala/io/finch/response/package.scala
@@ -27,6 +27,8 @@ package io.finch
 import com.twitter.finagle.Service
 import com.twitter.finagle.httpx.path.Path
 import com.twitter.finagle.httpx.Status
+import com.twitter.io.Buf
+import com.twitter.io.Buf.Utf8
 import com.twitter.util.Future
 
 /**
@@ -151,25 +153,35 @@ package object response {
    * An abstraction that is responsible for encoding the response of type `A`.
    */
   trait EncodeResponse[-A] {
-    def apply(rep: A): String
+    def apply(rep: A): Buf
     def contentType: String
+    def charset: Option[String] = Some("utf-8")
   }
 
   object EncodeResponse {
     /**
      * Convenience method for creating new [[io.finch.response.EncodeResponse EncodeResponse]] instances.
      */
-    def apply[A](ct: String)(fn: A => String): EncodeResponse[A] = new EncodeResponse[A] {
-      def apply(rep: A): String = fn(rep)
-      def contentType: String = ct
-    }
+    def apply[A](ct: String, cs: Option[String] = Some("utf-8"))(fn: A => Buf): EncodeResponse[A] =
+      new EncodeResponse[A] {
+        override def apply(rep: A): Buf = fn(rep)
+        override def contentType: String = ct
+        override def charset: Option[String] = cs
+      }
+
+    /**
+     * Convenience method for creating new [[io.finch.response.EncodeResponse EncodeResponse]] instances
+     * that treat String contents.
+     */
+    def fromString[A](ct: String, cs: Option[String] = Some("utf-8"))(fn: A => String): EncodeResponse[A] =
+      apply(ct, cs)(fn andThen Utf8.apply)
   }
 
   /**
    * An abstraction that is responsible for encoding the response of a generic type.
    */
   trait EncodeAnyResponse {
-    def apply[A](rep: A): String
+    def apply[A](rep: A): Buf
     def contentType: String
   }
 
@@ -179,7 +191,7 @@ package object response {
    */
   implicit def anyToConcreteEncode[A](implicit e: EncodeAnyResponse): EncodeResponse[A] =
     new EncodeResponse[A] {
-      def apply(rep: A): String = e(rep)
+      def apply(rep: A): Buf = e(rep)
       def contentType: String = e.contentType
     }
 
@@ -198,11 +210,11 @@ package object response {
   /**
    * Allows to pass raw strings to a [[ResponseBuilder]].
    */
-  implicit val encodeString: EncodeResponse[String] = EncodeResponse("text/plain")(s => s)
+  implicit val encodeString: EncodeResponse[String] = EncodeResponse.fromString[String]("text/plain")(identity)
 
   /**
-   * Allows to pass `Byte[Array]` to a [[ResponseBuilder]].
+   * Allows to pass `Buf` to a [[ResponseBuilder]].
    */
-  implicit val encodeByteArray: EncodeResponse[Array[Byte]] =
-    EncodeResponse("application/octet-stream") { a => new String(a, "UTF-8") }
+  implicit val encodeBuf: EncodeResponse[Buf] =
+    EncodeResponse("application/octet-stream", None)(identity)
 }
--- a/core/src/test/scala/io/finch/response/ResponseBuilderSpec.scala
+++ b/core/src/test/scala/io/finch/response/ResponseBuilderSpec.scala
@@ -24,6 +24,8 @@
 package io.finch.response
 
 import com.twitter.finagle.httpx.{Status, Cookie}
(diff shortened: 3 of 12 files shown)


core/src/main/scala/io/finch/response/package.scala
```diff
@@ -151,25 +152,28 @@ package object response {
    * An abstraction that is responsible for encoding the response of type `A`.
    */
   trait EncodeResponse[-A] {
-    def apply(rep: A): String
+    def apply(rep: A): Array[Byte]
     def contentType: String
+    def charset: Option[String] = None
```
vkostyukov: Can we promote "UTF-8" as a default charset?

argonaut/src/main/scala/io/finch/argonaut/package.scala
```diff
@@ -51,5 +51,5 @@ package object argonaut {
    * @return Create a Finch ''EncodeJson'' from an argonaut ''EncodeJson''
    */
   implicit def encodeArgonaut[A](implicit encode: EncodeJson[A]): EncodeResponse[A] =
-    EncodeResponse("application/json")(encode.encode(_).nospaces)
+    EncodeResponse("application/json")(encode.encode(_).nospaces.getBytes("UTF-8"))
```
travisbrown: Maybe we could add a `EncodeResponse.fromString("application/json")(whatever)` to clean this up a bit?

core/src/main/scala/io/finch/response/package.scala
```diff
@@ -151,25 +152,28 @@ package object response {
    * An abstraction that is responsible for encoding the response of type `A`.
    */
   trait EncodeResponse[-A] {
-    def apply(rep: A): String
+    def apply(rep: A): Array[Byte]
```
travisbrown: What about using `com.twitter.util.Buf` here (for the immutability, if nothing else)?

core/src/main/scala/io/finch/response/package.scala
yanana: I will. And I'm going to add a documentation that when treating binary contents, charset must be set `None` explicitly.

core/src/main/scala/io/finch/response/package.scala
vkostyukov: I also like the idea of using `Buf` instead of `Array[Byte]`.

yanana: I'll try to use `Buf` instead.

argonaut/src/main/scala/io/finch/argonaut/package.scala
yanana: Is it something like this? 
```scala
def fromString[A](ct: String, cs: Option[String] = Some("UTF-8"))(fn: A => String):  EncodeResponse[A] =
    new EncodeResponse[A] {
      def apply(rep: A): Array[Byte] = fn(rep).getBytes("UTF-8")
      def contentType: String = ct
      override def charset: Option[String] = cs
    }
```

vkostyukov: Perhaps, we don't need to pass charset since it's always "UTF-8" for strings.

core/src/test/scala/io/finch/response/ResponseBuilderSpec.scala
```diff
@@ -1,71 +1,98 @@
-/*
```
vkostyukov: Not sure what happened with this file, but it looks like a problem with either tabs of new lines. @yanana are you on Mac?

argonaut/src/main/scala/io/finch/argonaut/package.scala
yanana: Actually, the `cs` parameter is not considererd to encode bytes into String. The parameter affects only a `Content-Type` header value. Do situations like "I want to return UTF-8 text content, but don't want to set `;charset=utf-8` to `Content-Type` header." exist?

Using `Buf` and omitting `cs` parameter, above code would be:
```scala
def fromString[A](ct: String)(fn: A => String):  EncodeResponse[A] =
    new EncodeResponse[A] {
      def apply(rep: A): Buf = Utf8(fn(rep))
      def contentType: String = ct
      override def charset: Option[String] = Some("UTF-8")
    }
```

core/src/test/scala/io/finch/response/ResponseBuilderSpec.scala
yanana: I'm using Mac. Only the file's line ending was CRLF, but other files seem to have LF ending.

yanana:
Update done!

travisbrown:
This is looking really good to me—thanks, @yanana!

core/src/main/scala/io/finch/response/package.scala
```diff
@@ -151,25 +153,39 @@ package object response {
...
+        def apply(rep: A): Buf = fn(rep)
+        def contentType: String = ct
+        override def charset: Option[String] = cs
+      }
+
+    /**
+     * Convenience method for creating new [[io.finch.response.EncodeResponse EncodeResponse]] instances
+     * that treat String contents.
+     */
+    def fromString[A](ct: String, cs: Option[String] = Some("utf-8"))(fn: A => String): EncodeResponse[A] =
```
vkostyukov: Isn't it possible to this in terms of apply? Like `apply(ct, cs)(a => Utf8(fn(a)))`.

core/src/main/scala/io/finch/response/package.scala
```diff
@@ -151,25 +153,39 @@ package object response {
...
      */
-    def apply[A](ct: String)(fn: A => String): EncodeResponse[A] = new EncodeResponse[A] {
-      def apply(rep: A): String = fn(rep)
-      def contentType: String = ct
-    }
+    def apply[A](ct: String, cs: Option[String] = Some("utf-8"))(fn: A => Buf): EncodeResponse[A] =
+      new EncodeResponse[A] {
+        def apply(rep: A): Buf = fn(rep)
+        def contentType: String = ct
+        override def charset: Option[String] = cs
```
vkostyukov: For consistency, can we either add `override` to all the methods or remove it from `charset`?

core/src/main/scala/io/finch/response/package.scala
```diff
@@ -151,25 +153,39 @@ package object response {
...
+
+    /**
+     * Convenience method for creating new [[io.finch.response.EncodeResponse EncodeResponse]] instances
+     * that treat String contents.
+     */
+    def fromString[A](ct: String, cs: Option[String] = Some("utf-8"))(fn: A => String): EncodeResponse[A] =
+      new EncodeResponse[A] {
+        def apply(rep: A): Buf = Utf8(fn(rep))
+        def contentType: String = ct
+        override def charset: Option[String] = cs
```
vkostyukov: The same question about `override`.

core/src/main/scala/io/finch/response/package.scala
```diff
@@ -198,11 +214,11 @@ package object response {
...
    */
-  implicit val encodeString: EncodeResponse[String] = EncodeResponse("text/plain")(s => s)
+  implicit val encodeString: EncodeResponse[String] = EncodeResponse.fromString("text/plain")(identity)
 
   /**
    * Allows to pass `Byte[Array]` to a [[ResponseBuilder]].
    */
-  implicit val encodeByteArray: EncodeResponse[Array[Byte]] =
-    EncodeResponse("application/octet-stream") { a => new String(a, "UTF-8") }
+  implicit val encodeByteArray: EncodeResponse[Buf] =
```
vkostyukov: `encodeBuf`?

vkostyukov:
This looks pretty solid to me: just a couple of nitpicks. Thanks @yanana!

Could you please also rebase the PR agains master?

core/src/main/scala/io/finch/response/package.scala
yanana: Seems better!

core/src/main/scala/io/finch/response/package.scala
yanana: Though I don't know the reason, removing `override` from `charset` leads to compile error... :octopus:  So I'll add `override` to all methods.

core/src/main/scala/io/finch/response/package.scala
yanana: Fixed. Thanks!

yanana:
OK, rebased!

vkostyukov:
Nice! I'm merging this. Thanks @yanana!

BenWhitehead:
Great contribution @yanana!

PR #284, opened 2015-06-02
Fixed the broken codecov shield

rpless:
Also a good test to see if codecov.io PR commenting is on. We need to opt-in to that feature, so if we don't get a comment, it might not be on.

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,7 +9,7 @@ close as possible to the bare metal Finagle API.
 Badges
 ------
 [![Build Status](https://img.shields.io/travis/finagle/finch/master.svg)](https://travis-ci.org/finagle/finch)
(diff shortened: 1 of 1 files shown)


vkostyukov:
Cool! Why aren't we getting coverage reports already?

rpless:
@vkostyukov Unlike Coveralls, you need to opt-in to PR comments (they're also highly configurable). On the project page (https://codecov.io/github/finagle/finch) there is a features dropdown all the way to the right. Click on the Pull Request Comments Feature and turn it on.

vkostyukov:
Yeah. I've already checked it. Thanks, Ryan!

rpless:
Interesting. I took a look at the one here (https://github.com/rpless/finch/pull/2). Looks like there's ~5 minutes between Travis finishing and comment being posted (obviously not the most scientific way to determine this, but shows there might be a bit of a delay). So we should start seeing them.

PR #285, opened 2015-06-03
Disable code coverage for benchmarks

vkostyukov:
Do I understand this correctly that we don't really need to cover benchmarks?


travisbrown:
It would be nice to have coverage for the logic, but not the benchmark runners. That might take a bit of rearrangement, though.

travisbrown:
Just thinking a little more about this—it would be cool to have tiered coverage reports, where you get information about example projects and benchmarks without having them included in your top-level numbers.

vkostyukov:
I'm abondoning this.

PR #288, opened 2015-06-03
Deprecate finch-jawn (and update docs index)

### LINKED ISSUE #286: Deprecate finch-jawn
We discussed this on Gitter but it doesn't look like it made it into an issue. I agree that it might be a good idea to deprecate finch-jawn in 0.7.0 (it's not really the same kind of library as the others). I would like to experiment with using Jawn as a parser for e.g. finch-argonaut at some point to see if that buys us any performance improvement, but I can create a separate issue for that.

travisbrown:
Fixes #286.


vkostyukov:
Superb! Thanks @travisbrown!

PR #290, opened 2015-06-03
Fix benchmarks and clean up demo imports

travisbrown:
A recent change from `def` to `val` in the definition of the benchmark deletion service meant the benchmarks weren't running properly (it'd be nice to have this tested). I'm also including a bit of completely unrelated import clean-up.


vkostyukov:
Ah. Sorry, I changed it :(

PR #291, opened 2015-06-03
Fix #238: move implicit conversion toOptionalRule to ValidationRule companion object

### LINKED ISSUE #238: Revise all the implicit conversions

vanyacohen:
Also fixes a typo in the name of an implicit def (booleanToMatcher, route package object).


travisbrown:
Looks good to me, thanks!

vkostyukov:
Great! Merging this!

Issue #289, opened 2015-06-03
Move class definitions out of package objects


travisbrown:
I mentioned this on Gitter but am creating an issue for more inclusive discussion.

I think it'd be a good idea to move all class (and trait and object) definitions that aren't implicit classes out of package objects, since having `package$` prefixes on the JVM class names makes a number of things inconvenient (`javap`, Java interop, etc.).

This would be a breaking change but mostly (entirely?) source-compatible for Scala users. It'd be pretty straightforward to implement, and could even happen before 0.7.0 if people are on board.

Any thoughts?

BenWhitehead:
+1

travisbrown:
`-Xlint` says we're okay here after #292 and #295.

PR #292, opened 2015-06-04
Move class definitions from request package object to companion objects

travisbrown:
This is a major reorganization of the `request` package motivated by the reasons I mention in #289.

Moving implicit classes in particular to companion objects makes prioritization a little simpler (no more `LowPriorityX` stuff in `request`) and may speed up compilation a bit.

If this looks good to everyone we can give the same treatment to `response` and `route`.


vkostyukov:
:+1: 

Looks great! Thanks!

travisbrown:
Cool, I'll aim to wrap up `response` and `route` over lunch.

PR #293, opened 2015-06-04
Turn on -Xlint

travisbrown:
`-Xlint` actually provides "it is not recommended to define classes/objects inside of package objects" warnings (and no other new warnings, at a glance), so we might as well turn it on now that we're moving away from that pattern.


vkostyukov:
I'm surprised it was disabled. Thanks!

PR #294, opened 2015-06-04
Kill micro

vkostyukov:
The `micro` package was marked "experimental" so I belive we have rights to do that.


travisbrown:
Sounds reasonable to me, but is there a strong reason not to let it live out the deprecation cycle?

vkostyukov:
Makes sense. Let's keep this open until 0.7.0.

travisbrown:
I'd be happy to pull the trigger on this now that 0.7.0 is out.

vkostyukov:
I'm about to submit release notes (tag will be attached to HEAD). After that we can start deleting stuff!

vkostyukov:
Let's kill it!

PR #295, opened 2015-06-04
Finish moving class definitions out of package objects

travisbrown:
This finishes the work started in #292 and should complete #289.

There are a couple of judgment calls here—for example I created new `Matcher` and `Extractor` files even though their contents could reasonably go in `Router0` and `RouterN` respectively, because I tend to prefer smaller files, but I'm happy to change any of this.


vkostyukov:
Love it! Thanks @travisbrown!

PR #299, opened 2015-06-04
Add integration tests for service benchmarks

travisbrown:
This gets the test coverage for finagle-benchmarks up to a reasonable level (and gives us some protection against breaking the benchmarks at the same time). We could still disable coverage reporting for finagle-benchmark (as in #285), but I think I'd vote not to.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,6 +7,6 @@ jdk:
   - openjdk7
   - oraclejdk7
   - oraclejdk8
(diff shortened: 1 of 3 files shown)


travisbrown:
Ugh, Travis is failing on half of its runs (and not a half that makes any sense). I'm about to head out, but I'll try to figure out what's up later this evening.

travisbrown:
Ah, just telling the integration tests not to run in parallel seems to fix it (I'd thought this was the default setting for `IntegrationTest`, but it turns out it's not).

vkostyukov:
Awesome! Thanks @travisbrown!

PR #300, opened 2015-06-05
Update docs to an actual state as for 0.7.0

vkostyukov:
WIP: Pushing the branch to make sure it renders correctly.


vkostyukov:
Docs are in progress. I've finished `request.md` and `response.md`. `route.md` is on my list of todos.

vkostyukov:
This is ready to be reviewed.

CONTRIBUTING.md
```diff
@@ -12,27 +12,30 @@ elsewhere.
...
 
 That said, the Scala style checker `sbt scalastyle` should pass on the code. 
 
 ## Write Tests
-Finch uses [ScalaTest][2] with the following settings:
+Finch uses both [ScalaTest][2] and [ScalaCheck][3] with the following settings:
 
-* Every test should be a `FlatSpec` with `Matchers` mixed in
+* Every test should be a `FlatSpec` with `Matchers` and `Checkers` mixed in
 * An assertion should be done with `x shouldBe y`
```
travisbrown: `check` expects its argument to return a boolean (or something similar—not `Unit`, in any case), which means that assertions in properties have to be written with `===`.

CONTRIBUTING.md
```diff
@@ -12,27 +12,30 @@ elsewhere.
...
 * Exceptions should be intercepted with `an [Exception] shouldBe thrownBy(x)`
 
 ## Write Docs
-Write clean and simple docs in the `docs.md` file.
+Write clean and simple docs in the `docs` folder.
 
 ## Submit a PR
 * PR should be submitted from a separate branch (use `git checkout -b "fix-123"`)
-* PR should contain only one commit (use `git commit --amend` and `git --force push`)
+* PR should contain only one commit (use `git commit --amend` and `git --force push` or [squash][4] existing commits into one)
```
travisbrown: Maybe "generally"? I think there are cases where keeping multiple commits is useful. I could go either way, though.

vkostyukov: Yeah. There are exceptions though. I will add "generally".

vkostyukov:
Thanks @travisbrown! I've updated the PR.

docs/route.md
```diff
@@ -142,8 +148,49 @@ endpoint(request) handle {
 }
 ```
 
+### Coproduct Routers
+
+The `|` compositor is pretty useful for the simple cases when both underlying types `A` and `B` may be substituted with
+a single super type `C` such that `C >: A` and `C >: B` (usually that means that `A =:= B`). Although, it should be also
+possible to compose two routers of completely different types. This is pretty duable the `:+:` compositor that composes
```
travisbrown: s/duable/doable with/

docs/route.md
```diff
@@ -142,8 +148,49 @@ endpoint(request) handle {
...
+two routers into a `Router[C <: Coproduct]`, where `Coproduct` [Shapeless' disjoint union type][2].
```
travisbrown: "`Coproduct` is"

travisbrown:
This looks generally good to me!—I'm happy to make a closer pass over the rest, but probably won't be able to get to it until tomorrow evening.

vkostyukov:
I decided to go with 12 chars limit since IDEA doesn't support ([yet](https://youtrack.jetbrains.com/issue/IDEA-59662)) setting the customized limit for Markdown-files. Although, I would be happy to reduce the limit later.

vkostyukov:
Let's merge this then!

PR #301, opened 2015-06-06
Fix implicit widening warnings

travisbrown:
I should have caught these warnings in #296.


vkostyukov:
Should we rebase this to make the build passing?

travisbrown:
Travis rebuilds automatically after other stuff gets merged to the target branch, so this is ready to go (I'm happy to rebase, too, though).

vkostyukov:
Interesting. I didn't notice that! Merging this in. Thanks @travisbrown!

PR #302, opened 2015-06-06
Expand benchmarks and improve integration tests

travisbrown:
This breaks the benchmarking example into smaller pieces and makes it easier to test (which together with the Java options that I borrowed from Finagle SBT script and a few other changes should make the integration tests less likely to fail).

PR #303, opened 2015-06-07
Fix problem with benchmark service definition

travisbrown:
A fix for the problem mentioned in #302. I've also changed the `benchmarks` project configuration so that the benchmarks themselves are run with the extra JVM arguments.


vkostyukov:
Superb!

PR #304, opened 2015-06-07
Minor clean-up for clarity

travisbrown:
This is super trivial, but I noticed that there's some unnecessary noise in these pattern matches, and that the indentation could be a little clearer.


vkostyukov:
:+1:

PR #305, opened 2015-06-07
Update to Shapeless 2.2.1

travisbrown:
I don't think any of the bug fixes and performance improvements in 2.2.1 affect us, but we might as well update before the 0.7.0 release.


vkostyukov:
Great! Thank you @travisbrown!

PR #306, opened 2015-06-07
Unify Router0 and RouterN and eliminate / case class

travisbrown:
This is a proof-of-concept and isn't ready to be merged. It unifies `Router0` and `RouterN` and gets rid of the `/` case class while leaving usage mostly unchanged (it requires some changes to the `RouterSpec` tests, but none to the example projects).

I've removed variance annotations in a few places for the sake of simplicity, but some of these could be added back if people want them.


core/src/main/scala/io/finch/route/RouterN.scala
```diff
@@ -41,9 +41,14 @@ trait RouterN[+A] { self =>
   def apply(route: Route): Option[(Route, A)]
 
   /**
+   * Attempts to match a route, but only returns any unmatched elements, not the value.
+   */
+  def exec(route: Route): Option[Route] = apply(route).map(_._1)
```
vkostyukov: I remember you were talking about using `run` instead of `apply`? Do you think we can do it now? We still can have `apply` (marked as `@deprecated`) as an alias for `run`. Then `exec` might be renamed to `runIgnoreValue` (mark it as package-private method, since we only use it for tesging). What do you think?

core/src/main/scala/io/finch/route/Router0.scala
```diff
@@ -22,95 +22,29 @@
...
 package io.finch.route
 
+import shapeless._
+
 /**
  * A router that match the given route to some predicate.
  */
-trait Router0 { self =>
-
+trait Router0 extends Router[HNil] { self =>
```
vkostyukov: Maybe we can rename it to `Matcher`?

vkostyukov: Maybe we can rename it to `Matcher`?

vkostyukov:
Woot! Codecov has started commenting PRs!

This is unbelievably cool @travisbrown! Great and very smart simplification!

core/src/main/scala/io/finch/route/RouterN.scala
travisbrown: That sounds reasonable to me.

I was thinking of creating an issue with a question about de-coupling router matching from returning the value (to help with cases like the issue I ran into in #302 that's fixed in #303)—i.e. if two routes match a request right now, both of their values will be evaluated (since this happens before the length of their leftovers are compared), which can be a little surprising. I was thinking this would happen (if at all) after 0.7.0, but it would fit in with the kind of changes you suggest.

vkostyukov: Yeah. That would be a good improvement. Let's plan this for 0.8.0 then. How about to just make `exec` package-private for now?

travisbrown:
Okay, I've renamed `Matcher` and made `exec` package-private.

A couple of other questions:

I've [asked](https://twitter.com/travisbrown/status/607185103765258240) about a better name for `Smash`, and I'm not sure it belongs in the `route` package, since it's very generic. It be nice if Shapeless included it, but I can understand why it doesn't.

Anyone know why the `*ToMatcher` implicit conversions in `route` don't work if they're moved to the `Router` companion object?

This new stuff still needs tests, and there's some commented-out stuff that needs to be deleted.

vkostyukov:
I noticed the same issue with `*ToMatcher` implicit conversions and I have no idea why is that.

travisbrown:
This is ready to go as far as I'm concerned, but I want to be sure everyone's on board with the invariantization of `Router`, etc.

vkostyukov:
I'm personally pretty happy with invariant `Router`. Let's ask other folks on Gitter.

vkostyukov:
Seems that there are no concerns about this simplification. So we should probably merge this in since I can't wait this to land in master!

Amazing job, @travisbrown!

PR #307, opened 2015-06-07
Exclude packages that are entirely deprecated from code coverage

travisbrown:
This seems reasonable to me. I'm not sure whether it actually has an effect on the coverage, but it should be a more useful number.


vkostyukov:
:+1:

PR #308, opened 2015-06-08
Some edits and rearrangement for the request docs

travisbrown:
I've just taken a stab at providing some additional context, together with a few edits here and there. I've also moved the custom request types section to the end, since it doesn't really seem important enough to lead off with.

Rendered version is available [here](https://github.com/travisbrown/finch/blob/update-docs-0.7.0/docs/request.md).


vkostyukov:
Neat! @travisbrown what do you think about having some sort of agreement on line length for docs? For now, I was using the same settings with have for code - 120 chars. Mabe we should reduce it for docs (and note that in the contributing guide)?

vkostyukov:
Perhaps, we should be rebased? Seems that I'm not able to merge this in.

travisbrown:
Yeah, sorry about the line lengths. I'll can switch to 120 and rebase when I get back to a computer.

vkostyukov:
Please, don't worry about the line lengths. I can address them in the upcoming commits. I'm thinking about reducing the limit to 100 (or even 80) for docs. So I can do change them all at once.

travisbrown:
Okay, sounds good—I've just rebased.

vkostyukov:
Nice! Thanks! In a mean time, any thoughts about the rest of the docs?

PR #309, opened 2015-06-08
Delete Finch Json

rpless:
Removed the finch json library and all documentation on it as per #233. I can also remove the `jawn` project too, if we're planning on removing that for 0.8.0.


travisbrown:
Looks good to me. And yes, I think we should remove Jawn at the same time.

vkostyukov:
Let's merge this in and then remove Jawn in a separate PR.

PR #313, opened 2015-06-08
Add some example usage notes and initial data to the demo

travisbrown:
This makes it a little easier to get started with the demo project.


vkostyukov:
Thanks @travisbrown!

PR #316, opened 2015-06-10
Fix a broken link in doc

yanana:
Fixed a broken link in routes.md.

### DIFF
--- a/docs/route.md
+++ b/docs/route.md
@@ -116,7 +116,7 @@ val users =
 ### Endpoints
 
 **Important:** endpoints are deprecated in 0.7.0 (and will be removed in 0.8.0) in favour of
-[Coproduct Routers][router.md#coproduct-routers].
+[Coproduct Routers](#coproduct-routers).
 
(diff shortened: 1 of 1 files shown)


travisbrown:
Changing the brackets looks good to me, but I think there were two errors here—it says `router.md` instead of `route.md`.

docs/route.md
```diff
@@ -116,7 +116,7 @@ val users =
 ### Endpoints
 
 **Important:** endpoints are deprecated in 0.7.0 (and will be removed in 0.8.0) in favour of
-[Coproduct Routers][router.md#coproduct-routers].
+[Coproduct Routers](router.md#coproduct-routers).
```
vkostyukov: As @travisbrown mentioned, it should be `route.md`. @yanana could you please update also fix it?

yanana:
Thanks for pointing out. I fixed two errors.

vkostyukov:
Nice! Thanks @yanana!

PR #318, opened 2015-06-11
Fix #315: Extend Smash to detect HNils

### LINKED ISSUE #315: Imporve Smash
I was experimenting with some simple idea: can we be a bit of smarter in smashing. More precisely, can we shams `HNil` and `A` (and vice versa) into just `A` not `A :: HNil`.

Here is an example:

```scala
val Router: Router[Int :: HNil] = Get / "user" / int // used to be just Router[Int]
```

It seems pretty possible to me (and it's almost workable in my branch).

@travisbrown do you have any concerns about that? What I see is a bit of problems with `/>` compositor that should be easy to fix.

vkostyukov:
This addresses #315. `Smash` now handles `HNil` as a special case of `HList`, so `smash[HNil, A] => A` and vice-versa. I also tried to get rid of boilerplate classes `RArrowN` and replaced them with Shapeless `FnToProduct`. Finally, this PR adds a type alias `Router0 = Router[HNil]`.

It's ready to be reviewed.

### DIFF
--- a/core/src/main/scala/io/finch/route/Router.scala
+++ b/core/src/main/scala/io/finch/route/Router.scala
@@ -28,6 +28,7 @@ import io.finch._
 import io.finch.request._
 import io.finch.response._
 import shapeless._
+import shapeless.ops.function.FnToProduct
 
 /**
  * A router that extracts some value of the type `A` from the given route.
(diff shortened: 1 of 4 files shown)


travisbrown:
This looks generally good to me. Instead of reintroducing `Router0`, though, we could make `Matcher` a type alias and rename the current `Matcher` class (to `Router0`)? This would have the advantage of making the more semantically meaningful name the one people work with most of the time.

The problem with using `FnToProduct` instead of the boilerplate-y `RArrowN` is that we lose some nice type inference properties—e.g. this doesn't work:

```scala
Get / int / string /> ((i, s) => s * i)
```

I'm not sure which is better—more generality and less boilerplate, or ease of use.

vkostyukov:
Yeah. Let's keep the `RrArrowX` for cases 0, 1, 2, 3. But add a general case for N.

Introducing a type alias `Router0` I was thinking more about how [parboiled2 does it](FnToProduct). If we go with treating the 0, 1, 2, 3 case in a special way, we might want to introduce user-friendly aliases. For our cases, there won't be `Router1` but something like this:

```scala
type Router0 = Router[HNil]
type Router2[A, B] = Router[A :: B :: HNil]
type Router3[A, B, C] = Router[A :: B :: C :: HNil]
```

@travisbrown what do you think about it?

Re matcher, I was going to submit an upcoming PR that refactors everything around matchers. So I was going to eliminate `Matcher`.

travisbrown:
That plan sounds reasonable to me!

vkostyukov:
Great! I'm merging this then.

Issue #232, opened 2015-04-14
Idea: Compose Routers with RequestReader's


vkostyukov:
I remember we discussed something like this previously.

The idea is to introduce combinator `?` (like query string separator) on `Router` type that allows to compose routers and request readers together:

```scala
def getUser(userId: Int, id: Int): Future[User]
val e1: Router[Int / Int / String] = Get / "users" / int / int ? param("a")
val e2: Router[Future[User]] = Get ? param("a").as[Int] ~ param("b").as[Int] /> getUser
```

vkostyukov:
I believe this idea fits well into the concept of heterogeneous routers (see #221). Combining these two concept will allow us to write some crazy code like this:

```scala
val api: Router[Int] = 
  (Get ? param("a").as[Int]) | // answers with given param a in the body
  (Get / "users" / int /> User.getById) |
  (Get / "echo" / string) // echo service
  
```

rpless:
:+1: I'd like to see this. Its been in the back of my mind for a while because Finagle has extractors for params in addition to route elements and HTTP methods.

travisbrown:
I'm planning to put some time into a demo of a Shapeless-supported heterogeneous router this weekend (in preparation for a talk at a Twitter HQ meetup in a couple of weeks)—I'll look into this as well.

vkostyukov:
Thanks @travisbrown!

This should be relatively easy to do once we unify the composite type for both `RequestReader` and `Router`. Here is what I have in mind right now:

```scala
val a: Router[Int :: String :: HNil] = Get / int ? param("a")
val b: Router[String :: String :: Boolean :: HNil] = Get / string ? param("b") & param("c").as[Boolean]
```

Note that `&` is basically an alias for `?`. This gives us a DSL that looks pretty close to just URL.

Another crazy idea that is bothering me is to merge `RequestReader` and `Router` into a single abstraction called `Endpoint`. It should be quite possible once we close this ticket.

ImLiar:
damn man, thats what I was looking for! :+1:

vkostyukov:
It should be possible to do it now since we have `Smash`.

![Hulk](http://media.giphy.com/media/ErDZ73YFYLSPC/giphy.gif)

vkostyukov:
Let me work on that.

vkostyukov:
Just merged in #319.

[...]

PR #319, opened 2015-06-14
Introduce RouterInput

vkostyukov:
This PR is a first step towards #232. In order to compose routers with request readers, we have to carry the `HttpRequest` within `Route`. This PR introduces `RouterInput` for that purpose. I also went ahead and refactored the compositors around method matches. The previous approach felt weird and didn't really fitted well into the combinators API. The new approach is based on simple functions (compositors) `get`, `post`, etc that takes a router and returns a new router that also matches the method.

This still requires some polish but it's ready for review.


rpless:
:+1: I like the direction its going. It seems like it shouldn't be too difficult to extend this to include parameters/headers/etc. It also appears like we can easily traverse the route and inspect it programmatically.

vkostyukov:
Thanks @rpless!

@travisbrown do you mind take a look?

vkostyukov:
This needs to be rebased though.

vkostyukov:
Rebased it. Would appreciate any feedback on this. /cc @travisbrown @penland365.

travisbrown:
@vkostyukov Any change that makes more illegal states unrepresentable sounds good to me, and I like this implementation.

vkostyukov:
Okay.  I decided to not deprecate method matchers (a la `Get`, `Post`, etc) for now. So we can merge this in once it's green.

PR #323, opened 2015-06-16
remove playground project and all references in the docs

rpless:


vkostyukov:
Superb! Thanks @rpless!

PR #324, opened 2015-06-16
Simplify disjunction for routers

travisbrown:
This simplifies the `:+:` method on routers by adding the smash treatment for coproducts. This means that if we write the following:

```scala
val router = (int :+: string) :+: (string :+: long)
```

We get a `Router[Int :+: String :+: String :+: Long :+: CNil]`, not a router of a nested coproduct. This means that `router.toService` works with the router above.

Before this change you'd have to write the following to accomplish the same thing:

```scala
val router =
  (int :+: string).map(_.extendRightBy[String :+: Long :+: CNil]) orElse
  (string :+: long).map(_.extendLeftBy[Int :+: String :+: CNil])
```

This PR still needs tests, etc.—I just wanted to put it up for discussion and to address @IgorWolkov's [question on Gitter](https://gitter.im/finagle/finch?at=55800d03f1cd32e97eca8222).

(Note that the change in #315 isn't relevant for coproducts, since we'll never have a `Router[CNil]`.)

### DIFF
--- a/core/src/main/scala/io/finch/route/Router.scala
+++ b/core/src/main/scala/io/finch/route/Router.scala
@@ -29,6 +29,8 @@ import io.finch.request._
 import io.finch.response._
 import shapeless._
 import shapeless.ops.function.FnToProduct
+import shapeless.ops.adjoin.Adjoin
+
 
 /**
  * A router that extracts some value of the type `A` from the given route.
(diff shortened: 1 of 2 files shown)


vkostyukov:
This is very nice!

I was about to drop you a message with idea to use `Smash`-like approach here. But you were faster :)

vkostyukov:
BTW, @travisbrown do you think should we rename `Smash` to `HListShash` for consistency?

vkostyukov:
@travisbrown Looks great! Thanks!

PR #325, opened 2015-06-16
Upgrade to Shapeless 2.2.2

travisbrown:
Shapeless 2.2.2 fixes a binary compatibility issue in 2.2.1 and may also help with dependency resolution problems we're seeing.


vkostyukov:
:+1:

Issue #297, opened 2015-06-04
Standard conventions for imports and copyright headers


travisbrown:
I've been tending to leave stuff as I find it or make it match its neighbors, but there's some inconsistency here—some files have copyright headers, others don't, some have imports divided into sections, others alphabetize their imports, etc.

My vote would be to remove all copyright headers (the project-level license is plenty) and include authorship information in the Scaladoc when needed. I also prefer plain old alphabetization for imports, but can live with other conventions.

vkostyukov:
That's an awesome idea!

Authorship information is also available per each source file on Github.

travisbrown:
Yeah, I don't think we need it in the file, but if some contributors did want that, the Scaladoc annotation would probably be a better place than a contributors list in the header.

vanyacohen:
Happy to help with this. Just a quick clarification: do we want to move the contributors to an @author annotation by default? Or just remove the header and contributors for now and add them to the Scaladoc as needed?

rpless:
One possibility instead of @author is a CONTRIBUTORS file like finagle has (https://github.com/twitter/finagle/blob/master/CONTRIBUTORS).

vkostyukov:
I think it's not necessary to have CONTRIBUTORS file. Everyone who's contributed is listed on Github.

rpless:
I'm fine with that. Did we settle on an order for imports? I prefer alphabetization too, but I'm open to other conventions.

vkostyukov:
I'm on board with simple alphabetical order to imports.

vkostyukov:
This is not fixed yet. We still have to decide what to do with imports. I just checked and it looks like scalastyle doesn't really have a rule for that.

PR #326, opened 2015-06-17
WIP: Petstore Demo

mimilei:
This is the scrappy version I write today. The service file is definitely not complete, but I tried implementing the simpler methods. I was going to start writing the CreatePet method that takes form information and updates a pet, but then I figured I'd need to use RequestReaders for this....which I'm not entirely sure how to write. There seems to be an excessive usage of the "map" function in that department. In the meantime, I'd really appreciate it if you guys could give me feedback on the horrible petCategoryReader I added. Was that in the right direction?


petstore/src/main/scala/io/finch/petstore/Main.scala
```diff
@@ -0,0 +1,131 @@
+/*
```
vkostyukov: You can omit file headers as per #297.

petstore/src/main/scala/io/finch/petstore/endpoint.scala
```diff
@@ -0,0 +1,15 @@
+package io.finch.petstore
+
+import io.finch.petstore.model._
+import io.finch.petstore.service._
+import io.finch.petstoreDemo.AuthRequest
+
+object endpoint{
+  val getAllTags: Endpoint[AuthRequest, Seq[String]] = Get / "allTags" / long /> GetAllTags
```
travisbrown: Just to recap our conversation yesterday, it'd be more REST-y to use `"tag"` here, and if you're getting all tags you don't need a `long` parameter in the URL. I think it would be best to stick to the Swagger example's API for now, though, and it doesn't expose a resource for tags.

travisbrown:
One general comment: in order to get the code in the new `petstore` directory to compile, you'll need to add it as a subproject to the SBT build configuration in `build.sbt`. Since the configuration needs to be exactly the same as the `demo` one at this point, you can just [copy and past it](https://github.com/mimilei/finch/blob/petstore/build.sbt#L130) and edit the names:

```
lazy val petstore = project
  .settings(moduleName := "finch-petstore")
  .settings(allSettings)
  .settings(noPublish)
  .disablePlugins(JmhPlugin)
  .dependsOn(core, argonaut)
```

petstore/src/main/scala/io/finch/petstore/model.scala
```diff
@@ -0,0 +1,77 @@
...
+      category: Category,
+      name: String,
+      photoUrls: Seq[String],
+      tags: Seq[String],
+      status: String
+      )
+
+  object Pet{
+    // Provides an implementation of the EncodeJson type class from Argonaut
+    implicit def profileEncoding: EncodeJson[Pet] = jencode2L(
```
travisbrown: The number in the `jencode` methods should correspond to the number of fields in the case class (so in this case this should be `jencode6L`).

travisbrown:
Argonaut has three type classes associated with reading and writing JSON (this follows a fairly common pattern for type class-based serialization libraries):

* `EncodeJson[Pet]` describes how to take an instance of a `Pet` type (in this case a case class) and turn it into JSON.
* `DecodeJson[Pet]` describes how to take some JSON and (try to) turn it into a `Pet`.
* `CodecJson[Pet]` bundles these together into one thing.

If we've got these types:

```scala
case class Category(id: Long, name: String)
case class Tag(id: Long, name: String)

sealed trait Status
case object Available extends Status
case object HasAHome extends Status

case class Pet(
  id: Long,
  name: String,
  photoUrls: Seq[String],
  category: Category,
  tags: Seq[Tag],
  status: Status
)
```

Then we can define `EncodeJson` instances like this:

```
import _root_.argonaut._, Argonaut._

implicit val categoryEncode: EncodeJson[Category] =
  jencode2L((c: Category) => (c.id, c.name))("id", "name")

implicit val tagEncode: EncodeJson[Tag] =
  jencode2L((t: Tag) => (t.id, t.name))("id", "name")

implicit val statusEncode: EncodeJson[Status] =
  EncodeJson.jencode1[Status, String] {
    case Available => "available"
    case HasAHome => "unavailable"
  }

implicit val petEncode: EncodeJson[Pet] =
  jencode6L { (p: Pet) =>
    (p.id, p.name, p.photoUrls, p.category, p.tags, p.status)
  }("id", "name", "photoUrls", "category", "tags", "status")
```

And I've included `DecodeJson` instances in the [tutorial draft](https://github.com/travisbrown/finch/blob/topic/tutorial/docs/tutorial.md).

In some cases you only need to encode or decode certain types, so it makes sense only to provide `EncodeJson` or `DecodeJson` for your types. For this application you'll need both, though, so you want `CodecJson`, which combines them:

```scala
implicit val categoryCodec: CodecJson[Category] =
  casecodec2(Category.apply, Category.unapply)("id", "name")

implicit val tagCodec: CodecJson[Tag] =
  casecodec2(Tag.apply, Tag.unapply)("id", "name")

val statusEncode: EncodeJson[Status] =
  EncodeJson.jencode1[Status, String] {
    case Available => "available"
    case HasAHome => "unavailable"
  }
implicit val statusDecode: DecodeJson[Status] =
  DecodeJson { c =>
    c.as[String].flatMap[Status] {
      case "available" => DecodeResult.ok(Available)
      case "unavailable" => DecodeResult.ok(HasAHome)
      case other => DecodeResult.fail(s"Unknown status: $other", c.history)
    }
  }

implicit val statusCodec: CodecJson[Status] =
  CodecJson.derived(statusEncode, statusDecode)

implicit val petCodec: CodecJson[Pet] =
  casecodec6(Pet.apply, Pet.unapply)("id", "name", "photoUrls", "category", "tags", "status")
```

This all looks kind of confusing, but you only have to do it once for your data types and then you get nice stuff like `body.as[Pet]` everywhere.

petstore/src/main/scala/io/finch/petstore/package.scala
```diff
@@ -0,0 +1,76 @@
...
+import com.twitter.util.Future
+import io.finch._
+import io.finch.demo.model
+import io.finch.demo.model.User
+import io.finch.request.ToRequest
+import model._
+
+// A custom request type that wraps an `HttpRequest`.
+// We prefer composition over inheritance.
+case class AuthRequest(http: HttpRequest)
```
travisbrown: As discussed yesterday, we can scrap this type since we're going to start without authentication (since eventually it'll look different from this, anyway).

petstore/src/main/scala/io/finch/petstore/reader.scala
```diff
@@ -0,0 +1,36 @@
...
+   */
+//  val userReader: RequestReader[User] =
+//    param("username") should beLongerThan(4) map {
+//      name => User(Id(), username, "unknown", "unknown", "unknown", "unknown", "unknown")
+//    }
+
+  /*
+  RequestReader for Category
+   */
+  val petCategoryReader: RequestReader[Category] = {
```
travisbrown: You only need `RequestReader[A]` instances that correspond to `body` or `query` parameters in the [API](http://petstore.swagger.io/).

mimilei:
Moved working version to bestPetstoreEver.

Issue #328, opened 2015-06-17
Oddity with Encoding Raw Strings


rpless:
I've been fixing and expanding the [Finch implementation of the Tech Empower Benchmark](https://github.com/TechEmpower/FrameworkBenchmarks/tree/master/frameworks/Scala/finch).
For the [plain text test](http://example2frameworkbenchmarksdocs.readthedocs.org/en/latest/Project-Information/Framework-Tests/#plaintext), I have a router: 
```scala
  val plaintext: Router[String] = Get / "plaintext" /> "Hello, World!"
```
When I run Tech Empower's tests this yields an error saying that the response has 2 more bytes than expected. When I look at the response I see that there are 2 quotes in the body I don't think should there:
```
HTTP/1.1 200 OK
Content-Type: application/json;charset=utf-8
Date: Wed, 17 Jun 2015 18:09:55 GMT
Content-Length: 15
Server: Finch 0.7.0

"Hello, World!"
```
Are these quotes supposed to be in the response and if so how do I get rid of them? Or is this a bug?

vkostyukov:
Hm. Interesting. I believe these quotes happened after #283. We should probable write a tase case and try to fix it.

vkostyukov:
I hope this doesn't mean that we have the same issue with JSON encoding...

rpless:
It doesn't seem to. I've cleaned up the json test too and it passes their test with flying colors.

vkostyukov:
Just double-checked the demo project. The output JSON is not wrapped with quotes:

```
TTP/1.1 200 OK
Content-Type: application/json;charset=utf-8
Content-Length: 221

[{"id":0,"name":"Tom Paine","tickets":[]},{"id":1,"name":"Benjamin Franklin","tickets":[{"id":2,"label":"PHL -> CDG"},{"id":3,"label":"PHL -> LHR"}]},{"id":4,"name":"Betsy Ross","tickets":[{"id":5,"label":"PHL -> DCA"}]}]
```

vkostyukov:
Maybe @yanana can quickly spot the error?

rpless:
Yeah, so far it just seems to be strings. I've pushed my current work [here](https://github.com/rpless/FrameworkBenchmarks/blob/update-finch/frameworks/Scala/finch/src/main/scala/WebServer.scala). Might be a good starting put. I'll continue to look at it, but help other sets of eyes are appreciated.

travisbrown:
[...]

PR #329, opened 2015-06-17
Add a content type type member to response encoders

travisbrown:
This is a quick proof of concept to follow up on the conversation in #328 this afternoon. The issue is that it's currently easy to accidentally use an `EncodeResponse[String]` instance that's intended for encoding strings as JSON when you just want to encode a string as plain text.

This change adds a type member to `EncodeResponse` that tracks the content type as a type-level string (this allows us to avoid enumerating content types as singleton objects or whatever).

The behavior is still the same when you don't specify the content type:

```scala
scala> import io.finch.response._
import io.finch.response.EncodeResponse

scala> implicitly[EncodeResponse[String]].apply("foo")
res0: com.twitter.io.Buf = ByteBuffer(3)

scala> import io.finch.argonaut._
import io.finch.argonaut._

scala> implicitly[EncodeResponse[String]].apply("foo")
res1: com.twitter.io.Buf = ByteBuffer(5)
```

But when you do specify the content type you'll get the instance you want:

```scala
scala> implicitly[EncodeTextResponse[String]].apply("foo")
res2: com.twitter.io.Buf = ByteBuffer(3)

scala> implicitly[EncodeJsonResponse[String]].apply("foo")
res3: com.twitter.io.Buf = ByteBuffer(5)
```

Here `EncodeTextResponse` and `EncodeJsonResponse` are just type aliases:

```scala
type EncodeTextResponse[-A] = EncodeResponse.Aux[A, Witness.`"text/plain"`.T]
type EncodeJsonResponse[-A] = EncodeResponse.Aux[A, Witness.`"application/json"`.T]
```

(The `Witness` syntax is handy but kind of unpleasant looking.)

Note that the usage of `EncodeResponse.apply` is changed up a bit in order to help Scala's type inference be a little more useful. Otherwise there are no changes to tests or examples.

### DIFF
--- a/argonaut/src/main/scala/io/finch/argonaut/package.scala
+++ b/argonaut/src/main/scala/io/finch/argonaut/package.scala
@@ -27,7 +27,7 @@ package io.finch
 import _root_.argonaut.{EncodeJson, Json, Parse, DecodeJson}
 import io.finch.request.DecodeRequest
 import io.finch.request.RequestError
-import io.finch.response.EncodeResponse
+import io.finch.response.{ EncodeJsonResponse, EncodeResponse }
 import com.twitter.util.{Try, Throw, Return}
 
 package object argonaut {
@@ -50,6 +50,6 @@ package object argonaut {
    * @tparam A The type of data that the ''EncodeJson'' will encode use to create the json string
    * @return Create a Finch ''EncodeJson'' from an argonaut ''EncodeJson''
    */
-  implicit def encodeArgonaut[A](implicit encode: EncodeJson[A]): EncodeResponse[A] =
-    EncodeResponse.fromString[A]("application/json")(encode.encode(_).nospaces)
+  implicit def encodeArgonaut[A](implicit encode: EncodeJson[A]): EncodeJsonResponse[A] =
+    EncodeResponse("application/json").fromString[A](encode.encode(_).nospaces)
 }
--- a/core/src/main/scala/io/finch/response/EncodeResponse.scala
+++ b/core/src/main/scala/io/finch/response/EncodeResponse.scala
@@ -29,40 +29,61 @@ import com.twitter.io.Buf
 import com.twitter.io.Buf.Utf8
 import com.twitter.util.Future
 import io.finch._
+import shapeless.Witness
 
 /**
  * An abstraction that is responsible for encoding the response of type `A`.
  */
 trait EncodeResponse[-A] {
+  type C <: String
+
   def apply(rep: A): Buf
   def contentType: String
   def charset: Option[String] = Some("utf-8")
 }
 
 object EncodeResponse {
+  type Aux[A, C0] = EncodeResponse[A] { type C = C0 }
+
+  /**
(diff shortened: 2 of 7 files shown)


vkostyukov:
This looks cool!

I have just one question. Should we change `ResponseBuilder` API to support new `EncodeResponse` types?

travisbrown:
@vkostyukov Yeah, that makes sense, and should be pretty straightforward. I'm out for a bit but I'll update the PR when I get home.

vkostyukov:
I'm not sure I understand `Witness` correctly. Does it mean that we've enabled compile-time checking for the errors like `EncodeResponse("aplication/json")...` (single `p`)?

travisbrown:
Not really—the advantage is that it allows the compiler to select an implicit instance on the basis of the content string. I can ask for an instance for encoding strings as `text/plain` and I'll get that one instead of the `application/json` instance that may be in scope.

You could accomplish almost exactly the same thing by enumerating content types like this:

```scala
class ContentType(val contentType: String)
case object ApplicationJson extends ContentType("application/json")
case object PlainText extends ContentType("text/plain")
```

And then putting a `ContentType` type member in `EncodeResponse`. Using a type-level string is just more convenient, less boilerplate-y, allows users to add their own content types, etc.

vkostyukov:
Oh, I see. That makes sense. Thanks for the explanation @travisbrown!

travisbrown:
I started thinking a little more last night about how we want to make this new functionality available to users, and I'm not sure `RequestReader` is the best place to do it (or at least I'm not sure that it should be the only place).

Instead as a short-term solution we could provide a `toServiceAs("text/plain")` method that would be similar to `toService` but would only work if all of the values in the coproduct router have `EncodeResponse` instances for `"text/plain"`. Then users could create endpoints for specific content types and "or" them together with `:+:` (as in #324).

In the longer term I think there might be a nice way to handle content negotiation along these lines, but that would be a bigger API change. So that's probably something to consider before 1.0, but for now it would be nice to have something for 0.8 that handles e.g. the TechEmpower benchmark use case cleanly.

Any thoughts, @vkostyukov, @penland365, @rpless?

penland365:
I like this - the automatic preference to encode a plain string as JSON bit me last week ( easy fix though ). I do especially like the idea of a `toServiceAs("text/plain")` but that's also something that seems easy to those of that have been using Finch for awhile. Agreed that long term there needs to be more work around content negotiation, but that's a really hard topic to get right.

vkostyukov:
I had something in mind for a while (as part of Finch 1.0) that should give us an answer on how to do that. The fact that this issue's showed up gives me a hope that the version of Finch 1.0 I was imagining fits the real world pretty good so far.

It was a pretty busy week, so I didn't manage to put a roadmap doc together. I will try to finish it this weekend.

rpless:
I'm fine with `toServiceAs` as a short term solution. I'd prefer something closer to the current API, but I can't see a way to do that right now.

vkostyukov:
The ideas in this PR were merged in as part of #541. Thanks @travisbrown for the inspiration!

PR #331, opened 2015-06-21
Replace Smash with new Adjoin type class from Shapeless

### LINKED ISSUE #330: Move smash type classes for hlists and coproducts to Shapeless and remove ours
I've got [an issue open](https://github.com/milessabin/shapeless/issues/413) over at Shapeless and will try to put together an implementation of `Smash` and `CoproductSmash` that would fit there this weekend. Once (or if) these get released (maybe 2.3?) we'll be able to migrate and delete our implementations.

travisbrown:
This fixes #330 for `Smash` (I'll also rebase and update #324).

Shapeless's new `Adjoin` works on both `HList` and `Coproduct` (and so replaces both `Smash` and `CoproductSmash`). To preserve symmetry between the two, it returns a singleton `HList` in the case where a value is adjoined to `HNil`. As of #315 our `Smash` had a different behavior (it simply returns the value). This means that we still need an adapter, but most of the heavy lifting is done by Shapeless's `Adjoin` now.


vkostyukov:
@travisbrown this is great!

Maybe we shoud fully rely on `Adjoin` then? We can introduce a type-alias `Router1[A] = Router[A :: HNil]` to simplify the things? What do you think abou this?

```scala
val a: Router1[Int] = Get / int
val b: Router[User] = Get / int /> User
```

On the one hand I like the idea of completely reusing `Adjoin`. But on the ohter hand, I don't like the fact there is different type for `a` and `b`, while they look similar to me.

travisbrown:
Thanks, @vkostyukov.

I don't have strong opinions either way. I don't think the adapter is that terrible, and the new `Adjoin` will be even more of a benefit to us when we use it in #324 (which I'll try to get to today), where we won't need an adapter.

vkostyukov:
Yeah. That makes sense. Let's merge this then!

PR #332, opened 2015-06-23
Use -Yrepl-class-based to avoid getSimpleName errors

travisbrown:
`NotParsed` currently uses `.getSimpleName` in error strings, which causes "Malformed class name" problems for types defined in the REPL.

We may want to change this eventually, but in the meantime enabling `-Yrepl-class-based` for REPL sessions fixes the issue, and as far as I can tell doesn't break anything (I can run through the examples in [my new hands-on tutorial draft](https://github.com/travisbrown/finch/blob/topic/tutorial/docs/tutorial.md), for example).

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -41,7 +41,7 @@ val baseSettings = Seq(
       case _ => Seq.empty
     }
   ),
-  scalacOptions in (Compile, console) := compilerOptions,
+  scalacOptions in (Compile, console) := compilerOptions :+ "-Yrepl-class-based",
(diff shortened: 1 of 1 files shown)


travisbrown:
Ugh, it's that damn Wartremover exception again crashing the first job. Restarting now…

vkostyukov:
Great! Thanks!

travisbrown:
For the record, @retronym [pointed out on Twitter](https://twitter.com/retronym/status/613429241904959489) that this means that we can't define value classes in the REPL. I think that's much less of a problem than having JSON decoding errors crashing the REPL, but it's a bit of extra motivation to come up with a better solution for `NotParsed`.

PR #333, opened 2015-06-25
Update Scala to 2.11.7, Finagle to 6.26.0, etc.

travisbrown:
There are also new sbt-jmh versions but that's not a trivial upgrade so I didn't include it here.


vkostyukov:
Nice! Thanks @travisbrown!

Issue #327, opened 2015-06-17
Lazy Routing


vkostyukov:
The basic idea is to prevent evaluation of the `Router`s extracted value if we just want to check how many path's parts that router's matched. This would probably also involve renaming `apply` to `run` in the `Router`.

vkostyukov:
Fixed by #360.

PR #334, opened 2015-06-29
Compose Router and RequestReader

vkostyukov:
This is the initial attempt tp solve #232. While it requires some polish it's ready to be reviewed.

The change is pretty local, it doesn't really touch the user-facing API except for one (pretty major) thing: I had to give up on greedy routers, since otherwise it would require evaluating two futures. So, this PR also addresses #327.

With this PR it's possible to compile:

```scala
val putUser: Router[User] = 
  get("users" / int ? (param("name") :: param("age")) /> { (id, name, age) =>
    User(id, name, age)
  }
```

I would appreciate any ideas on the following topics:

1. Is there a better name for `?`?
2. How routers can be both lazy and greedy?
3. How can we move some logic from `Router` to `Router.Output` (with adding new functions like `flatMap`, etc)?


travisbrown:
I'm still hesitant to use `Future` to represent the success or failure of the match. It seems like what we really want is a `Path => Option[Path]` transition that returns the remainder in the case of a match, and some result `A` that gets evaluated if we both match and decide to use this route—i.e. something pretty similar to the original API, except that `Path => Option[(Path, A)]` isn't lazy enough about `A`.

vkostyukov:
Right. I was trying to avoid using `Future` here, but it seems to be impossible. `Future` doesn't really represents a successful/unsuccessful match. It represents an async value return from this router. I had to use it since `RequestReader` returns `Future`. The underlying type `Router.Output` represents a  successful/unsuccessful match.

My initial thoughts were to use something like `Input => Option[(Input, Future[A])]`, but I can't think of the implementation of the `flatMap` method in this case:

```scala
def flatMap[B](f: A => Router[B]): Router[B] = new Router[B] {
  def apply(input: Intput): Option[(Input, Future[B])] =
    self(input) match {
      case Some((i, fa)) => ???
      case _ => None
    }
}
```

vkostyukov:
@travisbrown Hm. I'm starting thinking about splitting `apply` onto `match` and `extract` as you suggested. Do you think we will be able to something like this:

```scala
trait Router[A] {
  def match(path: Path): Option[Path]
  def extract(req: HttpRequest): Future[A] // to be able to compose it with RequestReader
}
```

travisbrown:
@vkostyukov Yeah, but we may want to use information from the match when creating our `A`.

Do we actually need or want `Router` to be monadic? As a quick experiment I can everything to build and all tests to pass (except for the `flatMap` one) by writing `andThen` in terms of a new `ap(f: Router[A => B]): Router[B]` instead of `flatMap`. So monadic usage doesn't seem to be something we care enough about to use in the examples, at least.

vkostyukov:
Hm. That's interesting. I think we don't really want `Router` to be monadic, being applicative is sufficient for our use-cases. What signature did you use? `Input => Option[(Input, Future[A])]` one?

travisbrown:
No, I just used the current one. I'll try it with this branch either over lunch or this evening.

travisbrown:
This is borderline off-topic, but I've been vaguely wondering about using a free algebra for this kind of thing (not necessarily in the context of Finch). The idea is a little like the idea behind Stitch—you write your path matching logic using a syntax that could look identical to Finch's current syntax, but what's happening is that you're building up a little path-matching program that you then interpret in some context.

This could possibly simplify the choice of `Option` or `Future` here a bit, but it also has the nice side effect of supporting optimization—if I've got ten lines in my router that all begin with `Get / "foo" / "bar" / "baz" / <some other stuff>`, I'd end up with a path matching tree that doesn't fork until the `<other stuff>` (instead of forking at the root).

vkostyukov:
Oh, I see. I'm not sure we will be able to avoid introducing `Future` here since `RequestReader` returns one. So, the question is mostly about "where to introduce it?". I wish we could be able to do all the things: 1) compose `Router` with `RequestReader`; 2) do greedy routing; 3) do lazy routing.

Re free algebra. That sounds very interesting! I'm pretty sure we could think of something like this in the future. I'm totally on board with anything that doesn't change (much) the current version of user-facing API.

vkostyukov:
I have rebased this.

After thinking about this more. I'm kinda feeling Ok about returning `Future` from `Router`. That feels pretty natural to me given that `Router` is intended to replace `Service`. The only thing I don't like here in this PR is that we gave up on greedy routers. I will think about them more this weekend. Perhaps @travisbrown's idea with separated `match` and `extract` should help here. The only question is how to decouple those pretty dependent logics.

vkostyukov:
Rebased that again: some tests are failnig since we gave up on greedy routers.

vkostyukov:
Already merged in within #360.

Issue #337, opened 2015-06-30
Disable the Nothing wart


vkostyukov:
See some context [here](https://twitter.com/vkostyukov/status/615947659824599040).

PR #338, opened 2015-06-30
Add explicit type parameters to avoid inferring Nothing

travisbrown:
I agree that #337 is reasonable, but this is a quick and easy fix for these instances.


vkostyukov:
Thanks @travisbrown!

PR #340, opened 2015-07-03
Change Windows line terminators

travisbrown:
These two files are different from the rest of the codebase in that they use CRLF line terminators, which make automated changes (with e.g. `sed`) more noisy than they should be.

I've used Sublime's line-ending converter to make the switch. After this change ` find . -not -type d -exec file "{}" ";" | grep CRLF` doesn't list any source files.


vkostyukov:
Superb! Thanks @travisbrown!

PR #343, opened 2015-07-07
Add convenience traits for testing services

travisbrown:
This is an idea that came up in conversation with @mimilei today. `ServiceSuite` allows us to define a test class that implements a `createService` method and then get a `Service` fixture in each of our tests. We can use the same test definitions for tests that call the service directly (`ServiceSuite`) or from a client that calls a locally-served service (`ServiceIntegrationSuite`).

I've changed the demo to use the new convenience class, and added an integration test for the demo service.


vkostyukov:
Thanks @travisbrown!

PR #344, opened 2015-07-07
Fix a couple of old typos

travisbrown:
Minor follow-up to #342 to fix stuff noticed there but unrelated to it.


vkostyukov:
Great! Thanks @travisbrown!

PR #346, opened 2015-07-08
fix #345: Move Basic Auth into core

### LINKED ISSUE #345: finch-auth jar is empty
It looks like finch-auth_2.11-0.7.1.jar as published on maven central is empty. It only contains a MANIFEST.MF file.

Edit: it looks like the source is under `src/main/` instead of `src/main/scala/`

vkostyukov:
See #345 for the context.

This patch replaces the old-school Finagle filter `BaiscallyAuthrorize` with hipster `Router` combinator `basicAuth`.

Before:

```scala
val auth = new BaiscallyAuthrorize("user", "password")
Http.serve(":8081", auth.andThen(api.toService))
```

Now:

```scala
val auth = basicAuth("user", "password")
Http.serve(":8081", auth(api).toService)
```


travisbrown:
Looks fine to me.

Related: I think we need a convenient way to wrap a specific set of endpoints in a filter in the setting of a coproduct router—e.g. in order to be able to take some `Router[A :+: B :+: CNil]`, put it behind a login, and then compose it with an unfiltered `Router[C :+: D :+: CNil]` to get something we can serve.

vkostyukov:
@travisbrown good idea. If I understand you correctly. We need a way to convert `SimpleFilter` into something like `Router[A] => Router[A]`. Is that what you mean?

travisbrown:
@vkostyukov Right, exactly.

vkostyukov:
I think it's ready to go.

PR #354, opened 2015-07-20
Added scalastyle check for space after FOR and IF

mimilei:


mimilei:
I considered spaces around brackets but it seems problematic. Spaces aren’t used in select statements such as imports, so if I add that feature people will get lots of false alarms.

vkostyukov:
Sounds good. Let's start with `for` and `if`.

PR #360, opened 2015-07-29
Demonstration of lazy, greedy, non-monadic routers

travisbrown:
This is a follow-up and alternative to #334 that models routers more or less as `StateT[Option, Input, Future[A]]`. Failure to match is represented in the `Option`, while the laziness of the result is represented in the `Future` (which also provides convenient integration with request readers).

There are no real user-facing changes—all tests outside of `RouterSpec` pass as they are.

There are three important differences from the solution in #334:

1. This approach supports greediness.
2. My `Router` is no longer monadic (this isn't possible with `Future` so far inside). It is applicative, though, which seems to be all that's required in normal use cases.
3. A failed `Future` doesn't result in a failed match. This seems reasonable to me—if the user is asking for some kind of item that doesn't exist, we generally want to handle that specific error, not fall through to other possible matches. This is possibly a little counter-intuitive when it comes to integration with `RequestReader`, though, since e.g. a missing request parameter won't fail the match.

This PR is a demonstration and needs some doc clean-up before it's ready to merge.


vkostyukov:
This looks very cool! Thanks @travisbrown!

Re docs: don't worry about them for now: I will add docs right before the release. Let me know once you feel it's ready to get merged. I'm super excited about this PR!

travisbrown:
One other point: it's currently the case that if two routers `x` and `y` both match, then both of their futures will be evaluated in `x | y`. This isn't ideal, and breaks expectations about laziness. It's possible to fix this by splitting `apply` into separate matching and value methods, but I won't be able to get to that until this evening.

vkostyukov:
Good point. No worries. There is no rush on this at all.

yanana:
Super cool! :+1:

travisbrown:
It seemed like the most functional way to do this was to wrap all of the future values in `Function0`.

vkostyukov:
Agreed. Just out of curiosity, does it look like Cat's `Lazy` can help here?

Anyways. Let's merge this in. The approach seems to be very simple and pragmatic. I like it! Travis, are you ok with merging this?

travisbrown:
@vkostyukov Yes, `Lazy` would be perfect here, and this change as a whole would benefit a lot from abstractions from Cats or Scalaz. But it's not too bad as it is, so I'd say let's merge it and clean it up when we decide we're ready for a Cats dependency.

vkostyukov:
@travisbrown sounds good to me! Thanks a lot again!

PR #361, opened 2015-07-30
Update Shapeless and Finagle versions

travisbrown:
Very minor change but I thought we might as well go ahead and get these in.


vkostyukov:
:+1:

PR #365, opened 2015-07-30
Update sbt-unidoc version to support linking to Finagle docs

travisbrown:
The newer version of sbt-unidoc supports SBT's `autoAPIMappings`, which gives us links from the Finch API docs to the API docs of Finagle and Util. I've also fixed a few instances where Scaladoc couldn't find the class to link, and I've excluded the benchmarks and example projects.


vkostyukov:
:+1:

PR #366, opened 2015-08-01
Enable sbt-header; Regenerate headers

vkostyukov:
Played a bit with `sbt-header`. It's now possible to regenerate all the headers with `sbt header` command. This PR also includes the result of such regeneration.


vkostyukov:
Closed in favour #367.

PR #370, opened 2015-08-03
Introduce the Mapper type class

vkostyukov:
The type class `Mapper` is a variation of "magnet pattern" in Scala that allows us to workaround JVM's type-erasure. This PR introduces an `apply` method on `Router` that accepts a `Mapper` instaince. The compiler will substitute given `Mapper` with either `A => B` or `A => Future[B]`.

The motivation behind this change is to be able to write this:

Before:
```scala
val getUser: Router[User] = get("users" / string) /> { name: String =>
  User(name)
}

val patchUser: Router[User] = 
  patch("users" / string ? body.as[User]) />> { (name: Stirng, user: User) =>
     Future.value(user)
  }
```

After:
```scala
val getUser: Router[User] = get("users" / string) { name: String =>
  User(name)
}

val patchUser: Router[User] = 
  patch("users" / string ? body.as[User]) { (name: Stirng, user: User) =>
     Future.value(user)
  }
```

### DIFF
--- a/core/src/main/scala/io/finch/route/Mapper.scala
+++ b/core/src/main/scala/io/finch/route/Mapper.scala
@@ -1,0 +1,46 @@
+package io.finch.route
+
+import com.twitter.util.Future
+import shapeless.ops.function.FnToProduct
+
+/**
+ * A type class that allows the [[Router]] to be mapped to either `A => B` or `A => Future[B]`.
+ */
+trait Mapper[A] {
+  type Out
+
+  def apply(r: Router[A]): Router[Out]
+}
+
+trait LowPriorityMapperConversions {
+  type Aux[A, B] = Mapper[A] { type Out = B }
+
+  implicit def mapperFromFunction[A, B](f: A => B): Mapper.Aux[A, B] = new Mapper[A] {
+    type Out = B
+    def apply(r: Router[A]): Router[Out] = r.map(f)
+  }
+
+  implicit def mapperFromFutureFunction[A, B](f: A => Future[B]): Mapper.Aux[A, B] = new Mapper[A] {
+    type Out = B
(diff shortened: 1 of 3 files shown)


travisbrown:
I like the usage that this supports a lot. I'm always a little skeptical of implicit conversions, though, and in general instead of an implicit conversion from an `F` that can either be `A => B` or `A => Future[B]` to a type wrapping `Router[A] => Router[B]`, I think there are often reasons to prefer a type class that witnesses that a function `F` can be treated as a `A => Future[B]`, which can be used to convert a `Router[A]` to a `Router[B]`.

This is partly a knee-jerk reaction, but I think there's some legitimate reasoning behind it, and I'm pretty sure it would be possible to sketch out a very similar implicit conversion-less alternative that would support the same usage in a more composable way. I'll try to do that later this evening, but might not get to it until tomorrow night.

vkostyukov:
Thanks @travisbrown! I'm totally fine with not using the "magnet pattern" here. In fact, I wasn't really hapy while intruducing it: I just haven't seen another way of doing this.

Anyways, I feel like am getting the idea you're decribing and it sems pretty solid to me. So I would love to see where is it going. There is no rush on this. I just wanted to ship this API in 0.8.0 to be able to use in Finch workshop: I wrote `/>` and `/>>` like a couple of times and already got bored of them.

travisbrown:
For the record, I took a stab at this last night and it seems not straightforward to use an ordinary type class without implicit conversions because of type inference issues (the change would also have to happen after #371 because the overriding would need to be a little different).

I'd be happy to see this merged now, and to re-evaluate for 0.9.0. Even if the implementation needs to change a bit, I think we'll want to provide this syntax.

PR #372, opened 2015-08-04
Refactor tail extractors

vkostyukov:
This is a follow up PR for #353.

### DIFF
--- a/core/src/main/scala/io/finch/route/Extractor.scala
+++ b/core/src/main/scala/io/finch/route/Extractor.scala
@@ -5,7 +5,7 @@ import com.twitter.util.{Future, Try}
 /**
  * An universal extractor that extracts some value of type `A` if it's possible to fetch the value from the string.
  */
-case class Extractor[A](name: String, f: String => A) extends Router[A] { self =>
+case class Extractor[A](name: String, f: String => A) extends Router[A] {
   import Router._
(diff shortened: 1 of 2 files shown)


core/src/main/scala/io/finch/route/Extractor.scala
```diff
@@ -13,37 +13,64 @@ case class Extractor[A](name: String, f: String => A) extends Router[A] { self =
...
+      s <- input.path
+      a <- Try(f(s)).toOption
+    } yield a)))
 
-    override def toString = self.toString + ".*"
-  }
+  def apply(n: String): Router[Seq[A]] = copy[A](name = n)
 
-  override def toString: String = s":$name"
+  override def toString: String = s":$name*"
```
ilya-murzinov: Maybe ```s":${name}s"```?

vkostyukov: Hm. I was thinking about that too, but decided to go with `*` since a) we already have matcher `*` that somehow matches the tail b) visually it's easy to misread `ints` as `int`, while start (`*`) at the end clearly indicates that it's a tail extractor.

ilya-murzinov: Ok, that's reasonable.

ilya-murzinov:
:+1:

PR #374, opened 2015-08-04
Changed updatePetViaForm to updatePetNameStatus

mimilei:
To keep server-side and client-side separate. :)


vkostyukov:
@mimilei Cool!

PR #393, opened 2015-08-28
Finagle 6.28.0 (and other minor updates)

travisbrown:
Note that we can't update to SBT 0.13.9 until [this issue](https://github.com/scalastyle/scalastyle/issues/156) is fixed.


vkostyukov:
Thanks @travisbrown!

PR #394, opened 2015-08-31
Update circe to 0.2.0

travisbrown:
Will update after circe 0.2.0 is released, which will be shortly after Shapeless 2.3.0 is out. For now I'm just confirming that these updates don't require any changes to source.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -173,9 +173,9 @@ lazy val circe = project
   .settings(allSettings)
   .settings(
     libraryDependencies ++= Seq(
-      "io.circe" %% "circe-core" % "0.1.1",
-      "io.circe" %% "circe-jawn" % "0.1.1",
-      "io.circe" %% "circe-generic" % "0.1.1" % "test"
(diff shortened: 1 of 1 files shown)


vkostyukov:
@travisbrown, Any chances Circe 0.2.0 will be released in a couple of weeks so we can include this PR in 0.8.5?

travisbrown:
@vkostyukov The only thing I'm waiting for is Shapeless 2.3.0, and it seems likely that it'll be available in less than a couple of weeks, so unless you see a problem with moving from Shapeless 2.2.5 to 2.3.0 in a patch update, merging this before 0.8.5 should be possible.

vkostyukov:
@travisbrown mind rebeasing this? I can't wait to merge this in!

travisbrown:
@vkostyukov Done! I haven't run the tests yet locally, though.

travisbrown:
Okay, this is ready to go as far as I'm concerned. :smile:

vkostyukov:
Great! Thanks a lot @travisbrown!

Issue #375, opened 2015-08-05
Idea: Error handling


vkostyukov:
Instead of writing a Finagle filter we could provide a method `handle` on `Router` with the following signature:

```scala
def handle(pf: PartialFunction[Throwable, Response]): Router[A]
```

... and do something like this:

```scala
val getUser: Router[User] = post("users" ? body.as[User]) { u: User =>
  User.save(u)
} handle {
  case NotPresent(BodyItem) => ???
}
```

ImLiar:
Cool! We already use something like that, but for twitter `Service`, since we works with it at the end.

PR #401, opened 2015-09-23
[F-375] Error handler for endpoint

ImLiar:
This PR is about issue #375. But yet there is a question. 

Due to current `Endpoint` implementation it's type parameter `A` is a type for `Output[A]`. And when you're trying to return different from router applied function's result (look for specs i.e.), you'll get `Endpoint[Any]` at the end, since your handler might return `InternalServerError` just in case:

```scala
val r1: Endpoint[Any] = get(/) { 1/0 } handle {
      case e: ArithmeticException => InternalServerError("oh noes")
}
```

I think, there should be better solution. Any ideas?

### DIFF
--- a/core/src/main/scala/io/finch/Endpoint.scala
+++ b/core/src/main/scala/io/finch/Endpoint.scala
@@ -181,6 +181,24 @@ trait Endpoint[A] { self =>
    */
   def toService(implicit ts: ToService[A]): Service[Request, Response] = ts(this)
 
+  /**
+   * Handle exception occurred at endpoint with async PartialFunction
+   */
+  def rescue[B >: A](pf: PartialFunction[Throwable, Future[Output[B]]]): Endpoint[B] =  new Endpoint[B] {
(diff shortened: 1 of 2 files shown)


vkostyukov:
Thanks @ImLiar!

This is pretty close to what I had in mind per #375. Althoguh, there is a PR from @ilya-murzinov (https://github.com/finagle/finch/pull/395) that does similar trick but dropping down to the `Respose` type, thereby solving the issue with `Any` type.

I'm not sure yet which solution is preffered. I think we shoud do a combination of both PR at the end of the day. Let me think more about it and see if there anything we can do to avoid the `Any` type as well as not dropping down to the `Response`.

ImLiar:
Well, yea. At first I was thinking about `Response`, but in the end it looks kinda low-level.

vkostyukov:
Okay. I've finally got somthing in mind. Let me merge this in so I can use it as a start point. Thanks @ImLiar!

ImLiar:
@vkostyukov nice to hear it!

Issue #390, opened 2015-08-19
Testing Endpoints


vkostyukov:
As asked [on the Gitter channel](https://gitter.im/finagle/finch?at=55d32e8a4f6c9d3018316d5d), it would be nice to provide a meaning for testing Finch endpoints (routers).

The idea I'm suggesting relies on types we expose from endpoints (that's a very good excuse to have them) and allows the users to treat endpoints as a regular functions so they could be easily tested.

Here is the example.

```scala
case class User(name: String)

val getUser: Endpoint[User] = get("users" / long) { id: Long =>
  User(id.toString)
}

// test
getUser.lift(100) shouldBe User("100")
```

The `lift` (or whatever) method lifts the endpoint to a function `(A, B, ..., Z) => Out`, where `(A, B, ..., Z)` are values extracted fromt the request by endpoint and `Out` is a returned value.

Not sure if it's even possible, but I'd like to move Finch (make required chnages) towards the direction it will be possible.

vkostyukov:
As discussed on Gitter, we can think of two steps to solve this. The ultimate goal is to allow people test their endpoints separately w/o thinking much of HTTP.

1. Move [`OutputOps`](https://github.com/finagle/finch/blob/master/core/src/main/scala/io/finch/Output.scala#L170-L175) to finch-test and make it public
2. Provide a new, user-friendly API for building `Inputs`

We can discuss here how would the API for 2 look like. I can think of something like:

- `Input.request(HttpRequest)`
- `Input.get("/users/foo")`
- ...
- `Input.post("/users/foo")`

We can also think about providing `withBody`, `withParams` and `withHeaders` copy-constructors so it will be friendlier to use.

@BenFradet mentioned that he'd like to think (take care?) about this issue.

BenFradet:
Yup, I'll start working on this.

vkostyukov:
Ah right. Let's just make that API open in `finch-core`?

BenFradet:
fine with me

Issue #191, opened 2015-02-13
Reverse routers


vkostyukov:
It would be nice to finally support this. I don't know in which way we can do it now. But I decided to create an issue before I forget about this.

PR #403, opened 2015-09-25
Support tracking underlying functions for mapped endpoints

travisbrown:
This change introduces a new subtype of `Endpoint` that keeps track of the fact that the endpoint is the result of mapping a function over another endpoint. This supports a testing idiom similar to the proposal in #390:

``` scala
import io.finch._, io.finch.test._

case class User(name: String)

val getUser: Endpoint.Mapped[Long, User] = get("users" / long) { id: Long =>
  User(id.toString)
}

assert(getUser.lift(100L) == User("100"))
```

`lift` is an enrichment method provided by `io.finch.test` that uses an `underlying` method on `Endpoint.Mapped`. Because some mappings may be into combinations of `Future` and `Option`, the return type of `underlying` is `Future[Option[A]]`, which isn't terribly convenient for testing, so `lift` awaits and unwraps this result.

Note that this change doesn't provide the exact syntax proposed in #390, since the input type must be statically tracked.

One other note: it would conceivably be possible to put an input type member on `Endpoint` itself, and have it default to `Nothing` or `Request`, but I think the approach I'm using here is a little cleaner.

### DIFF
--- a/test/src/main/scala/io/finch/test/package.scala
+++ b/test/src/main/scala/io/finch/test/package.scala
@@ -1,0 +1,12 @@
+package io.finch
+
+import com.twitter.util.Await
+
+package object test {
+  /**
+   * Syntax for testing an [[Endpoint.Mapped]].
+   */
+  implicit class MappedEndpointOps[In, A](endpoint: Endpoint.Mapped[In, A]) {
+    def lift(in: In): A = Await.result(endpoint.underlying(in)).value
+  }
+}
--- a/core/src/main/scala/io/finch/Endpoint.scala
+++ b/core/src/main/scala/io/finch/Endpoint.scala
@@ -17,7 +17,7 @@ trait Endpoint[A] { self =>
   /**
    * Maps this endpoint to either `A => Output[B]` or `A => Output[Future[B]]`.
    */
-  def apply(mapper: Mapper[A]): Endpoint[mapper.Out] = mapper(self)
+  def apply(mapper: Mapper[A]): Endpoint.Mapped[A, mapper.Out] = mapper(self)
 
   /**
    * Extracts some value of type `A` from the given `input`.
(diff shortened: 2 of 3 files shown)


travisbrown:
I guess it's also worth noting that existing usage is unchanged, so `val getUser: Endpoint[User] = ...` in my example above would work just fine, since `Endpoint.Mapped[Long, User]` is a subtype of `Endpoint[User]`.

vkostyukov:
Thanks @travisbrown! This looks great!

Your last note reminded me the apporach [parboiled2 uses with `Rule` type](https://github.com/sirthias/parboiled2#rule-types-and-the-value-stack), which basically looks like `Rule[Input, Output]`.

I like the current implementation a lot: it's lightweight and simple. I'm just wondering that _every_ endpoint should be testable (mapped). I belive, users will end up writing `Endpoint.Mapped` every time, since it makes a lot of sense to be able to test endpoints.

With that said, maybe we should think about making this (mapped endpoint) a default setup? I can understand that it might be annoying typing so many types every time, but maybe it's worth it, given that it gives the user an easy to way to test and reuse things.

test/src/main/scala/io/finch/test/package.scala
```diff
@@ -0,0 +1,12 @@
+package io.finch
+
+import com.twitter.util.Await
+
+package object test {
+  /**
+   * Syntax for testing an [[Endpoint.Mapped]].
+   */
+  implicit class MappedEndpointOps[In, A](endpoint: Endpoint.Mapped[In, A]) {
+    def lift(in: In): A = Await.result(endpoint.underlying(in)).value
```
vkostyukov: Do you think we should do both `liftToValue(in: In): A` and `liftToOutput(in: Int): Output[A]` so users can assert on content type, headers, status code, etc?

ImLiar:
Let me bring some scepcis here.

I doubt that it's worth to write all those types in production code base for testing purpouses. It's always nice to keep your business and testing logic separetly.

When you would like to test underlying service function which well serve an endpoint - you write test for that service.
When you would like to test your routes (path parsing, order etc), you better to test all your routes to be sured, that there is no errors. And to keep those tests close to production mode you would like to use `Request` and feed your `Service`. 
I just cant find a sense in this "middle" testing. It's not a unit testing nor an integrational tests, but it still would require you to write your code with caution of ability to run your tests.

travisbrown:
Are there non-test use cases for accessing the underlying function? If not I'm inclined to agree with @ImLiar.

A kind of compromise would be to hold onto the underlying function in `core` (which has little to no runtime cost) and to provide tooling in `test` that allows you to say that you want to assert facts about a `Endpoint[A]` as a mapped input from some input type. If the endpoint in question isn't actually a mapped endpoint from the input type, you get a runtime error, but that's just fine in a test.

rpless:
@travisbrown Could this potentially be used to implement some kind of reverse routing like in #191?

travisbrown:
@rpless You could use something like this for reverse routing, but it'd need to keep track of a lot more information about how the router was built. An alternative would be to have an optional macro annotation that you'd use if you wanted reverse routing—this would have less of a runtime cost and not be in the way for people who don't care about the feature.

test/src/main/scala/io/finch/test/package.scala
arnihermann: I think that makes sense.

Issue #381, opened 2015-08-07
Respond 501 by default for any unhandled exception


vkostyukov:
When converting `Router` to `Service` we should rescue any unandled exception to 501 so clients will see an HTTP response, not the closed connection.

ilya-murzinov:
I'd like to work on this.

vkostyukov:
Finagle 6.31 will support that [out-of-the-box](https://github.com/twitter/finagle/commit/571bbe4a0168f582d725698ca8880321d465618d).

PR #409, opened 2015-10-12
Error handling

vkostyukov:
This PR is an attempt to solve #381 and #375.

It introduces two methods on the `Endpoint`: `rescue` and `handle`.

The `rescue` method has exact same meaning as `Future.rescue`: it _rescues_ from the exception occurred somewhere in the `Future` chain. Its type signature tells us we can rescue to a super type of the `Endpoint`.

```scala
val e: Endpoint[Int] = get("div" / int / int) { (a: Int, b: Int) =>
  Ok(a / b)
} rescue {
  case _: ArithmeticException => BadRequest(0)
}
```

`rescue` might be used to _react_ on the errors in endpoint: for example, errors returned by Finagle client to Postgres.

The `handle` (maybe we need a different name here? Like `mapErrors`) method is used to specify how to convert exceptions into HTTP response. It takes a partial function `Throwable => Response`, but due to an implicit conversion `Output[A] => Response` (which I'm not happy with), it's possible to pass `Output`s there.

```scala
val e: Endpoint[Int] = get("div" / int / int) { (a: Int, b: Int) =>
  Ok(a / b)
} handle {
  case _: ArithmeticException => BadRequest("Can not divide by zero!")
}
```

To summarize: `rescue` is used to _react_ on the exceptions in the stack; `handle` is used to convert them to HTTP responses.

This is not a perfect implementation: there are things I'm not happy with, siuch as implicit conversions. But, I think, this is a good start point.


vkostyukov:
I have something different in mind. Something that will allow us to avoid implicit conversions. I'm trying to implement it right now. Will update this PR once it's ready.

vkostyukov:
Okay. I stuck here and appriciate any hints how to implement `efmap` and `femap`. 

The basic idea is to treat `Output` as an ADT so users can safely do something like this:

```scala
  val deleteTodo: Endpoint[Todo] = delete("todos" / uuid) { id: UUID =>
    Todo.get(id) match {
      case Some(t) => Todo.delete(id); Ok(t)
      case None => NotFound("err" -> "todo_not_found", "id" -> id.toString)
    }
  }
```

We treat all the 100..200 responses as success (`Output[A]`) and the reset as failrue `Output[Nothing]`. Evetything looks great so far except for a fact I don't have a clue of how to implement `efmap` and `femap` functions (see [here](https://github.com/finagle/finch/blob/115b19b4781a70adc14db455f3cdd1c2dd69e5a5/core/src/main/scala/io/finch/Endpoint.scala#L59-L75)). Once it's done, I can think of getting rid of both `rescue` and `handle` for the sake of `def recover(pf: PartialFunction[Throwable, Output[Nothing]])` so we can do something like this:

```scala
val patchTodo: Endpoint[Todo] =
    patch("todos" / uuid ? patchedTodo) { (id: UUID, pt: Todo => Todo) =>
      Todo.get(id) match {
        case Some(currentTodo) =>
          val newTodo: Todo = pt(currentTodo)
          Todo.delete(id)
          Todo.save(newTodo)

          Ok(newTodo)
        case None => throw NotFoundError(id)
      }
    } rescue {
      case NotFound(id) => NotFound("id" -> id.toStirng)
    }
```

And that's it. No implicit conversions.

vkostyukov:
I think I'm most;y happy with this PR: will try to clean up it later so we can merge it early next week.

PR #410, opened 2015-10-12
Make finch-test a publishable module

arnihermann:
Move JsonCodecProviderProperties into its own module json-test to be able to publish finch-test.

As discussed in https://gitter.im/finagle/finch?at=561690a738d2a13c3c24caab.


arnihermann:
Please let me know if I've forgotten any steps to make `finch-test` publishable.

build.sbt
```diff
@@ -141,13 +150,13 @@ lazy val argonaut = project
   .settings(moduleName := "finch-argonaut")
   .settings(allSettings)
   .settings(libraryDependencies += "io.argonaut" %% "argonaut" % "6.1")
-  .dependsOn(core, test % "test")
+  .dependsOn(core, test % "test", `json-test` % "test")
```
travisbrown: I'm pretty sure the JSON implementation projects don't need a dependency on `test` now.

build.sbt
```diff
@@ -122,6 +122,15 @@ lazy val test = project
   .settings(allSettings)
   .settings(coverageExcludedPackages := "io\\.finch\\.test\\..*")
   .settings(
+    libraryDependencies ++= testDependencies
+  )
+  .dependsOn(core)
+
+lazy val `json-test` = project
```
travisbrown: I think I'd prefer to use `json-test` for the directory name, `finch-json-test` for the module name, and `jsonTest` as the project name. This is entirely a matter of taste, but I think it makes sense to be idiomatic in each domain.

You can specify the directory explicitly (when it doesn't match the project name) with `project.in(file("json-test"))`.

travisbrown:
Thanks for doing this, @arnihermann! Two additional comments:

1. You'll need to add both `test` and `jsonTest` to the `aggregate(...)` list on `core` in order for them to be built and published by default.
2. Can you also add `noPublish` to `jsonTest` so that we're not publishing artifacts for it?

arnihermann:
@travisbrown Thanks for the feedback. I think I've covered all of your points now.

arnihermann:
I'm not sure why travis fails, scalastyle passes on my machine.

travisbrown:
@arnihermann It looks like there is actually whitespace at the end of line 17 in `ServiceSuite`. This is my fault, and we hadn't seen it because `test` previously wasn't aggregated (just depended on by projects that were).

arnihermann:
My editor does something stupid so I didn't see it in `ServiceSuite`. It's fixed now.

travisbrown:
One last question from me:

I think it'd be a good idea to exclude `test-json` from the Scaladocs (using `unidocProjectFilter` in `build.sbt`). We're [currently publishing](http://finagle.github.io/finch/docs/#io.finch.test.json.package) them, but I think that's an oversight.

Once we hear from @vkostyukov about that (and about the PR more generally), I'm happy to see this merged. Thanks again, @arnihermann.

vkostyukov:
Thanks @arnihermann! This looks great! I'm happy to merge this once it's green (I've just restarted the build).

vkostyukov:
Thanks again @arnihermann!

arnihermann:
Happy to help!

PR #412, opened 2015-10-13
Add performance results

vkostyukov:
This PR adds a comparison table into a README. This doesn't look bad at all!


travisbrown:
Nice!

PR #413, opened 2015-10-13
Enable error handling in the Petstore example

kciesielski:
A proper `Filter` is implemented but not combined with the `service`.


petstore/src/main/scala/io/finch/petstore/service.scala
```diff
@@ -16,7 +16,7 @@ class PetstoreApp {
   db.addPet(Pet(None, "Cheshire Cat", Nil, Some(Category(None, "cat")), Some(Nil), Some(Available)))
   db.addPet(Pet(None, "Crookshanks", Nil, Some(Category(None, "cat")), Some(Nil), Some(Available)))
 
-  val service = endpoint.makeService(db)
+  val service = endpoint.handleExceptions andThen endpoint.makeService(db)
```
vkostyukov: I doubt it's necessary. The `handleExceptions` filter is [applied inside `makeService`](https://github.com/finagle/finch/blob/master/petstore/src/main/scala/io/finch/petstore/endpoint.scala#L55-L59) method.

kciesielski: Of course, I missed that.

PR #420, opened 2015-10-19
Adds Hotel Urbano as an adopter

andrestc:


vkostyukov:
Thanks @andrestc! I'm glad to hear you're using Finch!

Mind sharing a couple of details on how do you use it? What JSON backend, what version?

andrestc:
We have been using finch in an authentication/authorization API. Its simple service layer on top of our users db and it uses `finch-jackson` as the Json backend (mainly because we used Jackson before at our search api with finagle).

vkostyukov:
Great! Thanks for the details! 

While there is absolutelly nothing wrong with using Jackson with Finch, we're trying to encourage Finch users move to JSON libraries that support compile-time serialization/deserialization. That's why [Circe](https://github.com/travisbrown/circe) will be recomended as a defalt JSON backend in 0.9.0. But that doesn't mean we're going to stop supporting Jackson.

Anyways, feel free to ping me or @travisbrown on [Gitter](https://gitter.im/finagle/finch) if you ever dicide move to Circe/Argonaut and there is anything we can help with!

andrestc:
That looks great! We will definitely take a look at Circe.

PR #424, opened 2015-10-29
Fix missing output when / or ap is used in composition

vkostyukov:
This bug has been reported on [Gitter](https://gitter.im/finagle/finch?at=5631596490cea42a05840499) by @arnarthor.


travisbrown:
:+1: from me.

vkostyukov:
Thanks @travisbrown! We should really invest some time into refactoring our tests (especially `EndpointSpec`). I wish we used property-based testing more.

PR #426, opened 2015-10-30
fix some reference links to point at respective homepage

longcao:
This is a simple README fix for reference links since they didn't seem to be pointing to the right place.


vkostyukov:
Nice! Thanks @longcao!

PR #430, opened 2015-11-05
fix class name which leads user compile error in quickstart.md.

mike-neck:
* Change the document quickstart.md
  * In original document it says server can start via `Http#serve`
  * But `Http` object or class in `com.twitter.finagle.httpx` does not have `serve` method.
  * After grepping codes, I found it should be `com.twitter.finagle.Httpx`.

### DIFF
--- a/docs/quickstart.md
+++ b/docs/quickstart.md
@@ -35,10 +35,10 @@ val api: Router[String] =
   }
 ```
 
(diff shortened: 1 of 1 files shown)


vkostyukov:
Thanks @mike-neck! Nice catch!

mike-neck:
Thanks! :+1:

PR #435, opened 2015-11-11
Fix for typo in Failure

ImLiar:
https://gitter.im/finagle/finch?at=564301ec94df6212290da0db

in other case it makes no sense?


vkostyukov:
My bad. I'm probably going to release 0.9.1 with this fix. Thanks @ImLiar!

Issue #422, opened 2015-10-23
Make Mapper accept both Scala Futures and Twitter Futures


vkostyukov:
We can make a conversion (powered by Bijection) happen automatically in `Mapper` so users won't be worring about that.

jparkie:
Can I solve this issue? :)

vkostyukov:
Hey @jparkie! You're absolutelly welcome to work on this. Although, I can't promise we will merged this in before 1.0 (the current milestone is 0.8.5/0.9.0).

jparkie:
That is fine. :) By Bijection, do you mean Twitter's Bijection library?

vkostyukov:
Yep. [This](https://github.com/twitter/bijection/blob/develop/bijection-util/src/main/scala/com/twitter/bijection/twitter_util/UtilBijections.scala) gives us bijections to convert Scala Future to Twitter Future and vise versa.

travisbrown:
From what I remember of conversations with adopters who are using both futures heavily (e.g. SoundCloud, Tapad, etc.), it sounded like the consensus was that it's least painful to have clearly defined boundaries between the two—I'm not entirely sure allowing users to mix and match at the level proposed here is desirable.

It's also a bit of a shame to bring on a ~2 MB dependency for the sake of a single set of converters.

vkostyukov:
Thanks @travisbrown! That makes a lot of sense to me. Do you think it'd be ok to provide it as a part of separate package?

travisbrown:
@vkostyukov I don't see any reason not to. If it turns out that it actually does work well for people, it could always be moved into core.

jparkie:
I agree that introducing Bijection just for converting Scala Futures to Twitter Futures is cumbersome with the 2MB size. However, I find it useful to use Scala Futures, so I think porting the specific Bijection for Futures as an implicit import for may be nice for the core. Unless, there are very distinct uses for interoperability to require a new Finch package.

jparkie:
I actually have been quite busy, and I don't I can dive into Finch any time soon, so if anyone else wants to solve this issue, please do! :)

Saheb:
Can I grab this one?

vkostyukov:
[...]

PR #440, opened 2015-11-12
Generic Failures

vkostyukov:
Some context on [Gitter](https://gitter.im/finagle/finch?at=5644c477039e233306192220).

This PR alllows to constract generic failrues with a message of abritrary type. To keep the API unchanged, I haven't changed functions `BadRequest`, `InternalServerError`, etc so they still accept a vararg string tuples, but return `Output.Faulure[Map[String, String]]` instead of `Output.Failure`.

The following is possible now:

```scala
val divOrFail: Endpoint[Int] = post("div" / a / b) { (a: Int, b: Int) =>
  if (b == 0) 
    BadRequest("err" -> "Can not divide by zero.") // Failure[Map[String, String]]
  else if (a == 0) 
    Output.Failure("useless division", Status.InternalServerError) // Failure[String]
  else 
    Ok(a / b) // Payload[Int]
}
```

With one iteation (deprecation cycle), we'd be able to change predefined outputs to accept generic messages:

```scala
def BadRequest[A: EncodeResponse](message: A): Output.Failure[A] = 
  Output.Failure(message, Status.BadRequest)
```


mandubian:
I like it, it allows to do what I wanted (storing a body of any type in a Failure).
I'm not sure to know why you need to keep current API with `Map[String, String]` except not changing current API but for the rest, it seems quite good.
If you release 0.9.2(-SNAPSHOT) with this correction, I can test it in our live code. Thanks

akiradeveloper:
I like this too because S3 should respond error with XML on its body http://docs.aws.amazon.com/AmazonS3/latest/API/ErrorResponses.html

vkostyukov:
I decided to use a slightlly different approach, which makes way more sense to me. The idea behind this based on a fact that we prepresent domain errors as a hierarchy around `Exception`. That said, we'd need to tell Finch how to encode both successful results (payloads, `A: EncodeResponse`) and failures (exceptions) so it'd be able to turn any result into an HTTP response.

```scala
sealed trait Output[A]
final case class Payload[A](a: A) extends Output[A]
final case class Failure(cause: Exception) extends Output[Nothing]
```

Previously, `Endpoint.hadle` was responsible for both a) encoding exceptions and mapping them to HTTP status code. New approach is about reusing `EncodeResponse` abstraction for encoding exceptions as well and using `handle` only for mapping exceptions to status codes. By default, all the unmapped expections are converted into a very basic 500 response.

Some example usages:

```scala
scala> val e = get(int / int) { (a: Int, b: Int) => Ok(a / b) } handle { case e: ArithmeticException => BadRequest(e) }
e: io.finch.Endpoint[Int] = GET /:int/:int

scala> val rep1 = Await.result(e.toService.apply(Request("/10/0")))
rep1: com.twitter.finagle.http.Response = Response("HTTP/1.1 Status(400)")

scala> rep1.contentString
res0: String = / by zero

scala> implicit val ee: Encoder[Exception] = Encoder.instance { case a: ArithmeticException => Json.obj("err" -> Json.string("can not divide by 0")) }
ee: io.circe.Encoder[Exception] = io.circe.Encoder$$anon$2@18f21276

scala> val rep2 = Await.result(e.toService.apply(Request("/10/0")))
rep2: com.twitter.finagle.http.Response = Response("HTTP/1.1 Status(400)")

scala> rep2.contentString
res1: String = {"err":"can not divide by 0"}
```

The default isntance of `EncodeResponse[Exception]` is defined as

```scala
implicit val encodeException: EncodeResponse[Exception] =
    EncodeResponse.fromString("text/plain")(e => Option(e.getMessage).getOrElse(""))
```

That's why `rep1.contentString` gives `"/ by zero"`. Althogh, we could override this with Circe and tell Finch how to encode exceptions.

vkostyukov:
I'm pretty happy with this approach except for it breaks the current API for failures. I'm going to open a new PR that provides a a migration path through the deprecation cycle.

mandubian:
Let's publish that asap & I'll update my code with it to see how it behaves in real environment!

vkostyukov:
@mandubian It's already in MC! You can use [this example](https://github.com/vkostyukov/finch-101/blob/master/src/main/scala/i/f/workshop/finch/Todo.scala) to see how to a) tell Finch how to encode your exceptions (i.e., via Circe's `Encoder[Exception]`) and b) map excetions from an endpoint to a particular HTTP status code. Please note that if you don't do that explicilty, Finch will covert any unhandled exception into an _empty_ (it won't encode the exception into JSON or whatever) very basic 500 response.

#422 also made it into the SNAPSHOT, which contains a couople of breaking API changes: mostly changed imports: everything from finch-core is now available by a single import `io.finch._`.

Anyways, feel free to ping me on a gitter channel in case of any questions. And thanks for testing it!

mandubian:
Nice! Thanks & #422 fits my needs too (I depend much more on Scala Future API)
I'll tell you if I find anything wrong...

PR #441, opened 2015-11-15
Introduce monadic composition for Endpoint

akiradeveloper:
To pass inner endpoint a context from outer endpoint, Endpoint should be monadic. Endpoint being monadic benefits applications that pass some value (e.g. unique ID to the request) generated at the entry point to the exception handler of the inner endpoint.

### DIFF
--- a/core/src/main/scala/io/finch/Endpoint.scala
+++ b/core/src/main/scala/io/finch/Endpoint.scala
@@ -2,7 +2,7 @@ package io.finch
 
 import com.twitter.finagle.Service
 import com.twitter.finagle.http.{Cookie, Request, Response}
-import com.twitter.util.Future
+import com.twitter.util.{Await, Future}
 import shapeless._
 import shapeless.ops.adjoin.Adjoin
 
@@ -127,6 +127,13 @@ trait Endpoint[A] { self =>
       }
 
     override def toString = self.toString
+  }
+
+  def flatMap[B](fn: A => Endpoint[B]): Endpoint[B] = new Endpoint[B] {
+    override def apply(input: Input): Option[(Input, () => Future[Output[B]])] = for {
+      (r1, ao) <- self(input)
+      (r2, bo) <- fn(Await.result(ao()).value)(r1)
+    } yield (r2, bo)
   }
 
   /**
--- a/core/src/main/scala/io/finch/Mapper.scala
+++ b/core/src/main/scala/io/finch/Mapper.scala
@@ -32,6 +32,11 @@ trait LowPriorityMapperConversions {
       type Out = B
       def apply(r: Endpoint[A]): Endpoint[Out] = r.femap(f)
     }
+
+  implicit def mapperFromEndpointFunction[A, B](f: A => Endpoint[B]): Mapper.Aux[A, B] = new Mapper[A] {
+    type Out = B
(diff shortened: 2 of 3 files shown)


core/src/main/scala/io/finch/Endpoint.scala
```diff
@@ -129,6 +129,13 @@ trait Endpoint[A] { self =>
     override def toString = self.toString
   }
 
+  def flatMap[B](fn: A => Endpoint[B]): Endpoint[B] = new Endpoint[B] {
+    override def apply(input: Input): Option[(Input, () => Future[Output[B]])] = for {
+      (r1, ao) <- self(input)
+      (r2, bo) <- fn(Await.result(ao()).value)(r1)
```
vkostyukov: I'd be happy to accept this as long as there in `Await.result` in the code (except for tests). If you feel adventurous, you might think about changing the endpoint output type to something that'd allow you to write flatMap with awaits. Perpahs, #453 might help there in some way.

akiradeveloper: > as long as there is no Await.result in the code

Yeah.. I tried but it was too adventurous for me so I compromise with await, that blocks of course.

> changing the endpoint output type to something that'd allow you to write flatMap with awaits

using awaits but not block? such a magic exists?

vkostyukov: Sorry, should be "with **no** awaits". So no magic :)

akiradeveloper: OK. I will try again.

akiradeveloper: Just a few thoughts on this by playing with types

If we can write Eval Future Output as M then
Endpoint A = Input -> Option (Input, M A)
And now we are given
f :: A -> Endpoint B = A -> (Input -> (Input, M B))

What we implement is
flatMap :: Endpoint A -> f -> Endpoint B

If we can make this function
g :: (Input, M A) -> Option (Input, M B)
we can make flatMap because

flatMap =
self(Input).map(g) :: Option Option (Input, M B)
-> Option (Input, M B) // join

(Input, M A)
-> (Input, M (Input -> Option (Input, M B))) // map
-> M (Option (Input, M B)) // M (Input -> Option(Input, M B)) `ap` (pure Input)
-> // stuck. how can I eliminate M?

I hope someone smaller than me helps me

akiradeveloper:
I am not going to work on this anymore. So you can close this issue or keep this open which ever you like to.

vkostyukov:
I think we'd eventually end up having something like this for `Endpoint`, but w/o `Await`s. That will likely require some rethinking of the output type though.

akiradeveloper:
If you think so, why not merge this PR.
If the implementation is still immature users can write their codes with monadic endpoint that avoids rewriting a lot after introducing no-awaits implementation.
Backward compatibility is important and something is better than nothing.

Unlike modifications in error handling data types this is about control flow, the most intrinsic part of coding. So the users will end up with drastically rewriting their applications.

vkostyukov:
Closing this for now. @akiradeveloper thanks for working through this. Feel free to open a ticket if you still think it makes sense to have `Endpoint.flatMap`.

PR #451, opened 2015-11-17
Add XML encoder

akiradeveloper:
My S3 implementation strongly need this because S3's response is often XML.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -37,6 +37,7 @@ val baseSettings = Seq(
     "com.chuusai" %% "shapeless" % shapelessVersion,
     "com.twitter" %% "finagle-http" % finagleVersion,
     "org.scala-lang" % "scala-reflect" % scalaVersion.value,
+    "org.scala-lang.modules" %% "scala-xml" % "1.0.5",
     compilerPlugin("org.scalamacros" % "paradise" % "2.0.1" cross CrossVersion.full)
   ) ++ testDependencies.map(_ % "test"),
   resolvers ++= Seq(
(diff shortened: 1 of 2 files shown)


build.sbt
```diff
@@ -37,6 +37,7 @@ val baseSettings = Seq(
     "com.chuusai" %% "shapeless" % shapelessVersion,
     "com.twitter" %% "finagle-http" % finagleVersion,
     "org.scala-lang" % "scala-reflect" % scalaVersion.value,
+    "org.scala-lang.modules" %% "scala-xml" % "1.0.5",
```
vkostyukov: I'd vote for extracting it into a standalone sub-project and using something like [AntiXML](http://anti-xml.org/) instead.

akiradeveloper:
Like other for-JSON sub-projects?
Using Anti-XML also sounds a good way to go (sorry I didn't know the library)

So I will close this PR and send you a new PR later.

vkostyukov:
Right. Defining this in a sub project `finch-antixml` sounds pretty reasonable to me.

vkostyukov:
After looking at the AntiXML github page, I'm no longer sure it's a good library to use: the most recent commit there was made like four years ago. I'm not even sure it compiles with 2.10. Perhaps there are better XLM libraries around, but it's definetelly not scala-xml.

akiradeveloper:
Sorry ,what bothers you to choose scala-xml? It's pretty standard and still maintained.

akiradeveloper:
like for-JSON, there can be multiple sub-projects for XML encoder. How do you think?

PR #452, opened 2015-11-17
Add sbt-idea plugin

akiradeveloper:
To import to IntelliJ this plugin helps a lot.


fsvehla:
I had no troubles with the stock IDEA SBT support, what did not work for you?

akiradeveloper:
What's the stock IDEA SBT support?

fsvehla:
IDEA can import SBT projects natively

akiradeveloper:
with this plugin I can import finch to my IntelliJ without any problems but if I try to do without this plugin, import ends in failure. I want this plugin for stable import and enjoy participating finch.

rpless:
@akiradeveloper Can you outline what issues you are having with Intellij + Finch (errors, failures, etc.) and how this plugin resolves them?

akiradeveloper:
Wow... I tried "import" again and it succeeds now.

It was that nothing was imported so I couldn't see any subdirectories in the project explorer.

I guess some metadata has broken when I move the whole source tree (~) to other directory (~/share).

What I did was 

1. finch was at ~ and it was imported
2. move whole directory to ~/share
3. tried and failed
4. do gen-idea (but partially)
5. turned into success

I think gen-idea worked as a cleaner. So I still think adding sbt-idea is beneficial for this project.

akiradeveloper:
another problem is somehow it's named "root" in the IntelliJ's project selection. It should be "finch" as the directory name is so.

IntelliJ's embedded import seems to me fragile. So I do prefer sbt-idea so I can make things done before actually feed it to IntelliJ.

vkostyukov:
@akiradeveloper, You can put this pluing into `~/.sbt/0.13/plugins/build.sbt` on your compulter and it will be availalble for any sbt project you work with. I don't see a reason to have this pluging included in Finch (in fact it used to be there, but we've removed it a while ago by similar reasons).

akiradeveloper:
OK. I am convinced so I will this PR

Thank you for the comments guys

PR #453, opened 2015-11-17
Use Cats' Eval

vkostyukov:
This is just a first step towards Cats in Finch! But I think it's a good start!

This PR uses Cats' `Eval[A]` instead of `() => A`. The end-user API wasn't affected. I'm going to check the benchmarks later tonight and would appriciate any thoughts on this.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -11,6 +11,7 @@ lazy val buildSettings = Seq(
 lazy val finagleVersion = "6.30.0"
 lazy val circeVersion = "0.3.0-SNAPSHOT"
 lazy val shapelessVersion = "2.3.0-SNAPSHOT"
+lazy val catsVersion = "0.4.0-SNAPSHOT"
 
 lazy val compilerOptions = Seq(
   "-deprecation",
@@ -35,6 +36,7 @@ val testDependencies = Seq(
 val baseSettings = Seq(
   libraryDependencies ++= Seq(
     "com.chuusai" %% "shapeless" % shapelessVersion,
+    "org.spire-math" %% "cats-core" % catsVersion,
     "com.twitter" %% "finagle-http" % finagleVersion,
     "org.scala-lang" % "scala-reflect" % scalaVersion.value,
     compilerPlugin("org.scalamacros" % "paradise" % "2.0.1" cross CrossVersion.full)
--- a/core/src/main/scala/io/finch/Endpoint.scala
+++ b/core/src/main/scala/io/finch/Endpoint.scala
@@ -1,5 +1,6 @@
 package io.finch
 
+import cats.Eval
 import com.twitter.finagle.Service
 import com.twitter.finagle.http.{Cookie, Request, Response}
 import com.twitter.util.Future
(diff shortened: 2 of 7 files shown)


fsvehla:
Was the motivation to never evaluate lazy values by mistake?

vkostyukov:
@fsvehla what do you mean? Nothing actually has changed by this PR except for with `Eval`, we could do some nice things as `Eval.now` (which by defenition should be cheaper) for the output we're going to drop in any way.

fsvehla:
Ah I see, I’ve wondered about the utility vs. using by-name parameters, and this can to be more robust against evaluating the lazy value by mistake.

mandubian:
:+1:

travisbrown:
Nice—I'm looking forward to seeing what else we can do with this.

It's worth noting that `Eval[Future[x]]` here is a kind of approximation of `scalaz.concurrent.Task`—we need to be able to define a future but not have it begin execution immediately. This is a pretty generally useful thing to have, and it may be worth breaking out into its own thing—e.g. something like this (which is just a sketch):

```scala
package io.catbird.util

import cats.{ Applicative, Eval, Monad }
import com.twitter.util.Future

class Task[A](val value: Eval[Future[A]]) extends AnyVal {
  def toFuture: Future[A] = value.value
}

object Task {
  private[this] type EF[A] = Eval[Future[A]]
  private[this] val app: Applicative[EF] =
    Applicative[Eval].compose(Applicative[Future])

  def apply[A](a: => A): Task[A] = new Task(Eval.later(Future(a)))
  def fromFuture[A](f: => Future[A]): Task[A] = new Task(Eval.later(f))

  implicit val taskInstance: Monad[Task] = new Monad[Task] {
    def pure[A](x: A): Task[A] = new Task(app.pure(x))
    def flatMap[A, B](fa: Task[A])(f: A => Task[B]): Task[B] = new Task(
      fa.value.map(_.flatMap(a => f(a).value.value))
    )
  }
}
```

And then you can do this:

```scala
import com.twitter.util.{ Future, FuturePool }, io.catbird.util._

val task = Task.fromFuture(
  Future.collect(
    (0 to 10).map(i => FuturePool.unboundedPool { Thread.sleep(10000); println(i) })
  )
)
```

And it won't start running until you call `runFuture`, but then the futures will run asynchronously:

```scala
scala> task.runFuture
res2: com.twitter.util.Future[Seq[Unit]] = Promise@1493334289(state=Interruptible(List(),<function1>))

scala> 1
7
5
8
6
0
9
4
3
2
10
```

vkostyukov:
@travisbrown This is cool! I was actually thinking about making a type alias for `Option[(Input, Eval[Future[Output[A]]])]` beacuse it's quite annoying to type (and even hard to understand) it.

mandubian:
I've already plugged Finch on my own lazy-effect library called Précepte and I think your evolution might help me too...

vkostyukov:
@mandubian great to hear that! I think `Eval` should work fine for us and (perhaps) make things a bit faster while allocating less.

PR #456, opened 2015-11-18
Fix 406: Allow to override pretty printer for Circe and Argonaut

vkostyukov:
This PR allows users to override the default pretty printer for Circe and Argonaut.

Circe:

```scala
import io.finch.circe._
import io.finch.circe.dropNullKeys._
```

Argonaut:

```scala
import io.finch.argonaut._
import io.finch.argonaut.preserveOrder._
```

### DIFF
--- a/argonaut/src/main/scala/io/finch/argonaut/Decoders.scala
+++ b/argonaut/src/main/scala/io/finch/argonaut/Decoders.scala
@@ -1,0 +1,23 @@
+package io.finch.argonaut
+
+import scala.util.{Failure, Success}
+
+import argonaut.{CursorHistory, DecodeJson, Json}
+import com.twitter.util.{Return, Throw, Try}
+import io.finch.{DecodeRequest, Error}
+import jawn.Parser
+import jawn.support.argonaut.Parser._
+
+trait Decoders {
+
+  /**
+   * Maps Argonaut's [[DecodeJson]] to Finch's [[DecodeRequest]].
+   */
+  implicit def decodeArgonaut[A](implicit d: DecodeJson[A]): DecodeRequest[A] = DecodeRequest.instance[A] { s =>
+    val err: (String, CursorHistory) => Try[A] = { (str, hist) => Throw[A](Error(str)) }
+    Parser.parseFromString[Json](s)(facade) match {
+      case Success(value) => d.decodeJson(value).fold[Try[A]](err, Return(_))
+      case Failure(error) => Throw[A](Error(error.getMessage))
+    }
+  }
+}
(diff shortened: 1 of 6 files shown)


travisbrown:
I'm typically skeptical about using implicit values for configuration, but lots of people do it and I don't have a proposal for a nicer way to accomplish this, so I'd be happy to see it merged.

vkostyukov:
That's exactly how about I fee about that. I also don't like implicit configuration vaules, but didn't come up with something better. Although, I looks like there is only one boolean argument in Circe's printer that users might want to change and two agruments in Argonaut. Given that it's we actually implemented all the reasonable combinations of those values (`dropNullKeys`, `preserverOrder`), there is no need for users to do anything else with printers. I think we can drop the implicit value for `Printer` then.

vkostyukov:
The current approach is to use one of the supported configurations:
* For Circe: `io.finch.circe._`, `io.finch.circe.dropNullKeys._`
* For Argonaut: `io.finch.argonaut._`, `io.finch.argonaut.dropNullKeys_`, `io.finch.argonaut.preserveOrder._`, `io.finch.argonaut.preserveOrderAndDropNullKeys._`

I think I like this approach more and it gives us enoght flexibility to support all the reasonable use cases.

amartinsn:
This feature is quite handy. Have you also considered allow users to also specify the format convention they want to use. Some people user camelCase to name json attributes while others prefer snake_case. That would be awesome specifically for people migrating from Rails like APIs, which uses snake case to java based APIs. Which was my case. To do so, I had to use a Map so I could name attributes whatever I wanted.

Well done! Thanks!

travisbrown:
@amartinsn My feeling is that decisions involving the semantically significant content of the JSON (as opposed to the indentation, field order, preservation of nulls, etc.) should be handled at the codec layer, not here. I'm working on some ideas about ways to make this more convenient in circe.

vkostyukov:
Thanks @amartinsn! Good question. I'd say that the way we represent keys doesn't depend on printer we use, but on `Encoder` and `Decoder` (if we talk about Circe or Argonaut). So what you can do is to provide your own encoders/decoder by mapping the derived ones to snake_case names.

@travisbrown do you think it makes sense for Circe to do that out of the box:

```scala
import io.circe.generic.auto._ // default names
import io.circe.generic.auto.snakeCase._ // or simiar
```

travisbrown:
@vkostyukov I'm thinking more like `deriveEncoder[Foo].snakeCase` (for semi-automatic derivation) or `Encoder[Foo @@ SnakeCase].untagged` more generally, so that configuration is available per type and not determined by what implicits are in scope where.

vkostyukov:
@travisbrown Right. That's much better! I'm really looking forward to it!

amartinsn:
`Encoder[Foo @@ SnakeCase].untagged` looks great @travisbrown

vkostyukov:
Thank you all! I'm merging this.

PR #460, opened 2015-11-19
Fix #459: DecodeRequest from Generic

### LINKED ISSUE #459: Fix RequestReader[String].as[CaseClass]
Reported [on Gitter](https://gitter.im/finagle/finch?at=564d0a59167ee3a91d43d53a):

```scala
scala> case class Foo(s: String)
defined class Foo

scala> param("a").as[Int]
res0: io.finch.RequestReader[Int] = io.finch.RequestReader$$anon$3@21338f41

scala> param("a").as[Foo]
<console>:28: error: could not find implicit value for parameter decoder: io.finch.DecodeRequest[Foo]
       param("a").as[Foo]
```

vkostyukov:
This is very silly solution of the weird implicit lookup issue we've recently found.

```scala
scala> case class Foo(s: String)
defined class Foo

scala> param("a").as[Int]
res0: io.finch.RequestReader[Int] = io.finch.RequestReader$$anon$3@39694516

scala> param("a").as[Foo]
res1: io.finch.RequestReader[Foo] = io.finch.RequestReader$$anon$3@73cea421

scala> case class Bar(i: Int)
defined class Bar

scala> param("a").as[Int].as[Bar]
res3: io.finch.RequestReader[Bar] = io.finch.RequestReader$$anon$2@27d3ac40
```

### DIFF
--- a/core/src/main/scala/io/finch/DecodeRequest.scala
+++ b/core/src/main/scala/io/finch/DecodeRequest.scala
@@ -1,8 +1,10 @@
 package io.finch
 
-import com.twitter.util.{Throw, Try}
 import java.util.UUID
 import scala.reflect.ClassTag
+
+import com.twitter.util.{Return, Throw, Try}
+import shapeless.{::, Generic, HNil}
 
 /**
  * An abstraction that is responsible for decoding the request of type `A`.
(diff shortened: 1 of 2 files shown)


vkostyukov:
I think I'm gonna merge this for now, until we have a better solution.

PR #476, opened 2015-12-03
Update scalastyle sbt plugin

ImLiar:
0.8.0-SNAPSHOT doesn't resolve anymore since 0.8.0 was released. My Idea just went mad :(


travisbrown:
:+1: from me!

vkostyukov:
Thanks @ImLiar! Nice catch!

PR #482, opened 2015-12-06
Responses doc doesn't exist, make Authentication next, as per index.md

edwelker:


vkostyukov:
Nice! Thanks @edwelker!

PR #487, opened 2015-12-10
Fix #484: Catch exceptions thrown during multipart parsing

### LINKED ISSUE #484: Endpoint using `fileUpload` responds with a 500 status
This was discovered when testing that a 400 status is received for sending a non-multipart request body to an endpoint that is expecting multipart. See [this gist](https://gist.github.com/cruhland/7b40eef64c257e68ba6c) for a minimal test case.

The multipart parsing code in Netty throws an exception which should be converted into a 400 status by Finch, but somehow escapes. Attempting to deal with the exception using `Endpoint.handle` doesn't work, so I worked around it by defining my own version of `RequestReaders.fileUpload` which catches the exception by wrapping `req.multipart` from [this line](https://github.com/finagle/finch/blob/32f9ee812be534c1e08915998b0a95403270838c/core/src/main/scala/io/finch/RequestReaders.scala#L46) in a `scala.util.Try` instance.

I'm happy to submit a PR to fix this, but I'm not sure if my workaround is the best approach. It may be more appropriate to handle the error at a lower level so that, for example, `com.twitter.finagle.http.Request.multipart` will produce `None` instead of throwing an exception.

cruhland:
This is a simple patch that will produce the desired behavior as described in #484. However, the updated code replaces the exception with a `None` value, losing all information about the cause of the error. This seemed reasonable given that Finch is currently using `None` to indicate missing request components; in the future using `Try`, `Either`, or similar may be more useful. Of course, if `com.twitter.finagle.http.Request.multipart` were updated to handle the exceptions directly, this change would no longer be needed.

### DIFF
--- a/core/src/main/scala/io/finch/RequestReaders.scala
+++ b/core/src/main/scala/io/finch/RequestReaders.scala
@@ -4,7 +4,7 @@ import com.twitter.finagle.http.{Cookie, Request}
 import com.twitter.finagle.http.exp.Multipart.FileUpload
 import com.twitter.finagle.netty3.ChannelBufferBuf
 import com.twitter.io.{Buf, Charsets}
-import com.twitter.util.Future
+import com.twitter.util.{Future, Try}
 
 trait RequestReaders {
 
@@ -43,7 +43,7 @@ trait RequestReaders {
     req.cookies.get(cookie)
 
   private[finch] def requestUpload(upload: String)(req: Request): Option[FileUpload] =
-    req.multipart.flatMap(m => m.files.get(upload).flatMap(fs => fs.headOption))
+    Try(req.multipart).getOrElse(None).flatMap(m => m.files.get(upload).flatMap(fs => fs.headOption))
 
   // A convenient method for internal needs.
   private[finch] def rr[A](i: RequestItem)(f: Request => A): RequestReader[A] =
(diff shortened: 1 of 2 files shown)


core/src/main/scala/io/finch/RequestReaders.scala
```diff
@@ -43,7 +43,7 @@ trait RequestReaders {
     req.cookies.get(cookie)
 
   private[finch] def requestUpload(upload: String)(req: Request): Option[FileUpload] =
-    req.multipart.flatMap(m => m.files.get(upload).flatMap(fs => fs.headOption))
+    Try(req.multipart).getOrElse(None).flatMap(m => m.files.get(upload).flatMap(fs => fs.headOption))
```
vkostyukov: You could do `.toOption` instead.

vkostyukov:
Thanks @cruhland! This is a perfectly reasonable workaround for now. Just one nitpick from my side.

core/src/main/scala/io/finch/RequestReaders.scala
vkostyukov: Oh, we can't use `.toOption` here since it's already `Try[Option[_]]`. Please, ignore this comment.

vkostyukov:
Merging this. Thanks again @cruhland!

PR #488, opened 2015-12-11
add incompletes

teodimoff:
adding incomplete methods. For some reason patch is not working yet..


teodimoff:
I think a custom extractor c.t.f.http.Request -> Hlist will be more appropriate approach.

teodimoff:
i will try to port the coproduct stuff from the https://github.com/travisbrown/circe/blob/master/generic/shared/src/main/scala/io/circe/generic/decoding/DerivedDecoder.scala ,
but i am not sure if this will improve the situation because in finch(as in c.t.f.http.Request) we get request params with different methods(get, getOption,getall etc..) , and hcursor ( used in circe) hides all that.

PR #489, opened 2015-12-12
Typelevel code of conduct

vkostyukov:
Our community grows and I think we need to explicitly support a code of conduct (i.e., Typelevel code of conduct). /cc @travisbrown


travisbrown:
Sorry for the long delay, but this definitely gets a +1 from me.

PR #491, opened 2015-12-12
Add Jackson serialiser for Circe

mtnajim:
Provide support for Jackson serialiser from Circe to improve performance

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -9,7 +9,7 @@ lazy val buildSettings = Seq(
 )
 
 lazy val finagleVersion = "6.31.0"
-lazy val circeVersion = "0.2.1"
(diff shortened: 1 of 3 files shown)


vkostyukov:
The build is broken bue to binary incompatibility in the Circe's and Finch's Jackson versions. Let me update Jackson version to make this green.

travisbrown:
We could probably also downgrade on the circe side if that would make things easier.

vkostyukov:
@travisbrown No wories, I think it makes sense to upgrage Finch to the latest Jackson version anyway. So let's do it now.

vkostyukov:
I rebased this PR here: #492. Thanks again @mtnajim! I will merge once once it's green.

PR #493, opened 2015-12-14
Make Endpoint.mapX methods public

vkostyukov:
As discussed [on Gitter](https://gitter.im/finagle/finch?at=566ea5123078c074765111fe), we could use quite reasonable names for `Endpoint.mapX` functions and therby make them public.


mandubian:
:+1:

PR #496, opened 2015-12-18
Glopart in adopters list

ImLiar:
We've started to using finch while migrating from monolithic finatra app to microservices a few month ago.


vkostyukov:
@ImLiar glad to finally see you in the adopters list! Mind sorting the name? I'm happy to do it myself in the ucomming commit though.

ImLiar:
@vkostyukov here we go

vkostyukov:
Thanks @ImLiar!

Issue #336, opened 2015-06-30
Provide codecs that support streaming


travisbrown:
It's pretty easy to support streaming JSON parsing with Jawn and `Spool`:

``` scala
import com.twitter.concurrent.{ Spool, SpoolSource }
import com.twitter.io.Buf
import com.twitter.util.Future
import jawn.{ AsyncParser, Facade, Parser }

class StreamingParser[J](implicit facade: Facade[J]) {
  private[this] val parser: AsyncParser[J] = Parser.async(mode = AsyncParser.UnwrapArray)
  private[this] val source: SpoolSource[J] = new SpoolSource[J]()

  def feedUtf8(s: String): Unit = parser.absorb(s)
  def feed(buf: Buf): Unit = parser.absorb(
    Buf.ByteBuffer.Owned.extract(buf)
  ).fold(source.raise, _.foreach(source.offer))

  def close(): Unit = source.close()
  def results: Future[Spool[J]] = source()
}
```

It'd be nice to have generic encoding and decoding type classes that support streaming. I started [a draft of this](https://github.com/travisbrown/finch/tree/topic/streaming-codecs) last night, but it still needs a lot of work.

ImLiar:
By the way

It could be a little off topic, but.

We've discussed chunk upload/streaming in finch recently in our company and didnt came to any solution. AFAIK there is a `finagle-stream` which supports async HTTP codec.

Do we have some abstraction for it in finch?

travisbrown:
@ImLiar Not off-topic at all! There's now support for streaming in finagle-httpx proper, so you don't need finagle-stream (which I believe may be on the chopping block)—see e.g. the `Reader`s in `Request` and `Response`.

There's currently no abstraction for this in Finch, but it should be pretty straightforward to use `Reader` with `Request` and `Response` directly, and the streaming codecs I'm proposing here will make this even easier to use in conjunction with e.g. a streaming JSON parser.

ImLiar:
What about some uber abstraction? Iteratee for example. I think it would be perfect.

travisbrown:
@ImLiar In my initial sketch Twitter Util's `Spool` plays a role similar to what an iteratee would provide.

vkostyukov:
We should probably take a look at [AsyncStream](https://github.com/twitter/util/blob/master/util-core/src/main/scala/com/twitter/concurrent/exp/AsyncStream.scala), which is a modern replacement for `Spool`.

travisbrown:
[...]

PR #498, opened 2015-12-18
Async request reader and response writer

ImLiar:
Proof of concept with async reader/response based on finagle AsyncStream.
I think it could be useful while dealing with data streaming and JSON encoding/decoding in future.

AFAIR @travisbrown was digging in this way for Jawn and Spool (#336)

It would be nice to discuss this feature.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -193,6 +193,11 @@ lazy val oauth2 = project
   ))
   .dependsOn(core)
 
+lazy val streaming = project
+  .settings(moduleName := "finch-streaming")
+  .settings(allSettings)
+  .dependsOn(core)
+
 lazy val examples = project
   .settings(moduleName := "finch-examples")
   .settings(allSettings)
@@ -205,7 +210,7 @@ lazy val examples = project
       "com.twitter" %% "twitter-server" % "1.16.0"
     )
   )
-  .dependsOn(core, circe)
+  .dependsOn(core, circe, streaming)
 
 lazy val benchmarks = project
   .settings(moduleName := "finch-benchmarks")
--- a/examples/src/main/scala/io/finch/streaming/Client.scala
+++ b/examples/src/main/scala/io/finch/streaming/Client.scala
@@ -1,0 +1,19 @@
+package io.finch.streaming
+
+import com.twitter.finagle.Http
+import com.twitter.finagle.http.{Method, Request}
+import com.twitter.util.Await
+
+import scala.util.Random
+
+object Client extends App {
+
+  val client = Http.client.withStreaming(enabled = true).newService("127.0.0.1:8082")
+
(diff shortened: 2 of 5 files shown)


vkostyukov:
That's really cool! I love to see some movements in this area!

I had a quick look and here are my thoughts:

1. I think eventually streaming should be part of `core
2. I would for having separate body readers that return `AsyncStream`s: `asyncBody`, `asycnBodyOption`, etc.
3. I'm not sure it's possible but would be super awesome having something like this (reading via streaming a an incoming JSON array)

````scala
val r: RequestReader[AsyncStream[Foo]] = asycnBody.as[Foo]
```

vkostyukov:
Also as a side note I wonder if @travisbrown's [iteratee](https://github.com/travisbrown/iteratee) can be (should be?) used instead of `AsyncStream` thus we could be constent with Circe (but inconcistent with Finagle).

travisbrown:
I'm hoping to wrap up a full circe + iteratee example this weekend. I think it probably makes sense to use `AsyncStream` here, though, at least for now.

FWIW, part of my long-term goal for the iteratee project is for it to be competitive with `AsyncStream` in terms of both usability and performance, and as well-integrated with `com.twitter.util.Future` and `Reader` through the catbird instances.

ImLiar:
The main problem with JSON async parsing is that we have to implement some async interface for finch json libraries. Some of them have this feature out of the box, but some of them - dont. It could be challenging.

And I was thinking about to give user choice of which async container/library to use by providing most popular ones via type classes or smth like this (scalaz (or play?) iteratee, twitter asyncstream, etc). 
Thats why I moved it to separate project.

akiradeveloper:
wow..

vkostyukov:
@ImLiar do you want by any chance to continue working on this? I have a couple comments, but I belive we can make this land in master quite soon. Otherwise, I can merge this as is (after it's rebased) and then address my concers in the upcomming commits.

To clarify, I'm all for upporting very basic `Buf`-based streaming in 0.9.4. Let's make this happen.

vkostyukov:
Just wanted to ping @ImLiar again. Please, let me know if you're still interested in working on this. In the meantime, @akiradeveloper do you want by any chance to pick it up and help me to get this merged? I just thought that as a potential consumer of this feature you might be interested in taking part of it.

akiradeveloper:
@vkostyukov  I am working (not sure I can)

ImLiar:
Sorry, @vkostyukov, I had a fever for last few days. Didnt see your messages.
Anyway, good work here!

vkostyukov:
@ImLiar no worries, hope you feel better!

Thanks again on the initial work, that's a pretty solid step towards streaming in Finch!

PR #507, opened 2016-01-13
Remove double dot when using RequestReader.shouldNot

PChambino:
The `rule` string should not contain a final dot because the final dot is added by `NotValid.getMessage`.


vkostyukov:
Thanks @PChambino! Nice catch!

PR #508, opened 2016-01-19
Introduce an empty output

### LINKED ISSUE #503: An Output API for empty responses
@tarossi raised [this conversation](https://gitter.im/finagle/finch?at=56796c4b69a89a8f552fae15) on Gitter. Looks like the `Output` API is mising a big part on representing an _empty_ responses - responses that doesn't carry any payload/failure and are able to substitue `Output[A]`.

One way to achive that would be think about a third case `case class Empty(status: Status) extends Output[Nothing]` that returns `Buf.Empty`.

vkostyukov:
This is fixes #503 and introduces the third case for `Output` that represents an empty response (no payload). I slightly reworked the implementation and the API according to the Circe's [11th design principle](https://github.com/travisbrown/circe/blob/master/DESIGN.md).

I will let this PR soak here for a couple of days to make sure I'm 100% happy with this solution. Also, I'm going to amend this with backward compatible API.

```scala
scala> Output.empty(Status.Ok)
res0: io.finch.Output[Nothing] = Empty(Meta(Status(200),Map(),List(),None,None))

scala> Output.unit(Status.Ok)
res1: io.finch.Output[Unit] = Empty(Meta(Status(200),Map(),List(),None,None))

scala> Output.payload(100)
res2: io.finch.Output[Int] = Payload(100,Meta(Status(200),Map(),List(),None,None))

scala> Output.failure(new Exception())
res3: io.finch.Output[Nothing] = Failure(java.lang.Exception,Meta(Status(400),Map(),List(),None,None))
```

Note the `Output.unit` method that produces `Output[Unit]` useful for endpoints that always return nothing (a unit response, i.e., `Endpoint[Unit]`). For example, a redirect endpoint:

```scala
val redirect: Endpoint[Unit] = get("from") {
  MovedPermanently.withHeader("Location" -> "to") // MovedPermanently: Output[Unit]
}
```

### DIFF
--- a/core/src/main/scala/io/finch/Output.scala
+++ b/core/src/main/scala/io/finch/Output.scala
@@ -8,37 +8,114 @@ import io.finch.internal.ToResponse
  * An output of [[Endpoint]].
  */
 sealed trait Output[+A] { self =>
+
+  protected def meta: Output.Meta
+  protected def withMeta(meta: Output.Meta): Output[A]
+
+  /**
+   * The status code of this [[Output]].
+   */
+  def status: Status = meta.status
+
+  /**
+   * The header map of this [[Output]].
+   */
+  def headers: Map[String, String] = meta.headers
+
+  /**
+   * The cookie list of this [[Output]].
+   */
+  def cookies: Seq[Cookie] = meta.cookies
+
+  /**
+   * The content type of this [[Output]].
+   */
+  def contentType: Option[String] = meta.contentType
+
+  /**
+   * The charset of this [[Output]].
+   */
+  def charset: Option[String] = meta.charset
+
+  /**
+   * Returns the payload value of this [[Output]] or throws an exception.
+   */
   def value: A
-  def status: Status
-  def headers: Map[String, String]
-  def cookies: Seq[Cookie]
-  def contentType: Option[String]
-  def charset: Option[String]
 
-  def map[B](fn: A => B): Output[B]
-  def flatMap[B](fn: A => Output[B]): Output[B]
-  def flatten[B](implicit ev: A <:< Output[B]): Output[B]
-  def traverse[B](fn: A => Future[B]): Future[Output[B]]
+  def map[B](fn: A => B): Output[B] = this match {
+    case Output.Payload(v, m) => Output.Payload(fn(v), m)
+    case f @ Output.Failure(_, _) => f
+    case e @ Output.Empty(_) => e
+  }
 
-  protected def copy0(
-    headers: Map[String, String] = self.headers,
-    cookies: Seq[Cookie] = self.cookies,
-    contentType: Option[String] = self.contentType,
-    charset: Option[String] = self.charset
-  ): Output[A]
+  def flatMap[B](fn: A => Output[B]): Output[B] = this match {
+    case p @ Output.Payload(v, _) =>
+      val ob = fn(v)
+      ob.withMeta(ob.meta.copy(
+        headers = ob.headers ++ p.headers,
+        cookies = ob.cookies ++ p.cookies,
+        contentType = ob.contentType.orElse(p.contentType),
+        charset = ob.charset.orElse(p.charset)
+        )
+      )
 
-  def withHeader(header: (String, String)): Output[A] = copy0(headers = headers + header)
-  def withCookie(cookie: Cookie): Output[A] = copy0(cookies = cookies :+ cookie)
-  def withContentType(contentType: Option[String]): Output[A] = copy0(contentType = contentType)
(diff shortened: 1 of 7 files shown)


core/src/main/scala/io/finch/Outputs.scala
```diff
@@ -4,47 +4,66 @@ import com.twitter.finagle.http.Status
...
+  // is assembled.
 
   // 2xx
-  def Ok[A](a: A): Output.Payload[A] = Output.Payload(a, Status.Ok)                                               // 200
-  def Created[A](a: A): Output.Payload[A] = Output.Payload(a, Status.Created)                                     // 201
-  def Accepted[A](a: A): Output.Payload[A] = Output.Payload(a, Status.Accepted)                                   // 202
-  def NoContent[A](a: A): Output.Payload[A] = Output.Payload(a, Status.NoContent)                                 // 204
+  def Ok[A](a: A): Output[A] = Output.payload(a, Status.Ok)                                   // 200
+  def Created[A](a: A): Output[A] = Output.payload(a, Status.Created)                         // 201
+  val Accepted: Output[Unit] = Output.unit(Status.Accepted)                                   // 202
```
rpless: Seeing as Accepted "SHOULD include an indication of the request's current status and either a pointer to a status monitor or some estimate of when the user can expect the request to be fulfilled" would it not be better to make this `Output.payload` to encourage users to include a payload?

vkostyukov: Yeah, good question.

I was thinking this kind of information is better to provide in headers rather than a body since this is usually a URL pointing to some "monitor". Example: we create/post a new user, but it's an expensive process so we run it in a background. We could create an `Endpoint[Unit]` for that, not `Endopint[User]` indicating that there is no payload returned (which is true).

rpless: Providing info in headers definitely a valid solution and may well be the right answer given how easy it is to write `Output.payload` if you need to provide more. I think in the example you used, I would tend toward a more CQRS approach and have an `Endpoint[UserCreationStarted]` or something like that. But I guess its not really our place to push this kind of approach on Finch users, so keeping it as `Unit` is fine.

vkostyukov: Great point! I'm in the middle of updating best-practices to cover this very question. I will link the PR from here once it's ready.

core/src/main/scala/io/finch/Output.scala
```diff
@@ -8,37 +8,114 @@ import io.finch.internal.ToResponse
...
+        charset = ob.charset.orElse(charset)
+        )
+      )
+
+    case f @ Output.Failure(_, _) => f
+    case e @ Output.Empty(_) => e
+  }
+
+  def flatten[B](implicit ev: A <:< Output[B]): Output[B] = this match {
+    case p @ Output.Payload(v, _) => v
```
rpless: I'm not sure how I feel about binding to `p` seeing as it is unused. Maybe this is just all the Elixir I've been doing recently but that screams unused variable to me which is a little bit of a code stink in my mind. I do like how the structure of all the branches looks the same though.

core/src/main/scala/io/finch/Output.scala
```diff
@@ -8,37 +8,114 @@ import io.finch.internal.ToResponse
...
+    case f @ Output.Failure(_, _) => f
+    case e @ Output.Empty(_) => e
+  }
+
+  def traverse[B](fn: A => Future[B]): Future[Output[B]] = this match {
+    case p @ Output.Payload(v, _) => fn(v).map(b => p.copy(value = b))
```
rpless: Same as the comment on `flatten`.

core/src/main/scala/io/finch/Output.scala
vkostyukov: Nice catch! Let me fix it.

vkostyukov:
Thinking about this more I decided to do two things:

1. Remove all the 3xx responses from `Outputs` - until we have a better story for redirects, let's just keep them aside.

2. Both `Accepted` and `NoContent` are type-parametrized functions returning empty outputs - those two status codes are pain in the neck - I read the internet twice to see the use cases of them - still don't see the whole picture. That's why having them defined as generic functions might be a good idea for now. As an example, the following two cases are valid usages (simpliar applies to `NoContent`):

```scala
val a: Endpoint[User] = post("users" ? body.as[User]) { u: User =>
  if (u.small) Created(u) // can we create it right now
  else Accpeted.withHeader("Location" -> "foo") // let's do it later
}

val b: Endpoint[Unit] = post("users" ? body.as[User]) { u: User =>
  Accpeted[Unit].withHeader("Location" -> "foo") // let's create it later
}
```

PR #512, opened 2016-01-28
Update divOrFail example in endpoint.md

bskern:
I got confused by this example yesterday and brought it up in the Gitter chat  room. Discovered these examples were in error and ```/ a / b``` in these examples should be ```/int/int``` so I updated the two examples.


vkostyukov:
Thanks you @bskern! This is awesome!

PR #515, opened 2016-01-31
Add streaming request/response

akiradeveloper:
As I asked in finagle's Gitter, S3 API needs to support up to 5GB in a single PUT but unfortunately the max chunk size is limited to 2GB by netty so we need streaming to support larger objects. Aside the PUT issue, streaming is required for GETting large objects (that could be uploaded by multipart upload where each transfer is small but could be huge as a result) from the server.

This work is based on @ImLiar 's PoC. https://github.com/finagle/finch/pull/498
Thanks @ImLiar

### DIFF
--- a/core/src/main/scala/io/finch/RequestReaders.scala
+++ b/core/src/main/scala/io/finch/RequestReaders.scala
@@ -1,9 +1,10 @@
 package io.finch
 
+import com.twitter.concurrent.AsyncStream
 import com.twitter.finagle.http.{Cookie, Request}
 import com.twitter.finagle.http.exp.Multipart.FileUpload
 import com.twitter.finagle.netty3.ChannelBufferBuf
-import com.twitter.io.{Buf, Charsets}
+import com.twitter.io.{Buf, Charsets, Reader}
 import com.twitter.util.{Future, Try}
 
 trait RequestReaders {
@@ -133,6 +134,18 @@ trait RequestReaders {
    * [[Error.NotPresent]] exception.
    */
   val binaryBody: RequestReader[Array[Byte]] = binaryBodyOption.failIfNone
+
+  /**
+   * A [[RequestReader]] that reads a required chunked streaming binary body, interpreted as a `AsyncStream[Buf]`.
+   */
+  val asyncBody: RequestReader[AsyncStream[Buf]] = {
+    def read(chunkSize: Int, reader: Reader): AsyncStream[Buf] =
+      AsyncStream.fromFuture(reader.read(chunkSize)).flatMap {
+        case Some(buf) => buf +:: read(chunkSize, reader)
+        case None => AsyncStream.empty
+      }
+    RequestReader(request => read(Int.MaxValue, request.reader))
+  }
 
   /**
    * A [[RequestReader]] that reads an optional request body, interpreted as a `String`, into an `Option`.
--- a/core/src/main/scala/io/finch/internal/ToResponse.scala
+++ b/core/src/main/scala/io/finch/internal/ToResponse.scala
@@ -1,6 +1,8 @@
 package io.finch.internal
 
-import com.twitter.finagle.http.Response
+import com.twitter.concurrent.AsyncStream
+import com.twitter.finagle.http.{Response, Status, Version}
+import com.twitter.io.{Buf, Reader}
 import io.finch.EncodeResponse
 
 /**
@@ -35,4 +37,18 @@ object ToResponse {
 
       rep
     }
+
+  /**
+   * Provides implicit ToResponse for `AsyncStream[Buf]`.
+   * If it reaches the end of the stream it closes the connection.
+   *
+   * Http server should be initialized with `.withStreaming(enabled=true)`
+   */
+  implicit def asyncToResponse: ToResponse[AsyncStream[Buf]] = new ToResponse[AsyncStream[Buf]] {
+    override def apply(a: AsyncStream[Buf]): Response = {
+      val writable = Reader.writable()
+      a.foreachF(writable.write).ensure(writable.close())
+      Response(Version.Http11, Status.Ok, writable)
+    }
+  }
 }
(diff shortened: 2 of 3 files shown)


core/src/main/scala/io/finch/RequestReaders.scala
```diff
@@ -3,8 +3,9 @@ package io.finch
 import com.twitter.finagle.http.{Cookie, Request}
 import com.twitter.finagle.http.exp.Multipart.FileUpload
 import com.twitter.finagle.netty3.ChannelBufferBuf
-import com.twitter.io.{Buf, Charsets}
+import com.twitter.io.{Buf, Reader, Charsets}
```
vkostyukov: Please, sort these.

You can check style locally by running `sbt scalastyle`.

core/src/main/scala/io/finch/RequestReaders.scala
```diff
@@ -3,8 +3,9 @@ package io.finch
...
 import com.twitter.util.{Future, Try}
+import com.twitter.concurrent.AsyncStream
```
vkostyukov: Please, sort.

core/src/main/scala/io/finch/RequestReaders.scala
```diff
@@ -135,6 +136,18 @@ trait RequestReaders {
   val binaryBody: RequestReader[Array[Byte]] = binaryBodyOption.failIfNone
 
   /**
+   * A [[RequestReader]] that reads a required chunked streaming binary body, interpreted as a `AsyncStream[Buf]`.
+   */
+  def asyncBody(chunkSize: Int): RequestReader[AsyncStream[Buf]] = {
```
vkostyukov: Looks great! I'm not sure we need to give users `chunkSize` to specify. We could (and I think) should use `Int.MaxValue` so we will read the entire chunk sent by a client. Let's make it `val asyncBody`.

core/src/main/scala/io/finch/internal/ToResponse.scala
```diff
@@ -35,4 +37,18 @@ object ToResponse {
 
       rep
     }
+
+  /**
+   * Provides implicit ToResponse for `AsyncStream[Buf]` since vanilla [[Endpoint]] is synchronous by it's nature.
```
vkostyukov: I'm not sure I understand what does "since vanilla [[Endpoint]] is synchronous by it's nature" mean. `Endpoint`s are as asynchronous as they could be since a value they return wrapped within a `Future`. So, let's just leave that out for now.

vkostyukov:
Looks nice! Thank you both, @akiradeveloper, @ImLiar!

akiradeveloper:
CI fails due to type ambiguity.

```
[0m[[31merror[0m] [0m/home/travis/build/finagle/finch/core/src/test/scala/io/finch/OutputSpec.scala:75: ambiguous implicit values:[0m
[0m[[31merror[0m] [0m both value responseToResponse in object ToResponse of type => io.finch.internal.ToResponse[com.twitter.finagle.http.Response][0m
[0m[[31merror[0m] [0m and method asyncToResponse in object ToResponse of type => io.finch.internal.ToResponse[com.twitter.concurrent.AsyncStream[com.twitter.io.Buf]][0m
[0m[[31merror[0m] [0m match expected type io.finch.internal.ToResponse[Nothing][0m
[0m[[31merror[0m] [0m      of.toResponse().content === EncodeResponse.encodeException(of.cause)[0m
[0m[[31merror[0m] [0m                   ^[0m
[0m[[31merror[0m] [0m/home/travis/build/finagle/finch/core/src/test/scala/io/finch/OutputSpec.scala:81: ambiguous implicit values:[0m
[0m[[31merror[0m] [0m both value responseToResponse in object ToResponse of type => io.finch.internal.ToResponse[com.twitter.finagle.http.Response][0m
[0m[[31merror[0m] [0m and method asyncToResponse in object ToResponse of type => io.finch.internal.ToResponse[com.twitter.concurrent.AsyncStream[com.twitter.io.Buf]][0m
[0m[[31merror[0m] [0m match expected type io.finch.internal.ToResponse[Nothing][0m
[0m[[31merror[0m] [0m      of.toResponse().content === Buf.Empty[0m
[0m[[31merror[0m] [0m                   ^[0m
```

core/src/main/scala/io/finch/RequestReaders.scala
akiradeveloper: Int.MaxValue is 2GB. Since it's not always guaranteed that the server has enough room for allocating 2GB buffer I added the parameter `chunkSize`. Not sounds reasonable?

vkostyukov:
Oh, I see what's going on here. You can fix the tests by casting `Output[Nothing]` to some reasonable type (`Nothing` is never what we actually want). Try:

```scala
"Failure" should "propagate cause to response" in {
    check { of: Output.Failure =>
      (of: Output[Unit]).toResponse().content === EncodeResponse.encodeException(of.cause)
    }
  }
```

core/src/main/scala/io/finch/RequestReaders.scala
vkostyukov: The thing is that there is already a `Buf` (well, it's actually `ChannelBuffer`) read from the chunk by Netty. So it's already in memory and we actually can reference it from here w/o copying.

akiradeveloper:
Thanks. I fixed the test and it should pass.

I need your comment about chunkSize please?

core/src/main/scala/io/finch/RequestReaders.scala
akiradeveloper: Thanks, make sense. I will fix

vkostyukov:
Re commend about `chunkSize`, see [here](https://github.com/finagle/finch/pull/515/files#r51354572).
Also do you mind squashing the PR as per [contributors guide](https://github.com/finagle/finch/blob/master/CONTRIBUTING.md#submit-a-pr)?

akiradeveloper:
I squashed all the commits into one.

akiradeveloper:
Wow... the PR decreased coverage more than 1%..

vkostyukov:
This is fine, don't worry about it. Thanks, this looks pretty great!

Issue #504, opened 2015-12-22
Supports Cat's Xor as an Endpoint result


vkostyukov:
As [discussed](https://gitter.im/finagle/finch?at=5679ab54653b30761d75be97) on Gitter. It would be useful to support endpoints that return `Xor[A, B]` representing a failure and success case.

ImLiar:
We've made the same thing with scalaz `\/` in glopart. I could port some code here.

But it would be nice to know how to work with left branch in the end, when we fold it it `Response`. I mean, should we just apply encoder to an error? And what about HTTP status codes?

vkostyukov:
@ImLiar that's a great question actually. I probably need think this over, but for now I imagine this to be something like this:

```scala
case class Error(message: String)
val a: Endpoint[Xor[Error, Int]] = get("div" / int / int) { (a: Int, b: Int) =>
  if (b == 0) Output.payload(Xor.left(Error("Can't divide by 0")), Status.BadRequest)
  else Ok(a / b)
}
```

So basically this should work fine (not need to change `io.finch.internal.*`) as long as we have `EncodeResponse` defined for `Xor[A: EncodeResponse, B: EncodeResponse]`.

ImLiar:
Still need a folder for `Xor[Response, Response] => Response`, but yea, I got it.

ImLiar:
Well, it would be nice to pick `Content-Type` from instance of `EncodeResponse` which is applied to `A` or `B`, depending on `Xor` content. But with current API it seems impossible:

```
new EncodeResponse[Xor[A, B]] {

      override def apply(rep: Xor[A, B]): Buf = ???
      override def contentType: String = ???

    }
```

vkostyukov:
Fixed in #518.

PR #518, opened 2016-01-31
Cat's Xor Encoder

ImLiar:
Simple encoder for #504

Still there is a question about dynamic `contentType`.

### DIFF
--- a/core/src/main/scala/io/finch/Encode.scala
+++ b/core/src/main/scala/io/finch/Encode.scala
@@ -1,6 +1,7 @@
 package io.finch
 
 import cats.Show
+import cats.data.Xor
 import com.twitter.io.Buf
 import shapeless.Witness
 
@@ -60,4 +61,10 @@ object Encode extends LowPriorityEncodeInstances {
 
   implicit val encodeString: TextPlain[String] =
     textPlain(Buf.Utf8.apply)
+
+  implicit def encodeXor[A, B, CT <: String](implicit
+    ae: Encode.Aux[A, CT],
+    be: Encode.Aux[B, CT]): Encode.Aux[A Xor B, CT] = {
+    instance[A Xor B, CT](xor => xor.fold(ae.apply, be.apply))
+  }
 }


vkostyukov:
Thanks @ImLiar!

I'm starting to think that it might be a good idea to decouple content-type from `EncodeResponse` and have a separate type-class for this. Looking at @travisbrown's #329, we might want to represent it as a type-level string (a singleton type). Something like:

```scala
trait ContentType[A] {
  type C <: String
}

object ContentType {
  type Aux[A, C0 <: String] = ContentType[A] { type C = C0}

  def instance[A, C0 <: String](implicit w: shapeless.Witness.Aux[C0]): ContentType.Aux[A, C0] =
    new ContentType[A, C0] {
      type C = C0
      override def toString: String = w.value
    }
}

implicit val ctUser: ContentType[User] = ContentType.instance[User, shapeless.Witness.`"text/plain"`.T]
```

ImLiar:
@vkostyukov 

I'm missing a point of. What about value rather than type based cases of `Content-Type` application? 
I.e. `String` could be `plain/text` or `application/octet-stream`.

travisbrown:
@ImLiar Type class instances are resolved statically, so that process can't consult a value-level string. If you want e.g. different modules to provide encoding logic for JSON and XML content types, it takes a lot more coordination to provide a single encoder that knows about both of those operations and dispatches to the appropriate one at runtime based on a value-level content type string.

vkostyukov:
Thanks @travisbrown!

That's a known problem in Finch: it's hard to distinct two (or more) different content types returned from Finch endpoints. Generally, once you've imported `io.finch.circe._` you've made everything a JSON. Ideally we want to be able to tell for a given endpoint/output what encoder to use (depending on a content-type), which is why we need it life content-type into an actual type so it could affect the implicit encoder being picked up a compiler.

Indeed, encoders for `cats.Xor` and type-level content-type are two different problems that needs to be solved, but I think they overlap - it's worth trying to approach them together.

vkostyukov:
I'm also totally fine to handle content-type as a type in a separate PR (maybe it's even better).

vkostyukov:
Okay, thinking about this more, #329 is exactly what we need right now given that we now have proper tooling (`Endpoint.withConentType`, `Output.withConentType`) to make it available for users. @travisbrown are you interested by any chance to pick it up, rebase and update to the current Finch's state. I'm totally happy to do that myself, if you're swamped with other tasks at this moment.

ImLiar:
@vkostyukov @travisbrown 

AFAIK typed `CT` already in master. May be I should update this PR or there are some pitfalls?

vkostyukov:
Thanks for the follow up, @ImLiar!

Yeah, I guess, we can try doing that. Let's think of something like using matching content-type for left and right cases.

```scala
implicit def encodeXor[A, B, CT <: String](implicit 
  ea: Encode.Aux[A, CT], 
  eb: Encode.Aux[B, CT]): Encode.Aux[Xor[A, B], CT]
```

ImLiar:
Are you sure about single `Content-Type`?

ImLiar:
:checkered_flag:

vkostyukov:
I guess so. Given that which content-type to use is determined on the vey last step (`toService`), endpoints are content-type-less so I don't think we need to support two different content-types for `Xor`. It should be decided at a different layer, how to encode a value.

vkostyukov:
@ImLiar I'd probably vote for not providing a separate example for `Xor`. Perhaps, we could use it in some of the existing examples or even just mention in docs. Let's just remove it for now and figure this out later.

ImLiar:
@vkostyukov I suppose we could move it to cookbook. Just to make sure

vkostyukov:
Yeah, good idea! Let's add cookbook recipe!

vkostyukov:
@ImLiar Mind rebasing this? Feel free to keep just an implementation - I can add docs later in the upcoming commits.

ImLiar:
@vkostyukov Ok. Done

vkostyukov:
Thanks @ImLiar!

PR #522, opened 2016-02-04
Update finagle and scalacheck

vkostyukov:
Finagle 6.33 and Scalacheck 1.13 are here.


travisbrown:
Hey, I just published catbird 0.2 against Finagle 6.31 this morning! :smile:

The new ScalaCheck isn't likely to work against the current ScalaTest (see [this issue](https://github.com/scalatest/scalatest/issues/837)).

vkostyukov:
Thanks Travis! I'm going to keep Scalacheck on 1.12.5 then.

PR #523, opened 2016-02-10
Custom Decoder doc update

aroberts:
I ran into an issue trying to follow the doc for creating a custom decoder for use in a request reader, and updated the doc to show the appropriate function call.  The paragraph following my change may also need an update to be strictly correct, but for now I believe this is an improvement (the old example does not compile under finch-0.9.3 (scala 2.11.7)


vkostyukov:
Thank you @aroberts!

Issue #397, opened 2015-09-09
An endpoint that dispatches requests based on the header value


vkostyukov:
As [discussed on Gitter](https://gitter.im/finagle/finch?at=55ef76a3760979e727e9314d), it would be nice to be able to dispatch requests based on the header values. I'm not quite sure how the API should look like, but the `HeaderMatcher` implementation would be something like

```scala
class HeaderMatcher(header: (String, String)) extends Endpoint[HNil] {
    val (key, value) = header
    import Endpoint._
    def apply(input: Input): Option[(Input, () => Future[HNil])] =
      if (input.requst.headerMap.get(key).exists(_ == value)) Some((input, () => Future.value(HNil)))
      else None
  }
```

PR #524, opened 2016-02-11
Unify RequestReaders and Endpoints

### LINKED ISSUE #254: Single Abstraction for everything: Endpoint

vkostyukov:
This is still WIP (fixes #254 and #397), but I decided to open PR and start a discussion.

The changes proposed here are _mostly_ source compatible with the 0.9.3 version, except for the missing method `RequestReader.flatMap`. This is the only part I didn't manage to port between request readers and endpoints. Everything else works/compiles. For what's worth, all the examples under Finch's `examples` have compiled successfully with this change. Note, no behavior was changed. New endpoints `param`, `header`, etc works in the exact way request readers worked.  

The diff looks scary but mostly because I decided to rework all the tests around (used to be) request readers. Otherwise, the diff is pretty small itself.

I'm going to keep working on porting the rest of the tests to bring the coverage to its previous level.

The following is going to be updated as long as I come up with new pros/cons.

Quick demo:

```scala
scala> import io.finch._
import io.finch._

scala> import com.twitter.finagle.http.Request
import com.twitter.finagle.http.Request

scala> val a = param("a").as[Int].map(_ + 10).map(_.toString)
a: io.finch.Endpoint[String] = param(a)

scala> val e = "plus10" :: a
e: io.finch.Endpoint[String] = plus10/param(a)

scala> val s = e.toService
s: com.twitter.finagle.Service[com.twitter.finagle.http.Request,com.twitter.finagle.http.Response] = <function1>

scala> s(Request("a" -> "1000")).get
res17: com.twitter.finagle.http.Response = Response("HTTP/1.1 Status(404)")

scala> s(Request("/plus10", "a" -> "1000")).get.contentString
res18: String = 1010
```

**What I like**
1. It turns out that it's useful for request readers to have the _match_ ability (endpoints have it). For example, we now have `paramExist`, `headerExist` that match requests with a given header and param. Also, `body`, `bodyOption`, etc are now match only non-chunked requests - which makes a lot of sense to me given that chunked requests have no body. By analogy `asyncBody` now matches only chunked requests and so on. This allows the users to distinct streaming/non-streaming endpoints in a nice and an explicit way.
2. It's now easier to explain Finch to the new users: there is only one abstraction - no confusion between request readers and endpoints.
3. I really like how `finch-oauth2` looks as an `Endpoint` instead of a `RequestReader`. It makes so much sense: `issueAcessToken(..)` is a function that gives you an `Endpoint` that responds with OAuth tokens. Also, since within an endpoint we have an access to `Output` it's now properly handles OAuth2 errors and converts them into a standard responses/outputs that every HTTP client can understand. It's just less work for Finch users to enable OAuth2 support. With endpoints it's a pure joy.

**What I don't like**
1. It might be tricky to reason about endpoint now given that it involves two different stages (1 - match and extract , 2 - evaluate/read). This is only partially true, because, currently `(a / b ? c) / d)` is legal to write and its behavior is unclear: Would it match if request reader `c` fail and `d` matches? Or would it fail?
2. We've lost that nice syntax with `/` and `?` that looked like an URI. Right now those methods as well as request readers' `::` are all the same. I called this method `adjoin` - it takes two endpoints and adjoins them according to the implicit `shapeless.Adjoin` type-class. So `?`, `::` and `/` are just synonyms for `adjoin`.
```scala
scala> val p = param("p"); val q = param("q").as[Int]
p: io.finch.Endpoint[String] = param(p)
q: io.finch.Endpoint[Int] = param(q)

scala> val pq = p / q
pq: io.finch.Endpoint[shapeless.::[String,shapeless.::[Int,shapeless.HNil]]] = param(p)/param(q)

scala> val pq = p :: q
pq: io.finch.Endpoint[shapeless.::[String,shapeless.::[Int,shapeless.HNil]]] = param(p)/param(q)

scala> val pq = p ? q
pq: io.finch.Endpoint[shapeless.::[String,shapeless.::[Int,shapeless.HNil]]] = param(p)/param(q)
```


vkostyukov:
Alright, the code coverage is pretty decent right now - around 80%. I can make it 85/90 later.

With that said, this is pretty much ready to be merged.

PR #531, opened 2016-02-18
Play Json Support

bskern:
This is my initial attempt to get Play JSON support into Finch. Currently all the tests in ```PlayJsonSpec``` fail  with similar errors (Cannot find implicit values for decoder/encoders for the case classes defined in ```JsonCodecProviderProperties```).  The purpose of opening this PR is to get some feedback on what might be missing/why these tests fail. I understand the error message but I can't figure out comparing to other JSON libs whats different and how they are properly allowing Finch to create appropriate implicits.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -126,7 +126,7 @@ lazy val finch = project.in(file("."))
         |import io.circe._
       """.stripMargin
   )
-  .aggregate(core, argonaut, jackson, json4s, circe, benchmarks, test, jsonTest, oauth2, examples)
+  .aggregate(core, argonaut, jackson, json4s, circe, playjson, benchmarks, test, jsonTest, oauth2, examples)
   .dependsOn(core, circe)
 
 lazy val core = project
@@ -186,6 +186,12 @@ lazy val circe = project
       "io.circe" %% "circe-jackson" % circeVersion
     )
   )
+  .dependsOn(core, jsonTest % "test")
+
+lazy val playjson = project
+  .settings(moduleName :="finch-playJson")
+  .settings(allSettings)
+  .settings(libraryDependencies += "com.typesafe.play" %% "play-json" % "2.4.6")
   .dependsOn(core, jsonTest % "test")
 
 lazy val oauth2 = project
--- a/playjson/src/main/io/finch/playjson/package.scala
+++ b/playjson/src/main/io/finch/playjson/package.scala
@@ -15,5 +15,5 @@ package object playjson {
  * @tparam A the type of the data to encode from
  */
   implicit def encodeJson[A](implicit writes: Writes[A]): EncodeResponse[A] =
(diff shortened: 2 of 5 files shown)


playjson/src/test/scala/io/finch/playjson/Bar.scala
```diff
@@ -0,0 +1,3 @@
+package io.finch.playjson
+
+case class Bar(x: Int, y: Boolean)
```
vkostyukov: I guess you should provide `Writes[Bar]` and `Reads[Bar]` to make this compile. Try this:

```scala
import play.api.libs.json._

case class Bar(...)
object Bar {
  implicit val barWrites = Json.writes[Bar]
  implicit val barReads = Json.reads[Bar]
}
```

playjson/src/main/io/finch/playjson/package.scala
```diff
@@ -0,0 +1,19 @@
...
+ * @tparam A the type of the data to decode into
+ */
+  implicit def decodeJson[A](implicit reads: Reads[A]): DecodeRequest[A] =
+    DecodeRequest.instance(input => Try(Json.parse(input).as[A]))
+/**
+ * @param writes Play JSON `Writes` to use for encoding
+ * @tparam A the type of the data to encode from
+ */
+  implicit def encodeJson[A](implicit writes: Writes[A]): EncodeResponse[A] =
+    EncodeResponse.fromString[A]("application/json") {a =>Json.stringify(Json.toJson(a))}
```
vkostyukov: I'd vote for not using curly braces here. Please, use `(a => Json ... )` instead.

build.sbt
```diff
@@ -188,6 +188,12 @@ lazy val circe = project
   )
   .dependsOn(core, jsonTest % "test")
 
+lazy val playjson = project
```
vkostyukov: Please mention new module in the README and `docs/json.md` files.

playjson/src/test/scala/io/finch/playjson/Bar.scala
```diff
@@ -0,0 +1,9 @@
+package io.finch.playjson
+
+import play.api.libs.json._
+case class Bar(x: Int, y: Boolean)
+
+object Bar {
+  implicit val barWrites = Json.writes[Bar]
```
vkostyukov: Looks like it's still failing. Let's try specifying a type here:

```scala
implicit val barWrites: Writes[Bar] = Json.writes[Bar]
implicit val barReads: Reads[Bar] = Json.reads[Bar]
``

vkostyukov:
Okay, I know what the problem is. Sorry, it's been a while a had a look at the JSON implementation and I totally forgot that we use the same case class for testing - `ExampleCaseClass` . So you should remove `Bar.scala` and provide `Writes[ExampleCaseClass]` and `Reads[ExampleCaseClass]` in the `PlayJsonSpec.scala`.

vkostyukov:
Opened a new one at #538.

PR #532, opened 2016-02-18
Update cookbook with custom endpoints

vkostyukov:
Experimenting with the API for building custom patch matching endpoints (also added a new cookbook recipe).

### DIFF
--- a/docs/cookbook.md
+++ b/docs/cookbook.md
@@ -5,6 +5,7 @@
 * [Converting `Error.RequestErrors` into JSON](cookbook.md#converting-errorrequesterrors-into-json)
 * [Defining endpoints returning empty responses](cookbook.md#defining-endpoints-returning-empty-responses)
 * [Defining redirecting endpoints](cookbook.md#defining-redirecting-endpoints)
+* [Defining custom endpoints](cookbook.md#defining-custom-endpoints)
 * [Converting between Scala futures and Twitter futures](cookbook.md#converting-between-scala-futures-and-twitter-futures)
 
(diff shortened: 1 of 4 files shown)


docs/cookbook.md
```diff
@@ -214,6 +215,93 @@ val redirect: Endpoint[Unit] = get("redirect" :: "from") {
...
+  case e: Error.NotPresent => Unauthorized(e)
+}
+
+val getCurrentUser: Endpoint[User] = get("user" :: auth) { u: User =>
+  println(s"Got user: $u")
+  Ok(u) // echo it back
+}
+```
+
+**Note:** Even that an endpoint `auth` can't fail since we explicitly handled its only possible
```
aroberts: Even *though*

docs/cookbook.md
```diff
@@ -214,6 +215,93 @@ val redirect: Endpoint[Unit] = get("redirect" :: "from") {
...
+endpoints_. Even that the name was changed, the request-reader-flavored API (and behaviour) wasn't
+touched at all.
+
+In the following example, we define a new endpoint `foo` that reads an instance of the case class
+`Foo` from the request during the _evaluation_ stage. So it won't affect matching.
+
+```scala
+case class Foo(i: Int, s: String)
+
+val foo: Endpoint[Foo] = (param("i").as[Int] :: param("s").as[Foo]
```
aroberts: missing `)`:
`(param("i").as[Int] :: param("s")).as[Foo]`

vkostyukov:
Thanks for the review @aroberts!

PR #535, opened 2016-02-20
Adds docs example for parsing params

robinske:
As someone new to the project, this standard use case for parsing params was non-obvious from the docs and not included in any of the source examples. Adding to help out others exploring or implementing their first endpoints!


vkostyukov:
Thanks @robinske!

PR #539, opened 2016-02-24
Updating JSON docs for Play Json

bskern:
I never included this update in my initial PR because I wasn't sure I had all the instructions at the time because of compilation issues.

### DIFF
--- a/docs/json.md
+++ b/docs/json.md
@@ -4,6 +4,7 @@
 * [Argonaut](json.md#argonaut)
 * [Jackson](json.md#jackson)
 * [Json4s](json.md#json4s)
+* [PlayJson](json.md#playjson)
 
 --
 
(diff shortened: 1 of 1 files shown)


docs/json.md
```diff
@@ -98,12 +99,23 @@ import org.json4s.DefaultFormats
 implicit val formats: Formats = DefaultFormats ++ JodaTimeSerializers.all
 ```
 
+### PlayJson
```
vkostyukov: I think we should also mention that `finch-playjson` requires `Reads[Foo]` ,`Writes[Foo]` to be defined for a type we want to serialize/deserialize. Let's mention that.

docs/json.md
```diff
@@ -98,12 +99,32 @@ import org.json4s.DefaultFormats
...
+* For any type you want to serialize or deserialize you are required to create the appropriate Play JSON `Reads` and `Writes`. 
+
+```scala
+import io.finch.playjson._
+import play.api.libs.json._
+
+case class Foo(name:String,age:Int)
+
+object Foo {
+  implicit val fooReads = Json.reads[Foo]
```
vkostyukov: Let's put types here `Reads[Foo]` and `Writes[Foo]`.

docs/json.md
```diff
@@ -98,12 +99,32 @@ import org.json4s.DefaultFormats
...
+### PlayJson
+
+* Bring the dependency to the `finch-playjson` module.
+* For any type you want to serialize or deserialize you are required to create the appropriate Play JSON `Reads` and `Writes`. 
+
+```scala
+import io.finch.playjson._
+import play.api.libs.json._
+
+case class Foo(name:String,age:Int)
```
vkostyukov: Please add spaces after `name:`, after `String,` and after `age:`.

docs/json.md
```diff
@@ -98,12 +99,32 @@ import org.json4s.DefaultFormats
...
+
+object Foo {
+  implicit val fooReads = Json.reads[Foo]
+  implicit val fooWrites = Json.writes[Foo]
+}
+```
```
vkostyukov: Please remove these blank lines.

vkostyukov:
Nice! Thanks a lot @bskern!

PR #541, opened 2016-02-24
ContentType as a type-level string

vkostyukov:
Opening this to be able ask for a help when I'm stuck (and I'm almost there). The following works now:

```scala
scala> import io.finch._
import io.finch._

scala> Encode[String]("text/plain")
res1: io.finch.Encode.Aux[String,String("text/plain")] = io.finch.LowPriorityEncodeInstances$$anon$1@4b02ec22

scala> Encode[String]("application/json")
<console>:14: error: could not find implicit value for parameter e: io.finch.Encode.Aux[String,String("application/json")]
       Encode[String]("application/json")
                     ^

scala> Encode[Unit]("application/json")
<console>:14: error: could not find implicit value for parameter e: io.finch.Encode.Aux[Unit,String("application/json")]
       Encode[Unit]("application/json")
                   ^

scala> Encode[Unit]("")
res4: io.finch.Encode.Aux[Unit,String("")] = io.finch.LowPriorityEncodeInstances$$anon$1@57664e8e

scala> Encode[Exception]("")
<console>:14: error: could not find implicit value for parameter e: io.finch.Encode.Aux[Exception,String("")]
       Encode[Exception]("")
                        ^

scala> Encode[Exception]("text/plain")
res6: io.finch.Encode.Aux[Exception,String("text/plain")] = io.finch.LowPriorityEncodeInstances$$anon$1@2bf44d9
```

The problem I'm trying to solve is how to propagate content-type (string) from `Output` all the way down to `ToService` and `ToResponse` type-classes.


core/src/main/scala/io/finch/Encode.scala
```diff
@@ -0,0 +1,60 @@
...
+
+object Encode extends LowPriorityEncodeInstances {
+
+  class Implicitly[A] {
+    def apply[CT <: String](w: Witness.Aux[CT])(implicit
+      e: Encode.Aux[A, CT]
+    ): Encode.Aux[A, CT] = e
+  }
+
+  @inline def apply[A]: Implicitly[A] = new Implicitly[A]
```
ImLiar: What the point of using `Implicitly` class?
Why not like following:
```scala
def apply[A, CT <: String](implicit e: Encode.Aux[A, CT]): Encode.Aux[A, CT] = e
```

vkostyukov: Right. This is just for a convenience so we can pass a method a string `"foo"`, not `Witness.`"foo"`.T`.

ImLiar: Wouldn't it be better to use predefined type aliases instead?

vkostyukov:
Today, @travisbrown, @rpless and I (plus a couple of more people from NE Scala) sit together and tried to brainstorm this PR. I think we (well, of course it was Travis) came up with a brilliant idea. My approach was a dead end because I was trying to use `Output` to enforce the requirement on a a content type, this looks ok at the first glance, but makes not much sense when you think about the whole picture. Finch's `Endpoint[Foo]` doesn't actually contain an information about serialization types (eg: JSON vs. XML) - that's the cornerstone idea of Finch. This is why appropriate decoders are picked at the very last step - `.toService` conversion, not the moment we wrap the value with `Output`. Our take was let's make Finch's endpoints (and output) content-type-free given that a) content-type depends on the encoder (and vice versa) and b) we resolve encoders only when do `.toService` conversion, which happens much later.

What Travis suggested is to encode content-type requirement as part of the to service conversion. This may or may not mean we'd need to change the API or `ToService` type-class. We should probably keep the existing behavior unchanged and default to JSON.

Travises idea was to introduce something like `ServiceBuilder` that takes an HList of possible content types and _captures_ the appropriate encoders at compile time:

```scala
val e: Endpoint[Foo] = ???
val b: Endpoint[String] = ???

val s = ServiceBuilder[Json :: PlainText :: HNil](e :+: b).build // or something like this 
```

This means that we have to endpoints that might be served as JSON or plaintext and we can actually dispatch that on the `Accept` header sent by a client, which is pretty neat. If we want to be more specific and say enforce some of the content type - we won't need to explicitly specific that using a different service builder. For example, an HList that contains an HLists of content-types per each endpoint in coproduct.

My 5c to this is to use something like configurations (via type-classes). The idea is to keep the `toService` call but make it accept different configurations:

```scala
val json: Endpoint[User] = ???
val text: Endpooint[String] = ???

val bothJson = (json :+: text).toService[All[Json :: HNil]]
val jsonAndText = (json :+: text).toService[Each[(Json :: HNil) :: (Text :: HNil)]]
```

It was real fun to talk to you @travisbrown and @rpless today - thanks a lot for you help!

travisbrown:
Thanks for the summary, @vkostyukov! I'm working on this today, but if someone else wants to try out ideas as well, I don't think it'd be a bad idea to have multiple experiments to compare.

fsvehla:
Lovely solution. :heart_eyes:

vkostyukov:
Thanks @travisbrown! I'm on a train right now and trying to start with very simple point using one `ContentType <: String` for entire `ToService` conversion. My general idea is to get rid of that `endpoint.map(folder)` that converts an `Endpoint[C <: Coproduct]` to `Endpoint[Response]`. Let's wrap coproduct endpoint with `Service[Request, Response]` that does the fold underneath an a per response basis. Request comes it - response (of a coproduct type) comes out and we map that to response (by calling `.toResponse` that knows how to convert coproduct). This just looks much clearer for me.

vkostyukov:
Just updated the PR with that idea /cc @travisbrown (not a fully working solution - just a sketch).

vkostyukov:
Something is already working:

```scala
scala> import io.finch._, io.finch.internal._
import io.finch._
import io.finch.internal._

scala> import shapeless._
import shapeless._


scala> val e = body
e: io.finch.Endpoint[String] = body

scala> e.toServiceAs[Witness.`"application/json"`.T]
<console>:22: error: could not find implicit value for parameter ts: io.finch.internal.ToService.Aux[String,String("application/json")]
       e.toServiceAs[Witness.`"application/json"`.T]
                    ^

scala> e.toServiceAs[Witness.`"text/plain"`.T]
res1: com.twitter.finagle.Service[com.twitter.finagle.http.Request,com.twitter.finagle.http.Response] = <function1>
```

vkostyukov:
Almost got this working (passing core/test). Although, I'm not 100% sure about the `foldWithContentType` instance - no surprise my streaming endpoints don't compile.

travisbrown:
@vkostyukov Cool! I've also got something that'll be ready to post later this morning.

core/src/main/scala/io/finch/internal/ToResponse.scala
```diff
@@ -1,54 +1,69 @@
...
+    val rep = Response()
+    val writable = rep.writer
+
+    a.foreachF(writable.write).ensure(writable.close())
+    rep.contentType = w.value
+
+    rep
+  }
+
+  protected object foldWithContentType extends Poly1 {
```
vkostyukov: Stuck here - do no understand why `ToResponse.Aux[C <: Coproduct, ?]` not resolving. Appreciate any pointers.

vkostyukov:
Should be finally green now. Thanks @travisbrown for the tips!

vkostyukov:
@travisbrown how far away your branch from this PR? I wondering if it may make sense to merge this in for now (as a first point) and then build the rest of the story on top of it. Does it make any sense? Or will it just complicate things later?

travisbrown:
Yeah, that sounds reasonable. I'll have something this weekend.

vkostyukov:
Cool! Thanks. I'm merging it then.

PR #550, opened 2016-03-09
Added Project September as an adopter

mandubian:


vkostyukov:
Thank you @mandubian! I'm really glad to see Project September as a Finch's adopter!

Issue #548, opened 2016-03-05
Cats instances for Endpoint


vkostyukov:
I guess we could add `Functor` and `Applicative` for `Endpoint`. Not sure if there is anything else needs to be Cats-friendly - most of the things (i.e., `Output`, `Input`). Maybe we could add `StateT` for `Input` and do something for `Output`, but that might be done later.

ImLiar:
Could I take it?

vkostyukov:
Absolutely! Thanks for the help!

vkostyukov:
I guess we can close this now since there is #518, #551, #555 and #556.

PR #551, opened 2016-03-11
Implementation of cat's Applicative for Endpoint

ImLiar:
#548

### DIFF
--- a/core/src/main/scala/io/finch/Endpoint.scala
+++ b/core/src/main/scala/io/finch/Endpoint.scala
@@ -2,7 +2,7 @@ package io.finch
 
 import scala.reflect.ClassTag
 
-import cats.Eval
+import cats.{Alternative, Eval}
 import com.twitter.finagle.Service
 import com.twitter.finagle.http.{Cookie, Request, Response}
 import com.twitter.util.{Future, Return, Throw, Try}
@@ -494,4 +494,31 @@ object Endpoint {
    * Generically derive a very basic instance of [[Endpoint]] for a given type `A`.
    */
   def derive[A]: GenericDerivation[A] = new GenericDerivation[A]
+
+  implicit val alternativeEndpoint: Alternative[Endpoint] = new Alternative[Endpoint] {
+
+    override def ap[A, B](ff: Endpoint[(A) => B])(fa: Endpoint[A]): Endpoint[B] = fa.ap(ff)
+
+    override def map[A, B](fa: Endpoint[A])(f: (A) => B): Endpoint[B] = fa.map(f)
+
+    override def product[A, B](fa: Endpoint[A], fb: Endpoint[B]): Endpoint[(A, B)] = {
+      fa.ap(fb.map(b => (a: A) => a -> b))
+    }
+
+    override def pure[A](x: A): Endpoint[A] = new Endpoint[A] {
+      override def apply(input: Input): Result[A] = Some(input -> Eval.now(Future.value(Ok(x))))
+    }
+
+    override def pureEval[A](x: Eval[A]): Endpoint[A] = new Endpoint[A] {
+      override def apply(input: Input): Result[A] = Some(input -> x.map(out => Future.value(Ok(out))))
+    }
+
+    override def empty[A]: Endpoint[A] = new Endpoint[A] {
+      override def apply(input: Input): Result[A] = None
+    }
+
+    override def combineK[A](x: Endpoint[A], y: Endpoint[A]): Endpoint[A] = x | y
+
+  }
+
 }
--- a/core/src/test/scala/io/finch/AlternativeEndpointSpec.scala
+++ b/core/src/test/scala/io/finch/AlternativeEndpointSpec.scala
@@ -1,0 +1,14 @@
+package io.finch
+
+import cats.laws.discipline.AlternativeTests
+import com.twitter.finagle.http.Request
+
+class AlternativeEndpointSpec extends FinchSpec with MissingInstances {
+
+  val input = Input(Request(""))
+  implicit val eqEndpointString = eqEndpoint[String](input)
+  implicit val eqEndpointStrings = eqEndpoint[(String, String, String)](input)
+
+  checkAll("Endpoint[String]", AlternativeTests[Endpoint].applicative[String, String, String])
+
+}
--- a/core/src/test/scala/io/finch/FinchSpec.scala
+++ b/core/src/test/scala/io/finch/FinchSpec.scala
@@ -2,9 +2,12 @@ package io.finch
 
 import java.util.UUID
 
+import cats.Eval
 import cats.std.AllInstances
 import com.twitter.finagle.http._
 import com.twitter.io.Buf
+import com.twitter.util.Future
+import io.finch.Endpoint.Result
 import org.scalacheck.{Arbitrary, Gen}
 import org.scalatest.prop.Checkers
 import org.scalatest.{Matchers, FlatSpec}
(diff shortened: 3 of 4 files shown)


core/src/main/scala/io/finch/Endpoint.scala
```diff
@@ -494,4 +490,24 @@ object Endpoint {
    * Generically derive a very basic instance of [[Endpoint]] for a given type `A`.
    */
   def derive[A]: GenericDerivation[A] = new GenericDerivation[A]
+
+  implicit def applicativeEndpoint: Applicative[Endpoint] = new Applicative[Endpoint] {
```
vkostyukov: Why it's `def`, not `val`? My take it should work fine with `val`.

core/src/main/scala/io/finch/Endpoint.scala
```diff
@@ -494,4 +490,24 @@ object Endpoint {
...
+
+    override def pure[A](x: A): Endpoint[A] = new Endpoint[A] {
+      override def apply(input: Input): Result[A] = None
```
vkostyukov: Let's do `Some()` here with `Eval.now`. We already have a value `x`, why not to put it into an `Endpoint` context?

core/src/main/scala/io/finch/Endpoint.scala
```diff
@@ -494,4 +490,24 @@ object Endpoint {
...
+
+    override def product[A, B](fa: Endpoint[A], fb: Endpoint[B]): Endpoint[(A, B)] = {
+      fa.ap(fb.map(b => (a: A) => a -> b))
+    }
+
+    override def map[A, B](fa: Endpoint[A])(f: (A) => B): Endpoint[B] = fa.map(f)
+  }
+
+
+  implicit def functorEndpoint: Functor[Endpoint] = new Functor[Endpoint] {
```
vkostyukov: I don't think we need this. `Applicative` extends `Functor` at the end of the day.

vkostyukov:
Thanks @ImLiar! Just a couple of nits from me.

core/src/main/scala/io/finch/Endpoint.scala
```diff
@@ -269,8 +269,7 @@ trait Endpoint[A] { self =>
    *
    * @param rule text describing the rule being validated
    * @param predicate returns true if the data is valid
-   *
-   * @return an endpoint that will return the value of this reader if it is valid.
+    * @return an endpoint that will return the value of this reader if it is valid.
```
vkostyukov: Oh sorry. Is that extra space before `*` intentional?

ImLiar: No, looks like an reformatting issues

core/src/main/scala/io/finch/Endpoint.scala
```diff
@@ -494,4 +494,20 @@ object Endpoint {
    * Generically derive a very basic instance of [[Endpoint]] for a given type `A`.
    */
   def derive[A]: GenericDerivation[A] = new GenericDerivation[A]
+
+  implicit val applicativeEndpoint: Applicative[Endpoint] = new Applicative[Endpoint] {
+
+    override def pure[A](x: A): Endpoint[A] = new Endpoint[A] {
+      override def apply(input: Input): Result[A] = Some(input -> Eval.now(Future value Ok(x)))
```
vkostyukov: Sorry for the nit: can we do `Future.value(OK(x))` instead?

travisbrown:
Just noticed this PR, and this looks great to me. I have two suggestions:

It'd be handy to have this be an [`Alternative`](https://github.com/typelevel/cats/blob/master/core/src/main/scala/cats/Alternative.scala) instance instead of just `Applicative`. The identity would just return `None`, and `|` serves as the `combineK`.

This captures the nature of `Endpoint` more precisely, allows users to work with them generically as `Alternative`, and someday if Shapeless has a Cats dependency (or someone starts an integration project) I have plans for abstracting out the `:+:` here to use only `Alternative` + `Adjoin`.

core/src/main/scala/io/finch/Endpoint.scala
ImLiar: Yea, no problem.

I prefer this syntax in case if there is a lot of braces. Otherwise it starts looking like a CL dialect :smile:

travisbrown:
Oh, that was only the first suggestion. I actually have two others:

It might be worth providing a `pureEval` implementation in case people want to inject a value into an `Endpoint` lazily. This would just map `Future.value` into the argument and wrap the result in the same way that `pure` currently does.

It also may make sense to provide an `ApplicativeError` instance as well, where the error would end up inside the `Future`. I haven't thought that through yet, though.

travisbrown:
One other thing: as we add these instances it'd be great to check the laws as well—it makes sure the instances are valid, and in the case of things like `composeK` being defined in terms of `|`, provides some additional testing for existing methods. [Here](https://github.com/travisbrown/catbird/blob/master/tests/src/test/scala/io/catbird/util/future.scala)'s what the law testing for `com.twitter.util.Future` in catbird looks like, for example.

I don't think that needs to happen in this PR, but we should keep it in mind (and possibly open an issue if this gets merged without it).

travisbrown:
Also, I'd be happy to see this merged as it is without the `Alternative` or `pureEval` methods—they're easy enough to add separately. So :+1: from me as is, but if you want to include that extra stuff here I'd be glad to review.

ImLiar:
@travisbrown 

Nice to see you remarks. I was thinking about laws but decided to get some feedback first.

ImLiar:
Done with `Alternative` implementation.

vkostyukov:
Looks great! `Alternative` fits so nice here!

travisbrown:
:+1: to the `Alternative` and `pureEval`. This is super nitpicky, but in general I think it's best to group related implementations together in type class instances—e.g. here `pure` and `pureEval` would be next to each other, and `empty` and `combineK`. Just makes it easier to see what's implemented or derived at a glance.

ImLiar:
About `ApplicativeError`

Are we ok to have following signature?

```scala
implicit def alternativeEndpoint[E <: Throwable]: Alternative[Endpoint] with ApplicativeError[Endpoint, E] = new Alternative[Endpoint] with ApplicativeError[Endpoint, E] { ... }
```

`ApplicativeError` extends `Applicative` instance and requires implementation of all those methods. Again

travisbrown:
@ImLiar Right, I'd have one instance provide both `Alternative` and `ApplicativeError`, but I'd suggest just `ApplicativeError[Endpoint, Throwable]` and a `val`, not a generic `def`. The specific `E <: Throwable` will be forgotten once it goes in the future, anyway.

I think that should probably happen in a separate PR, though—I'm still not sure it even makes sense here.

ImLiar:
Well, Sir, in that catse, I'm done here! :smile: 

![Image]
(http://awesomegifs.com/wp-content/uploads/cat-gone.gif)

vkostyukov:
Can we also check `Alternative` laws here in this PR? It should be pretty straightforward given that Finch already has discipline dependency.

ImLiar:
Yea, okay

vkostyukov:
Thanks @ImLiar! This looks good to me. @travisbrown any chance you can have a look before I merge that?

travisbrown:
I'd be happy to see this merged for now, but we really ought to provide a more complete arbitrary instance for `Endpoint` and equality checking that tests more than one input. Maybe we could merge it and immediately open a new issue to track these additions?

vkostyukov:
Thanks! I'll merge this then and file a new ticket about the tests.

travisbrown:
@vkostyukov I've started the work on the law-checking—should be just a few minutes.

vkostyukov:
Thanks a lot @travisbrown! 

In the meantime, I'm experimenting with representing an `Endpoint` as

```scala
final class Endpoint[A](
  val run: Endpoint.State[A]
}

object Endpoint {
  type State[A] = StateT[Option, Input, Eval[Future[Output[A]]]]
}
``` 

I've almost got `core` compiled and it looks pretty clean so far. Really happy with the result.

PR #553, opened 2016-03-11
Update Scala 2.11 and other dependencies and plugins

travisbrown:
This doesn't include the Finagle update, since I didn't know whether we wanted to publish a new finagle-oauth2 first, but it updates everything else.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,7 +8,7 @@ jdk:
   - oraclejdk7
   - oraclejdk8
 script:
-  - sbt ++$TRAVIS_SCALA_VERSION clean coverage test benchmarks/it:test scalastyle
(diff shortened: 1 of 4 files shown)


vkostyukov:
Thanks a lot @travisbrown!

I think I'm going to use :heart: for every single Finch PR now!

travisbrown:
Ugh, looks like the 1.3.3-to-1.3.5 update for sbt-scoverage is having the same effect as I saw [here](https://github.com/travisbrown/iteratee/issues/54). Let me take a closer look…

travisbrown:
Okay, this PR no longer drops coverage to 0.0%!

vkostyukov:
Thanks again! Merging this.

Issue #552, opened 2016-03-11
Add Endpoint.transform


vkostyukov:
As discussed on Gitter, to simplify things like [this](https://gist.github.com/vkostyukov/750196650f381e24c7aa), it would be nice to provide the following method on `Endpoint`:

```scala
final def transform[B](fn: Future[A] => Future[B]): Endpoint[B]
```

vkostyukov:
Fixed by #554.

PR #554, opened 2016-03-11
Endpoint.transform and Output.transform methods

ImLiar:
#552

### DIFF
--- a/core/src/main/scala/io/finch/Endpoint.scala
+++ b/core/src/main/scala/io/finch/Endpoint.scala
@@ -128,6 +128,18 @@ trait Endpoint[A] { self =>
   }
 
   /**
+  * Transforms this endpoint to the given function `Future[Output[A]] => Future[Output[B]]`
+  */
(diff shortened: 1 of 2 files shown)


vkostyukov:
Thanks @ImLiar!

Two comments from my side:

1. It's quite possible to implement this without `Output.transform`:

```scala
  final def transform[B](fn: Future[A] => Future[B]): Endpoint[B] = new Endpoint[B] {
    def apply(input: Input): Endpoint.Result[B] =
      self(input).map {
        case (remainder, output) =>
          (remainder, output.map(f => f.flatMap(o => o.traverse(fn.compose(Future.value)))))
      }
  }
```

2. I'm no longer think we're going to achieve what we wanted with this signature since `Future[A]` from the `fn` signature is not the one we have underneath. Let's change the transform signature to `Future[Output[A]] => Future[Output[B]]` so we can actually pass a future we're interested in. This should simplify the implementation a lot.

ImLiar:
@vkostyukov done

vkostyukov:
It's merged! Thanks @ImLiar!

PR #557, opened 2016-03-13
Add module for streaming with iteratees [WIP]

travisbrown:
This is just a sketch (from experiments at the Typelevel summit last week) but I wanted to go ahead and open a PR to track it. Needs tests and examples, and it should be a little cleaner and more efficient to bypass `AsyncStream` and use `Reader` / `Writer` directly.

### DIFF
--- a/iteratee/src/main/scala/io/finch/iteratee/package.scala
+++ b/iteratee/src/main/scala/io/finch/iteratee/package.scala
@@ -1,0 +1,51 @@
+package io.finch
+
+import cats.data.OptionT
+import com.twitter.concurrent.AsyncStream
+import com.twitter.finagle.http.{ Response, Status, Version }
+import com.twitter.io.{ Buf, Reader }
+import com.twitter.util.Future
+import io.catbird.util._
+import io.circe.{ Decoder, Encoder }, io.circe.streaming._
+import io.finch.internal.ToResponse
+import io.iteratee._
+
+package object iteratee extends Module[Future] {
+  final def fromAsyncStream[A](as: AsyncStream[A]): Future[Enumerator[Future, A]] = as.head.map {
+    case None => empty[A]
+    case Some(h) => generateM((h, as.tail)) {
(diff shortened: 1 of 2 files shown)


iteratee/src/main/scala/io/finch/iteratee/package.scala
```diff
@@ -0,0 +1,51 @@
...
+  ): Endpoint[Enumerator[Future, Buf]] =
+    asyncBody.mapAsync(fromAsyncStream).map(
+      _.mapE(decodeBuf[A].andThen(transformation).andThen(encodeToBuf))
+    )
+    
+  private[this] def foreach[A](f: A => Unit): Iteratee[Future, A, Unit] = fold(()) {
+    case (_, a) => f(a)
+  }
+
+  implicit final val enumeratorToResponse: ToResponse[Enumerator[Future, Buf]] =
```
vkostyukov: Given that we now have `Encode.ApplicationJson` can we write that in a generic way and convert `Enumerator[Future, A: Encode.ApplcationJson]` into streamed JSON array? I mean we can have both `Buf` and `A: Encode.ApplicationJson` defined here.

vkostyukov:
Going back to this.

@travisbrown I think it would be totally awesome if we could include this into `finch-circe` (like Circe support in Finch comes with built-in iteratee-powered streaming). I think we have all the required machinery now (including `Decode` and `Encode` type-classes parametrized with content-types) to make this happen.

What do you think @travisbrown?

travisbrown:
@vkostyukov Sounds good to me, if you don't mind the extra iteratee.io dependency for finch-circe. I might not get a chance to get things updated until after Scala by the Bay, though.

vkostyukov:
I'm closing this as @ImLiar took good care about iteratee in #812.

Thanks @travisbrown for the inspiration and sharing your ahead-of-time perspective!

travisbrown:
Yes, thanks @ImLiar—it's great to see this get merged!

PR #559, opened 2016-03-13
Use Cats StateT to reprensent an Endpoint

vkostyukov:
I was experimenting with representing an `Endpoint[A]` as `StateT[Input, Option, Eval[Future[Output[A]]]]` . This approach seems pretty reasonable to me given how it simplifies a couple of things and make it a great use of Cats abstractions.

I'm not 100% sure, I'm happy with the current implementation. But I think it might be a reasonable step forward - I was focusing on a minimized diff rather than a super clean implementation. Maybe as a next step we could think about cleaning up `item` and `toString` methods on `Endpoint` so the entire story will be much simpler.

### DIFF
--- a/core/src/main/scala/io/finch/Endpoint.scala
+++ b/core/src/main/scala/io/finch/Endpoint.scala
@@ -3,6 +3,8 @@ package io.finch
 import scala.reflect.ClassTag
 
 import cats.{Alternative, Eval}
+import cats.data.StateT
+import cats.std.option._
 import com.twitter.finagle.Service
 import com.twitter.finagle.http.{Cookie, Request, Response}
 import com.twitter.util.{Future, Return, Throw, Try}
@@ -56,16 +58,22 @@ import shapeless.ops.hlist.Tupler
  *   Http.server.serve(foobar.toService)
  * }}}
  */
-trait Endpoint[A] { self =>
+abstract class Endpoint[A] extends Serializable { self =>
 
-  type ContentType <: String
+  /**
+   * The [[StateT]] instance underneath this endpoint.
+   */
+  def embed: Endpoint.State[A]
 
+  /**
+   * Request item (part) that's this endpoint work with.
+   */
   def item: items.RequestItem = items.MultipleItems
 
   /**
    * Maps this endpoint to either `A => Output[B]` or `A => Output[Future[B]]`.
    */
-  def apply(mapper: Mapper[A]): Endpoint[mapper.Out] = mapper(self)
+  final def apply(mapper: Mapper[A]): Endpoint[mapper.Out] = mapper(this)
 
   // There is a reason why `apply` can't be renamed to `run` as per
   //   https://github.com/finagle/finch/issues/371.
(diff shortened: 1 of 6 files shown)


mandubian:
Have you tested the impact on perf and potential issues with infamous SI-2712 because the type becomes quite complicated ?

core/src/main/scala/io/finch/Endpoint.scala
```diff
@@ -56,16 +58,22 @@ import shapeless.ops.hlist.Tupler
...
  * }}}
  */
-trait Endpoint[A] { self =>
+abstract class Endpoint[A] extends Serializable { self =>
 
-  type ContentType <: String
+  /**
+   * The [[StateT]] instance underneath this endpoint.
+   */
+  def run: Endpoint.State[A]
```
travisbrown: Maybe call this something like `state` or `underlying` to avoid confusion with `.run.run`?

travisbrown:
@vkostyukov @mandubian I'm also curious about the performance implications. I'd guess it's not going to be noticeable, since the main difference here is some extra allocations that only happen once for each decoder (not per-request), but I'm not 100% sure about that.

mandubian:
if it happens only at Decoder initialization, it's OK... let's be sure but anyway I like that we identify clearly the real concepts behind Endpoint!

core/src/main/scala/io/finch/Endpoint.scala
vkostyukov: Yeah, that reminded me fs2 (scalaz-streams) with `.run.run` and I though it would be ok to do the same thing here.

I guess I like `state`.

travisbrown: Yeah, I don't think `run` is terrible, but I don't personally find the convention quite as clear here. I don't feel strongly either way.

vkostyukov:
I'm going to put together a couple of benchmarks as part of of the separate PR so we can measure the difference. Let's hold on this until we have a clearer picture about performance.

core/src/main/scala/io/finch/Endpoint.scala
vkostyukov: Renamed `run` to `embed`.

vkostyukov:
I run a number of benchmarks and I don't see any difference in the load test (wrk), but I do see some significant changes in micro benchmarks. Results are here: https://gist.github.com/vkostyukov/7a81a6d8493482fa3115.

Moreover, the overall picture looks not so bad (`StateT` adds 2-4% of running time and allocations), but the difference in matcher looks pretty dramatic. I'll try to figure out what's going on there.

vkostyukov:
I looked into the allocation profile and I noticed that a number of allocated `Option` objects has been increased dramatically. It actually about 36% of all the allocations for the matcher benchmark. Looking at the `StateT` implementation, I think, I see the reason to this. `StateT` wraps `val runF: F[S => F[(S, A)]]`, which has a double-nested `F` type constructor (an `Option` in our case), while our previous version was just `S => F[(S, A)]`. That simply means we allocate twice as much options with `StateT` rather that with just a function.

Not sure what do about that, but I'm now less optimistic about using `StateT` in Finch. Any thoughts @mandubian, @travisbrown?

mandubian:
Yes, StateT in cats is double-wrapped and it has a cost... A custom structure doing the same restricted to our needs and a bit optimized could be interesting...

vkostyukov:
Hmm. I think I like the idea of having an internal version of `State[A]` fixed to `Input => Option[(Input, Eval[Future[Output[A]]]]`.

vkostyukov:
Just to mention that further experiments in this direction should be based on [Macavity](https://github.com/travisbrown/macavity).

vkostyukov:
I'm abandoning this for efficiency reasons.

PR #563, opened 2016-03-15
Replace our benchmark with wrk

vkostyukov:
I was playing with `wrk` as a replacement for our service benchmark (fixes #563). I think I like that more than use JMH for doing load-testing.

Here I've created two separate applications (Finch + Circe vs. Finagle + Jackson) that does the same thing: round trip a simple JSON object. I tried to run it on my laptop with 4 threads maintaining 24 connections and I got Finch running on 82% of Finagle's throughput (I think that looks quite realistic).

We're still using JMH for micro-benchmarking separate components, but I think it just makes much more sense to use something like wrk for a load-testing.


travisbrown:
This isn't directly related, but for what it's worth I've just been profiling some simple JSON processing examples with circe and have some pretty big improvements ready for 0.4.0: https://github.com/travisbrown/circe/pull/233

vkostyukov:
Great news! And I think that's pretty much related here. I like the idea of reasoning of Finch's performance in terms of Finagle's performance, We can plan a performance milestone and try to reduct the gap there. There is definitely be a gap, but I think we should be able to shrink that to 5-10% quite easily (eg: try `Rerunnable`).

Issue #520, opened 2016-02-02
Enable streaming for JSON arrays


vkostyukov:
We do support some pretty basic `Buf`-based streaming in Finch via `AsyncStream[Buf]`. Let's make it stream JSON objects. The user-facing API I imagine is something like:

```scala
case class User(id: Int)
val b: RequestReader[AsyncStream[User]] = asyncBody.as[User] // we use the same pattern
val tr: ToResponse[AsyncStream[User]] = implicitly[ToResponse[AsyncStream[User]]] // should compile
```

To solve this, we need a nice way of distinct encoders (JSON vs. XML vs. whatever). Perhaps, #518 will give us such.

vkostyukov:
Let's start with supporting endpoints that return `AsyncStream[User]` (streaming response).

ImLiar:
Working on it

ImLiar:
It's gonna be a hard time to implement async request reading from JSON since some JSON libraries doesn't support stream parsing or do it at very low "manual" level.

vkostyukov:
Done in #566.

PR #566, opened 2016-03-15
JSON streaming for `AsyncStream[A]`

ImLiar:
#520 

By the way, it didn't work properly for `Buf` without `Transfer-Encoding: chunked`

### DIFF
--- a/core/src/main/scala/io/finch/internal/ToResponse.scala
+++ b/core/src/main/scala/io/finch/internal/ToResponse.scala
@@ -3,7 +3,7 @@ package io.finch.internal
 import com.twitter.concurrent.AsyncStream
 import com.twitter.finagle.http.Response
 import com.twitter.io.Buf
-import io.finch.Encode
+import io.finch.{Encode, JSON}
 import shapeless._
 
 /**
@@ -44,13 +44,27 @@ trait LowPriorityToResponseInstances {
 
 object ToResponse extends LowPriorityToResponseInstances {
 
+  private val newLineDelimeter = Buf.Utf8("\n")
+
   implicit def asyncBufToResponse[CT <: String](implicit
     w: Witness.Aux[CT]
-  ): Aux[AsyncStream[Buf], CT] = instance { a =>
+  ): Aux[AsyncStream[Buf], CT] = asyncStreamResponseBuilder(identity)
+
+  implicit def asyncStreamToResponse[A](implicit
+    e: Encode.ApplicationJson[A]
+  ): Aux[AsyncStream[A], JSON] = {
+    asyncStreamResponseBuilder[A, JSON](data => e(data).concat(newLineDelimeter))
+  }
+
+  private def asyncStreamResponseBuilder[A, CT <: String](writer: A => Buf)(implicit
+    w: Witness.Aux[CT]
+  ): Aux[AsyncStream[A], CT] = instance { a =>
     val rep = Response()
+    rep.setChunked(true)
+
     val writable = rep.writer
 
-    a.foreachF(writable.write).ensure(writable.close())
+    a.foreachF(data => writable.write(writer(data))).ensure(writable.close())
     rep.contentType = w.value
 
     rep
--- a/core/src/main/scala/io/finch/package.scala
+++ b/core/src/main/scala/io/finch/package.scala
@@ -1,5 +1,7 @@
 package io
 
+import shapeless.Witness
+
 /**
  * This is a root package of the Finch library, which provides an immutable layer of functions and
  * types atop of Finagle for writing lightweight HTTP services.
@@ -11,6 +13,10 @@ package object finch extends Endpoints with Outputs with ValidationRules {
 
   @deprecated("Use io.finch.Encode instead", "0.11")
   type EncodeResponse[A] = Encode[A]
+
+  type JSON = Witness.`"application/json"`.T
+  type Text = Witness.`"text/plain"`.T
+  type HTML = Witness.`"text/html"`.T
 
   object items {
     sealed abstract class RequestItem(val kind: String, val nameOption:Option[String] = None) {
--- a/examples/src/main/scala/io/finch/streaming/Main.scala
+++ b/examples/src/main/scala/io/finch/streaming/Main.scala
@@ -3,11 +3,13 @@ package io.finch.streaming
 import java.util.concurrent.atomic.AtomicLong
 
 import com.twitter.concurrent.AsyncStream
-import com.twitter.finagle.Http
+import com.twitter.finagle.{Http, Service}
+import com.twitter.finagle.http.{Request, Response, Status}
 import com.twitter.io.Buf
-import com.twitter.util.{Await, Try}
+import com.twitter.util.{Await, Future, Try}
+import io.circe.generic.auto._
 import io.finch._
-import shapeless.Witness
+import io.finch.circe._
 
 /**
  * A simple Finch application featuring very basic, `Buf`-based streaming support.
@@ -53,6 +55,10 @@ object Main extends App {
     Ok(loop(1, 0).map(i => Buf.Utf8(i.toString)))
   }
 
+  val exampleGenerator: Endpoint[AsyncStream[Example]] = get("examples" :: int) { num: Int =>
+    Ok(AsyncStream.fromSeq(List.tabulate(num)(i => Example(i))))
+  }
+
   // This endpoint takes a streaming request (a stream of numbers) and responds
   // to each number (chunk) a sum that has been collected so far.
   //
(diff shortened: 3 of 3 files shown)


core/src/main/scala/io/finch/internal/ToResponse.scala
```diff
@@ -48,6 +48,8 @@ object ToResponse extends LowPriorityToResponseInstances {
     w: Witness.Aux[CT]
   ): Aux[AsyncStream[Buf], CT] = instance { a =>
     val rep = Response()
+    rep.setChunked(true)
```
vkostyukov: Ahh, nice catch!

core/src/main/scala/io/finch/internal/ToResponse.scala
```diff
@@ -56,6 +58,21 @@ object ToResponse extends LowPriorityToResponseInstances {
     rep
   }
 
+  implicit def asyncStreamToResponse[A, CT <: String](implicit
```
vkostyukov: Probably, I wasn't super clear on what I mean. My idea was to actually stream JSON array `{ obj, obj, obj }`, not a sequence of objects. Given that we now can ask for a JSON encoder, we can define this behavior pretty simple. I had something like [this in mind](https://gist.github.com/vkostyukov/463b9a7c992e2bf93fca).

I'm also thinking that we can expose that via the `ToResponse[AsyncStream[Buf]]` instance (reuse it) in some way.

ImLiar: For now it looks like this:

```
curl http://localhost:8081/examples/3
{"x":0}{"x":1}{"x":2}
```

It works correctly for concatenated JSON case [https://en.wikipedia.org/wiki/JSON_Streaming](url)

ImLiar: It could be nice to use new-line delimeter, but it works correctly in terms of _streaming_

vkostyukov: Hm. I guess you're right! Maybe make it new-line delimited by default? I wonder which approach is considered a default.

vkostyukov: In this case, we don't even need a new case for `AsyncStream[A]`, we can generalize our previous version `AsyncStream[Buf]` to `AsyncStream[A: Encode]`.

ImLiar: > The use of a newline as a delimiter enables this format to work very well with traditional line-oriented UNIX tools.

I think we do need EOL :)

ImLiar: > In this case, we don't even need a new case for AsyncStream[A], we can generalize our previous version AsyncStream[Buf] to AsyncStream[A: Encode].

Yea, but to use EOL we need a special case for `AsyncStream[A]`. I did some refactoring

core/src/main/scala/io/finch/internal/ToResponse.scala
```diff
@@ -46,11 +46,22 @@ object ToResponse extends LowPriorityToResponseInstances {
...
     w: Witness.Aux[CT]
-  ): Aux[AsyncStream[Buf], CT] = instance { a =>
+  ): Aux[AsyncStream[Buf], CT] = asyncStreamResponseBuilder(identity)
+
+  implicit def asyncStreamToResponse[A, CT <: String](implicit
+    e: Encode.Aux[A, CT],
+    w: Witness.Aux[CT]
+  ): Aux[AsyncStream[A], CT] = asyncStreamResponseBuilder(data => e(data) concat Buf.Utf8("\n"))
+
+  private def asyncStreamResponseBuilder[A, CT <: String](writer: A => Buf)(implicit
```
vkostyukov: This would work for sure, but my take is we don't need `asyncStreamResponseBuilder` since there is already `Encode[Buf]` defined as `identity`. Just remove the previous version and keep your one - everything should work.

core/src/main/scala/io/finch/internal/ToResponse.scala
```diff
@@ -46,11 +46,22 @@ object ToResponse extends LowPriorityToResponseInstances {
 
   implicit def asyncBufToResponse[CT <: String](implicit
     w: Witness.Aux[CT]
-  ): Aux[AsyncStream[Buf], CT] = instance { a =>
+  ): Aux[AsyncStream[Buf], CT] = asyncStreamResponseBuilder(identity)
+
+  implicit def asyncStreamToResponse[A, CT <: String](implicit
+    e: Encode.Aux[A, CT],
+    w: Witness.Aux[CT]
+  ): Aux[AsyncStream[A], CT] = asyncStreamResponseBuilder(data => e(data) concat Buf.Utf8("\n"))
```
vkostyukov: Also, let's extract `Buf.Utf8("\n")` into a val so we can save on allocations.

core/src/main/scala/io/finch/internal/ToResponse.scala
vkostyukov: Ahh I see. You want to make sure that If we have `AS[Buf]` we don't add new lines?

core/src/main/scala/io/finch/internal/ToResponse.scala
ImLiar: Won't work because there is a difference between buffer streaming and JSON streaming.

- In case of JSON streaming we add EOL symbol
- In case of buffer streaming we cant add EOL since it could completly break data in stream

Different behaviours

ImLiar:
Oh, dat _usability_ with inline comments. I think it would be nice to continue discussion here :smile:

core/src/main/scala/io/finch/internal/ToResponse.scala
```diff
@@ -44,13 +44,26 @@ trait LowPriorityToResponseInstances {
...
 object ToResponse extends LowPriorityToResponseInstances {
 
+  private val newLineDelimeter = Buf.Utf8("\n")
+
   implicit def asyncBufToResponse[CT <: String](implicit
     w: Witness.Aux[CT]
-  ): Aux[AsyncStream[Buf], CT] = instance { a =>
+  ): Aux[AsyncStream[Buf], CT] = asyncStreamResponseBuilder(identity)
+
+  implicit def asyncStreamToResponse[A, CT <: String](implicit
```
vkostyukov: Shouldn't we fix content type to JSON here given that this particular encoding only makes sense for JSON objects?

core/src/main/scala/io/finch/internal/ToResponse.scala
```diff
@@ -44,13 +44,26 @@ trait LowPriorityToResponseInstances {
...
+    e: Encode.Aux[A, CT],
+    w: Witness.Aux[CT]
+  ): Aux[AsyncStream[A], CT] = asyncStreamResponseBuilder(data => e(data) concat newLineDelimeter)
```
vkostyukov: Please, add dots around  `concat`.

examples/src/main/scala/io/finch/streaming/Example.scala
```diff
@@ -0,0 +1,3 @@
+package io.finch.streaming
+
+case class Example(x: Int)
```
vkostyukov: Let's define this in the Main.scala file.

examples/src/main/scala/io/finch/streaming/Main.scala
```diff
@@ -53,6 +56,11 @@ object Main extends App {
     Ok(loop(1, 0).map(i => Buf.Utf8(i.toString)))
   }
 
+  val exampleGenerator: Endpoint[AsyncStream[Example]] = get("examples" :: int) { num: Int =>
+    val pseudoStream = AsyncStream.fromSeq(List.tabulate(num)(i => Example(i)))
+    Ok(pseudoStream)
```
vkostyukov: Maybe inline that into `Ok` directly?

core/src/main/scala/io/finch/internal/ToResponse.scala
ImLiar: Yea, right

core/src/main/scala/io/finch/package.scala
```diff
@@ -12,6 +14,10 @@ package object finch extends Endpoints with Outputs with ValidationRules {
   @deprecated("Use io.finch.Encode instead", "0.11")
   type EncodeResponse[A] = Encode[A]
 
+  type JSON = Witness.`"application/json"`.T
```
ImLiar: Let's make those sources more friendly!

vkostyukov:
Merging this in. Thanks again @ImLiar!

Issue #562, opened 2016-03-14
Refactor Decode type-class


vkostyukov:
My take is that `Decode` should be refactored in a form of

``` scala
trait Decode[A] {
  type CT <: String
  def apply(b: Buf): Try[A]
}
```

Note that
1. it has a type-level content-type so we can control what decoders to use `body.asJson[User]` vs. `body.asText[User]`
2. It takes a `Buf` instead of `String` so we can move towards #511 and convert `Buf` body directly to JSON

This makes both `Encode` and `Decode` type-class be body-oriented, which makes the whole picture a bit clearer: right now `Encode` only used for bodies and `Decode` used for everything (params, headers, bodies). We'd need to figure out what to do with `param("foo").as[Int]` - I guess we should probably use a different type-class there that takes a `String` instead of `Buf`. This would help us move towards #528.

ImLiar:
How about abstract `apply` input? `Decode[I, A]` would allow us to work not only with `Buf` or `String` but with `AsyncStream[_]` too.

Something like following:

```
trait Decode[I, A] {
  type CT <: String
  def apply(i: I): Try[A]
}
```

ImLiar:
It also wouldn't require of us to teach all of JSON libraries to work with `Buf` or `AsyncStream[Buf]`, but only those which has some kind of support for those mechanics.

ImLiar:
@vkostyukov Do you mind if I try to refactor `Decode` this way?

vkostyukov:
I'm thinking about making a progress on this (a baby step, actually). Will try to avoid any API-breaking changes as much as possible, but the idea (I like) is beeing explicitly about decoding:

```scala
param("foo").as[Text.Plain, Int]
body.as[Application.Json, User]
````

Is that too verbose? I kind of like it for `body`, but not sure it makes a lot of sense for `param`. Perhaps making `params` and `headers` special cases (using different type-class?) would help here.

Issue #511, opened 2016-01-27
Eliminate meaningless to-string conversion


vkostyukov:
When it comes to Netty -> Finagle -> Finch, the HTTP body lifecycle is `ChannelBuffer` -> `Buf` -> `String` -> `CaseClass`. If we'd be able to teach jawn to work with `Buf` instead of `String`, we'd be able to go directly from `Buf` to `CaseClass` thereby eliminating that meaningless to-string conversion, which in fact alllocates O(n^2) bytes.

This would be a huge win for Finch users, but this requires some refactoring around `DecodeRequest`. The idea of top of my head is to actually keep `DecodeRequest` a string-based type-class and use it for params, headers, etc (i.e., `param("foo").as[Int]`) and introduce a new type-class `FromRequest = Request => A` (as a symmetric version to `ToResponse`) to be able to grab `content: Buf` from the request if we need it and not perform the to-string conversion. In this case, `body` readers should use the `FromRequest` type-class via the `as` method.

I'm totally fine to start with Circe (make it decode JSON from `Buf`) and keep the rest of the JSON lirbaries we support working with strings.

vkostyukov:
Fixed in #671 (kudos to @rpless).

PR #570, opened 2016-03-16
#562 Decode Refactoring

ImLiar:
This PR is my vision of `Decode` refactored implementation #562

I propose to introduce second type parameter `I` (for Input) for `Decode` trait:
```
trait Decode[I, A] {
  type ContentType <: String
  def apply(i: I): Try[A]
}
```

In this case we could iteratively move from `String` to `Buf` as @vkostyukov suggested in #511 and support multiple inputs via kind of "ad-hoc" polymorphism. I.e. it'll be super useful for implementation of Decode for `AsyncStream` (#520) where `String` or `Buf` couldn't be a case obviously. Also it will allow us to live some time with current JSON decoders which doesn't support `Buf` without dirty hacks (or doesn't support it at all :fearful:)

I'd be very glad of your advices, opinions and suggestions.

### DIFF
--- a/core/src/main/scala/io/finch/Decode.scala
+++ b/core/src/main/scala/io/finch/Decode.scala
@@ -3,23 +3,36 @@ package io.finch
 import java.util.UUID
 
 import com.twitter.util.{Return, Throw, Try}
-import shapeless.{::, Generic, HNil}
+import shapeless.{::, Generic, HNil, Witness}
 
 /**
  * An abstraction that is responsible for decoding the value of type `A`.
  */
-trait Decode[A] {
-  def apply(s: String): Try[A]
+trait Decode[I, A] {
+  type ContentType <: String
+  def apply(i: I): Try[A]
 }
 
 trait LowPriorityDecodeInstances {
 
+  type Aux[I, A, CT <: String] = Decode[I, A] { type ContentType = CT }
+
+  type ApplicationJson[I, A] = Aux[I, A, Witness.`"application/json"`.T]
+  type TextPlain[I, A] = Aux[I, A, Witness.`"text/plain"`.T]
+
   /**
    * Creates an instance for a given type.
    */
-  def instance[A](f: String => Try[A]): Decode[A] = new Decode[A] {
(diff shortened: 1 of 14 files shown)


vkostyukov:
Thanks a lot @ImLiar! This looks pretty much similar to what I had in mind. Although, I still have a a couple concerns about using the same decoders for for both bodies and the rest of the HTTP entires (params, headers). Here is why:

1. Bodies is usually the most interesting part where different JSON/XML/whatever libraries kicks in
2. Params, headers, etc are all string-based and it should be easy to separate their concept from bodies 
3. I want to keep the API unchanged so when you say `param.as[Int]` it should use text decoder, but for `body.as[User]` - a JSON one. (is the later is broken right now?)

If don't want to repeat ourself, it should be pretty simple to implement `Decode.Aux[A, Text]` in terms of what we will have for params, headers, etc.

Anyways, I can only see all of this possible to do once our JSON decoders will know how to decode `Buf`. So there's should some major refactoring happen.

core/src/main/scala/io/finch/Decode.scala
```diff
@@ -3,77 +3,94 @@ package io.finch
...
+  type ContentType <: String
+  def apply(i: I): Try[A]
 }
 
 trait LowPriorityDecodeInstances {
 
+  type Aux[I, A, CT <: String] = Decode[I, A] { type ContentType = CT }
+
+  type ApplicationJson[I, A] = Aux[I, A, Witness.`"application/json"`.T]
+  type TextPlain[I, A] = Aux[I, A, Witness.`"text/plain"`.T]
```
teodimoff: @ImLiar if you force the type to string here it will compile (at least compiles my code):
```
type TextPlain[A] = Aux[String,A, Witness.`"text/plain"`.T]
```
i have tried with implicit view but it did not work: 
```
type TextPlain[I <: String, A] = Aux[I, A, Witness.`"text/plain"`.T]
```
we can do the same for applicationJson. first leave it as string and then make it c.t.u.Buf when it is ready.
Again there might be better.

PR #571, opened 2016-03-17
Revert sbt-scoverage

travisbrown:
Okay, I made a kind of annoying mistake last week.

Last fall I updated circe to sbt-scoverage 1.3.3, ran into [a terrible bug](https://github.com/travisbrown/circe/pull/72) that meant the 2.10 builds weren't running, and then reverted the update back to sbt-scoverage 1.2.0.

Last week I forgot entirely about this, saw that sbt-scoverage 1.3.5 was out, and updated both circe and Finch, without confirming that the bug (https://github.com/scoverage/sbt-scoverage/issues/146) had been fixed. In fact I'd forgotten about the bug entirely.

This means that since 81d89e20bc8c814a95d4aef88ddd956899e6d822 Travis CI hasn't been running 2.10 builds for Finch. Which wouldn't necessarily be a big deal, but some of the new content type stuff doesn't work on 2.10—or rather it works, but requires explicit type annotations.

I still don't think this is a terribly big deal. Lots of projects (like Play) are already dropping 2.10 support altogether, and we could provide helper methods if maintaining first-class 2.10 support is a priority here.

I've switched back to sbt-scoverage and added the type annotations to make the tests compile on 2.10. Sorry about this—I should have remembered the problem and been more careful.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -142,7 +142,7 @@ lazy val core = project
 lazy val test = project
   .settings(moduleName := "finch-test")
   .settings(allSettings)
-  .settings(coverageExcludedPackages := "io\\.finch\\.test\\..*")
+  .settings(ScoverageKeys.coverageExcludedPackages := "io\\.finch\\.test\\..*")
   .settings(libraryDependencies ++= testDependencies)
   .dependsOn(core)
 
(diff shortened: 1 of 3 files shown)


vkostyukov:
Thanks @travisbrown! Please, don't worry about that! I totally don't mind those extra types in the `toResponse` call. Especially that it's mostly for internal needs.

Thanks again for taking a time and fixing that!

PR #573, opened 2016-03-17
No ToService

vkostyukov:
I was experimenting with multiple-content-types support per Endpoint (coproduct), but stuck at the very last step. My idea was to allow passing coproduct of content-types into a `.toServiceAs[C]` call where each part of coproduct content-type corresponds to a part of Endpoint's coproduct.

I almost got that working, but stuck with a last case so I decided to open a PR half-way through - everything that I simplified while working on it.

I had something like this in mind. Please, let me know what do you think about that.

```scala
scala> val e = (foo :+: text).toServiceAs[JSON :+: JSON :+: CNil]
e: com.twitter.finagle.Service[com.twitter.finagle.http.Request,com.twitter.finagle.http.Response] = <function1>

scala> val e = (foo :+: text).toServiceAs[JSON :+: CNil]
<console>:43: error: could not find implicit value for parameter tro: io.finch.internal.ToResponse.Aux[io.finch.Output[shapeless.:+:[Foo,shapeless.:+:[String,shapeless.CNil]]],shapeless.:+:[JSON,shapeless.CNil]]
       val e = (foo :+: text).toServiceAs[JSON :+: CNil]
                                         ^
scala> val e = (foo :+: text).toServiceAs[TEXT :+: JSON]
<console>:45: error: could not find implicit value for parameter tro: io.finch.internal.ToResponse.Aux[io.finch.Output[shapeless.:+:[Foo,shapeless.:+:[String,shapeless.CNil]]],shapeless.:+:[TEXT,JSON]]
       val e = (foo :+: text).toServiceAs[TEXT :+: JSON]
                                         ^

scala> val e = (foo :+: text).toServiceAs[TEXT :+: JSON :+: CNil]
<console>:45: error: could not find implicit value for parameter tro: io.finch.internal.ToResponse.Aux[io.finch.Output[shapeless.:+:[Foo,shapeless.:+:[String,shapeless.CNil]]],shapeless.:+:[TEXT,shapeless.:+:[JSON,shapeless.CNil]]]
       val e = (foo :+: text).toServiceAs[TEXT :+: JSON :+: CNil]
                                         ^
scala> val e = (foo :+: text).toServiceAs[JSON :+: TEXT :+: CNil]
e: com.twitter.finagle.Service[com.twitter.finagle.http.Request,com.twitter.finagle.http.Response] = <function1>
```


ImLiar:
I still have some scepsis around this idea in terms of user and business applicability.

`.asService` usually lies on the different from `Endpoint`s layer. Almost all the time you convert all of your endpoints (their composition) into `.toService` and provide resulting `Service[Request, Response]` as an argument for twitter Http server while bootstrapping App. Let's imagine a case when you have dozens of endpoints. In this situation it would be a joke.

I believe that `Endpoint` should be responsible for `Content-Type` settings since it is much closer to business logic rather than utility one.

vkostyukov:
@ImLiar Yeah, I totally agree that this might be painful to keep track while dealing with quite a number of endpoints. The easies way (and probably most straightforward) to do content-types in Finch is to say that each endpoint should have one. We'd end up having like `Endpoint.Json`, `Endpoint.Xml` as type aliases to `Endpoint.Aux[A, CT]`. This implies some syntactic overhead, but it should work given that a coproduct endpoint will capture both types (two coproducts) so we can destructure it and resolve appropriate encoders when converting to a service.

Content-type negotiation fits pretty nicely here. We could say something like

```scala
val e : Endpoint.Aux[JSON :: XML :: HNil, User] = get(...)
```

This may look a bit scary for new users, but this is also couples endpoints with content-types (which I'm not sure is good or bad), but that would be pretty easy to implement.

vkostyukov:
Thinking about this more - I might want to take my words back. Not sure it will be _that easy_ to implement. I guess there is a middle ground to that. Instead of coupling content type to endpoint or specifying it at the very last step, let's attach it to the endpoint at the previous step.

There is usually a place in the code that slaps all the endpoints together with `:+:`. It might look as
```scala
val api = (getUser :+: postUser :+: healthcheck).toService
```
I think we can use that stage and introduce something like a thin wrapper to endpoint that also captures a content type (either one or `HList`).

```scala
val api = (
  getUser.in[JSON :: XML :: HNil] :+: 
  postUser.in[JSON] :+: 
  healthcheck.in[TEXT]
).toService
```

I guess we can also group endpoints of the same content-type together.

vkostyukov:
Okay. I think I broke something here (w.r.t. compile time):

Before:

````
[tw-mbp-vkostyukov finch (master)]$ time (sbt examples/clean && sbt ++2.11.8 examples/compile)
[success] Total time: 23 s, completed Mar 18, 2016 3:31:42 PM

real	0m43.734s
user	1m55.959s
sys	0m4.056s
````

After:

```
tw-mbp-vkostyukov finch (vk/multiple-content-types)]$ time (sbt examples/clean && sbt ++2.11.8 examples/compile)
[success] Total time: 96 s, completed Mar 18, 2016 3:28:42 PM

real	1m54.713s
user	3m55.831s
sys	0m6.440s
```

It looks like it takes 4-5x times more to compile the same code.

**Comment by vkostyukov** [comment]
Compile time is back to normal:

```scala
[tw-mbp-vkostyukov finch (vk/multiple-content-types)]$ time (sbt examples/clean && sbt ++2.11.8 examples/compile)
[success] Total time: 22 s, completed Mar 20, 2016 8:06:33 PM
```

PR #576, opened 2016-03-21
Fix a markdown error (missing backtick) in json docs.

peterneyens:
The [json docs](https://github.com/finagle/finch/blob/master/docs/json.md) is missing a backtick.


vkostyukov:
Thanks! Merging this.

Issue #574, opened 2016-03-18
FromParams should respect optional values


ImLiar:
As far as I understand, following line:
https://github.com/finagle/finch/blob/master/core/src/main/scala/io/finch/internal/FromParams.scala#L29

isn't tolerate to missing fields in `Request` even if it's optional in target object. I'm not sure how to solve this problem since we're using `LabelledGeneric` in the end and it requires clear evidence of types equality between `Repr` (in our case it's a record) and `A`'s fields, so it's not possible to simply use `paramOption` but I believe it should be marked as bug.

ImLiar:
Probably it's better to introduce `shapeless.Poly` with cases where `HV` equal to `Option[A]` and just some `A`. So we could use `paramOption` in first case and `param` in other one.

Summoning one of the shapeless gods @travisbrown

ImLiar:
In the end I've came to following solution:
https://gist.github.com/ImLiar/eeb80eda1c26b623d244

It's very straightforward but seems to work as expected. If it's okay I could submit PR

ImLiar:
I've found that `Seq` (params) won't work out of the box too, so I've updated gist.

teodimoff:
this is cool.. but cuple of nits: we need to make it for Option[Seq[A] too but in  general we need to do it with CanBuildFrom( or do we?) to make it work with collections more generally

ImLiar:
@teodimoff Thanks

Yea, I thought about it. I guess its possible to implement such case. Nice catch with `CanBuildFrom`!

teodimoff:
i am playing with it i post what i came with in  10-15 min

teodimoff:
for some reason it does not convert it to C[V]
```scala
         implicit def canBuildFromExtract[V,C[_]](implicit
   dh: DecodeRequest[V],
   ct: ClassTag[V],
   cbf: CanBuildFrom[Seq[V], V, C[V]]
  ): Case.Aux[String,Endpoint[C[V]]] = at[String]{key =>
    params(key).as(dh,ct).map(seqV => cbf(seqV).result())
   }

     implicit def canBuildFromExtractOption[V,C[_]](implicit
   dh: DecodeRequest[V],
   ct: ClassTag[V],
   cbf: CanBuildFrom[Seq[V], V, C[V]]
  ): Case.Aux[String,Endpoint[Option[C[V]]]] = at[String]{key =>
    params(key).as(dh,ct).map(seqV => cbf(seqV).result()).lift
   }
```

[...]

PR #577, opened 2016-03-24
FromParams should be tolerant to optional and seq fields

ImLiar:
#574 

I didn't succeed to make it work for all collections using `CanBuildFrom` due to some strange behaviour of implicit resolution and HKT. Tried for a few days, but still failed. Maybe I've missed something, but it just doesn't work for type `T[_]` even if there is an implicit instance of `cbf` in scope.

Anyway, I believe it would be useful even in it's current state. 

If someone could make it in more generic way, it would be super cool.

### DIFF
--- a/core/src/main/scala/io/finch/internal/FromParams.scala
+++ b/core/src/main/scala/io/finch/internal/FromParams.scala
@@ -3,8 +3,9 @@ package io.finch.internal
 import scala.reflect.ClassTag
 
 import io.finch._
-import shapeless.{::, HList, HNil, Witness}
+import shapeless._
 import shapeless.labelled._
+import shapeless.poly._
 
 /**
  * A type class empowering a generic derivation of [[Endpoint]]s from query string params.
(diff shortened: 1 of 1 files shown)


core/src/main/scala/io/finch/internal/FromParams.scala
```diff
@@ -13,20 +13,44 @@ trait FromParams[L <: HList] {
   def endpoint: Endpoint[L]
 }
 
-object FromParams {
+private[internal] object Extractor extends Poly1 {
```
vkostyukov: Can we move that under `FromParams` companion object?

core/src/main/scala/io/finch/internal/FromParams.scala
```diff
@@ -13,20 +13,44 @@ trait FromParams[L <: HList] {
...
+
+  implicit def extractor[V](implicit
+    dh: Decode[V],
+    ct: ClassTag[V]
+  ): Case.Aux[String, Endpoint[V]] = at[String] { key =>
+    param(key).as(dh, ct)
+  }
+}
+
+object FromParams extends Cases {
```
vkostyukov: What's `Cases`?

core/src/main/scala/io/finch/internal/FromParams.scala
ImLiar: Yea, ok

core/src/main/scala/io/finch/internal/FromParams.scala
ImLiar: It's shapeless's trait which contains `CaseN` types (49 line)

teodimoff:
@ImLiar can you look at updated gist https://gist.github.com/teodimoff/2d429d2e3068a6fcc293 . i have a hunch that params funcion will allow empty params to be Seq(). and we can make it work for at least Option[Seq[V]].

vkostyukov:
@teodimoff @ImLiar, why not always use `params` for `Seq[A]`? Seems like a reasonable default for me. `Option[Seq[A]]` doesn't make much sense for me, since `None: Option` might be viewed as `Nil: Seq`.

teodimoff:
@vkostyukov  my intuition is that if i declare case class Test(a: String,b: Int, seq: Seq[Int]) i would expect seq has at least one element but thats me. @ImLiar solution uses params which can be Nil which is perfectly fine too.

ImLiar:
@vkostyukov @teodimoff 
`Option[Seq[_]]` seems redundant to me.

I think we could change returning type of `paramsNonEmpty` to cats `NonEmptyList` and make a case for it. It also would be correct in terms of types.

vkostyukov:
Oh @ImLiar, you're reading my mind! Since we depend on Cats I think it would be fine to have `paramsNonEmpty` returning `NonEmptyList`. But I think it should be the part of separate discussion apart from this PR.

ImLiar:
@vkostyukov Okay, let's merge this PR first and then I'll make NEL and one more case for `Extractor`

vkostyukov:
Merging this. Thanks @ImLiar!

PR #582, opened 2016-03-30
paramsNonEmpty returns NonEmptyList[A]

ImLiar:
As it was discussed earlier: https://github.com/finagle/finch/pull/577#issuecomment-202103686

### DIFF
--- a/core/src/main/scala/io/finch/Endpoints.scala
+++ b/core/src/main/scala/io/finch/Endpoints.scala
@@ -2,6 +2,7 @@ package io.finch
 
 import java.util.UUID
 
+import cats.data.NonEmptyList
 import com.twitter.concurrent.AsyncStream
 import com.twitter.finagle.http.{Cookie, Method, Request}
 import com.twitter.finagle.http.exp.Multipart.FileUpload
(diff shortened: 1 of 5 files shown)


core/src/main/scala/io/finch/Endpoints.scala
```diff
@@ -337,11 +338,13 @@ trait Endpoints {
    * at least one element) multi-value query-string param `name` from the request into a `Seq` or
    * raises a [[Error.NotPresent]] exception when the params are missing or empty.
    */
-  def paramsNonEmpty(name: String): Endpoint[Seq[String]] =
+  def paramsNonEmpty(name: String): Endpoint[NonEmptyList[String]] =
```
vkostyukov: How do you feel about introducing a new function `paramsNel` and keeping the old one (behind deprecation)? I want to keep the API unchanged and I think `paramsNel` sounds like a reasonable name.

ImLiar: Okay, let's introduce new method

core/src/main/scala/io/finch/Endpoints.scala
```diff
@@ -357,6 +359,19 @@ trait Endpoints {
     }).shouldNot("be empty")(_.isEmpty)
 
   /**
+   * An evaluating [[Endpoint]] that reads a required multi-value query-string param `name`
+   * from the request into a `NonEmptyList` or raises a [[Error.NotPresent]] exception
+   * when the params are missing or empty.
+   */
+  def paramsNel(name: String): Endpoint[NonEmptyList[String]] =
+    option(items.ParamItem(name))(requestParams(name)).mapAsync({
+      case values if values.exists(_.nonEmpty) =>
```
vkostyukov: Can we simplify that and perform just a single traversal? Something like:

```scala
.mapAsync { values => values.filter(_.nonEmpty) match {
  case hd :: tl => Future.value(NonEmptyList(hd, tl))
  case _ => Future.exception(Error.NotPresent(items.ParamItem(name)))
}}
```

ImLiar: Emh, but in that case it could contain empty parameters

ImLiar: Ah, I see.

examples/src/main/scala/io/finch/div/Main.scala
```diff
@@ -24,6 +23,10 @@ import shapeless._
  */
 object Main extends App {
 
+  (1,2,3) match {
```
vkostyukov: Some unwanted changes.

ImLiar: I should go sleep :(

vkostyukov:
Thanks @ImLiar! Merging this!

Issue #575, opened 2016-03-21
Support multiple content-types per service


vkostyukov:
The easies solution would be to introduce a wrapper around `Endpoint` that captures two types - its underlying type and a content type. Something like:

``` scala
val service = (
  getUsers.respond[Application.Json]
  getFoos.respond[Application.Xml]
).toService
```

I think we have to keep in mind that at some point we will introduce content-type member for decoder and I guess there should be a symmetric variant of `respond` for that.

``` scala
val user = body.as[User].accept[Application.Json]
```

vkostyukov:
My most recent idea is to try to attach content-type to `Endpoint`:

```scala
trait Endpoint[A] {
  type ContentType
}
```

But keep `Output` content-type-free. We can default to JSON for `body` and endpoints created with `Mapper`, but use text/plain for everything else (headers, params). This would allow us to reuse this type when resolving decoders. So basically having a single content-type on endpoint means we can use it in both places: when resolving encoders  and decoders and this is really nice.

When composing two endpoints, the content-type will turn into a coproduct so we can expect one-by-one mapping.

Given that JSON is going to be a default, we can keep the API unchanged but mark `Endpoint` trait deprecated in favour explicit `Endpoint.Json` or even `Endpoint.Aux`.

vkostyukov:
I wonder how people feel about the following syntax:

```scala
val getUsers = post[Application.Json]("users" :: body[Application.Json].as[User]) {
  Ok("users")
}
```

crispywalrus:
If I assume that ```post[Application.Json]``` could be ```:+:```ed  with ```post[Text.XML]``` and it could be attached to other extractors (such as ```body``` from your original sketch) then I like that. 👍

vkostyukov:
I'm throwing this into the 0.16 milestone. The idea is to ship something similar to #583 (as an experimental API) and see if gets some adoption.

PR #583, opened 2016-04-05
Introduce ServiceBuilder

vkostyukov:
This is an attempt to solve #575 and a last PR in 0.11.

The idea is to use new abstraction `ServiceBuilder` that maps an `HList` of endpoints (even coproduct endpoints) to an `HList` of content-types. When an actual conversion (`toService`) happens we _fold_ un underlying list of endpoints and apply them one-by-one to a given request until it matches. The `HNil` case behaves as 404 (which is very nice).

The high-level API is following:

``` scala
val service = ServiceBuilder()
  .respond[Application.Json](getUsers :+: postUsers :+: deleteUser)
  .respond[Text.Plain](healthcheck)
  .toService
```

Note that `toService` still works, but marked as deprecated.

This is probably not the most efficient and clean solution, but I admire its simplicity and how small the diff itself. None of the `Endpoint`, `ToResponse`, `Encode` internals/API were touched - everything is done by new code.

Please, let me know how do you like the API. The only thing I'm worried about right now is whether or not it's good enough for further steps in this direction (content-type negotiation).

### DIFF
--- a/core/src/main/scala/io/finch/ServiceBuilder.scala
+++ b/core/src/main/scala/io/finch/ServiceBuilder.scala
@@ -1,0 +1,34 @@
+package io.finch
+
+import com.twitter.finagle.Service
+import com.twitter.finagle.http.{Request, Response}
+import io.finch.internal.ToService
+import shapeless._
+
+/**
+ * Captures the `HList` of endpoints as well as the `HList` of their content-types in order to do
+ * the `toService` conversion.
+ *
+ * {{{
+ *
+ * val api: Service[Request, Response] = ServiceBuilder()
+ *   .respond[Application.Json](getUser :+: postUser)
+ *   .respond[Text.Plain](healthcheck)
+ *   .toService
+ * }}}
+ */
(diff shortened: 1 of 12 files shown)


core/src/main/scala/io/finch/ServiceBuilder.scala
```diff
@@ -0,0 +1,34 @@
...
+      self.copy(e :: self.endpoints)
+  }
+
+  def respond[CT <: String]: Respond[CT] = new Respond[CT]
+
+  def toService(implicit ts: ToService[ES, CTS]): Service[Request, Response] = ts(endpoints)
+}
+
+object ServiceBuilder {
+  def apply(): ServiceBuilder[HNil, HNil] = ServiceBuilder(HNil)
```
travisbrown: This is nitpicky, but I think I'd prefer either `empty` or just a `respond[CT <: String]` method here that would give you a `ServiceBuilder(HNil).respond[CT]` (just because `Foo()` screams "side effects" to me and is kind of a workaround for syntactic limitations in Scala anyway). I'd also be happy to keep it as is, though.

vkostyukov:
See #794 instead.

PR #584, opened 2016-04-06
Earnest uses Finch ;)

jenscobie:


vkostyukov:
Very glad to hear that @jenscobie! Thanks for sharing!

Any chance you can share just a little of details (like whether or not you use Circe or maybe finagle-oauth2)?

PR #586, opened 2016-04-09
Update circe to second 0.4.0 release candidate

travisbrown:
I'm planning to have circe 0.4.0 out by Monday, and I've tested this locally, but I'm updating to RC2 just in case other people want to try it out with Finch as well.


vkostyukov:
Thanks @travisbrown!

PR #588, opened 2016-04-11
io.finch.Error.NoValid => io.finch.Error.NotValid

jimschubert:
Docs listed incorrect error type (`io.finch.Error.NoValid`). Also fixed 'should has' typo.


vkostyukov:
Thanks a lot @jimschubert!

PR #589, opened 2016-04-15
Update circe to 0.4.1 and avoid deprecated methods in exmaple

travisbrown:
Let's just pretend 0.4.0 didn't happen. :smile:


vkostyukov:
Thanks @travisbrown!

PR #590, opened 2016-04-17
derive instances for nested structures

teodimoff:
We will able to derive instances for 

```
case class Test(name: String, age: Int , nested: Nested)
case class Nested(a: String, b: Int)

val test = derive[Test].fromParams
```

### DIFF
--- a/core/src/main/scala/io/finch/internal/FromParams.scala
+++ b/core/src/main/scala/io/finch/internal/FromParams.scala
@@ -4,10 +4,10 @@ import scala.reflect.ClassTag
 
 import cats.data.NonEmptyList
 import io.finch._
+import io.finch.Endpoint._
 import shapeless._
 import shapeless.labelled._
 import shapeless.poly._
-import io.finch.Endpoint._
 
 /**
  * A type class empowering a generic derivation of [[Endpoint]]s from query string params.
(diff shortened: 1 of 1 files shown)


teodimoff:
If we want to make it work with Option we should introduce lifted variant for nestedExtractor

teodimoff:
sorry for this i did not know that pull request is tracking subsequent changes

ImLiar:
@teodimoff 

Nice work here1 But I can't imagine how this request with nested items will look like. Could you provide an example?

teodimoff:
@ImLiar it is possible that i have said something wrong. By nested i mean 
Test has nested case class Nested which fromParams can`t create right now.

ImLiar:
@teodimoff 

Nvm, I think, I understand how this works.

So we have `test` endpoint which will derive `Test` instance from some request like:
```
?name=John&age=45&a=str&b=0
```

Am I right?

teodimoff:
yes exactly. it will "convert" the Nested case class back to params and so on. i will add example to examples if this is necessary.

teodimoff:
need to fix couple of nits (like introducing Lazy in fromParams because the code is diverging in implicit resolution) and will add tests to make sure it compiles.

teodimoff:
i am having trouble to make this work. Endpoint implicit expansion is diverging somewhere and i cant find where. I will provide my working version (if that`s ok of course) later today so perhaps someone spot the difference.

teodimoff:
This version of FromParams is working https://github.com/teodimoff/rOut/blob/master/generic/src/io/rout/decode/FromParams.scala.
Any suggestions are welcome.

PR #592, opened 2016-05-10
Flexible Basic Auth

PChambino:
We are using finch at a project where we needed a more flexible BasicAuth implementation so we implemented one and I thought of contributing back.

The PR includes two big changes. The first, as per [rfc7235 section 4.1], is to always include a `WWW-Authenticate` header with a 401 Unauthorized response.
[rfc7235 section 4.1]: https://tools.ietf.org/html/rfc7235#section-4.1

The second is to replace the `BasicAuth` parameters by `authenticate: (String, String) => Boolean`. This way it is possible to reuse the same BasicAuth logic and authenticate the credentials by searching a database, reading from a file, or comparing to hard-coded credentials in the code.

Example:

```scala
val basicAuth: BasicAuth = BasicAuth("app-realm") { (user, pass) =>
  DB.query[Option[Boolean]](
    "SELECT true FROM users WHERE username = ? and password = ?", user, pass
  ).getOrElse(false)
}

// val basicAuth: BasicAuth = BasicAuth("user", "password") // This still works :)
// val basicAuth: BasicAuth = BasicAuth("app-realm", "user" -> "pass", "user2" -> "word") // This also works

val e: Endpoint[String] = basicAuth(Endpoint(Ok("secret place")))
```

In our case, and I believe in most cases, we started with a single hard-coded credential, then moved to a list of hard-coded credentials, and in the future we are planning to store credentials in a database.

Is this something you would like to include in `finch-core`?

### DIFF
--- a/core/src/main/scala/io/finch/Endpoints.scala
+++ b/core/src/main/scala/io/finch/Endpoints.scala
@@ -466,22 +466,44 @@ trait Endpoints {
   /**
    * Maintains Basic HTTP Auth for an arbitrary [[Endpoint]].
    */
-  case class BasicAuth(user: String, password: String) {
-    private[this] val userInfo = s"$user:$password"
-    private[this] val expected = "Basic " + Base64StringEncoder.encode(userInfo.getBytes)
-
+  case class BasicAuth(realm: String)(authenticate: (String, String) => Future[Boolean]) {
     def apply[A](e: Endpoint[A]): Endpoint[A] = new Endpoint[A] {
-      private[this] val failedOutput = new Rerunnable[Output[A]] {
-        override def run = Future.value(Unauthorized(BasicAuthFailed))
+      private[this] val unauthorized = new Rerunnable[Output[A]] {
+        override def run = Future.value(Unauthorized(BasicAuthFailed)
+          .withHeader("WWW-Authenticate" -> s"""Basic realm="$realm""""))
       }
 
       def apply(input: Input): Endpoint.Result[A] =
-        input.request.authorization.flatMap {
-          case `expected` => e(input)
-          case _ => Some(input.copy(path = Seq.empty) -> failedOutput)
+        e(input).map { case (input, output) =>
+          input -> authenticated(input).flatMap(if (_) output else unauthorized)
         }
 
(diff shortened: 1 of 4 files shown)


core/src/main/scala/io/finch/Endpoints.scala
```diff
@@ -466,24 +466,52 @@ trait Endpoints {
   /**
    * Maintains Basic HTTP Auth for an arbitrary [[Endpoint]].
    */
-  case class BasicAuth(user: String, password: String) {
-    private[this] val userInfo = s"$user:$password"
-    private[this] val expected = "Basic " + Base64StringEncoder.encode(userInfo.getBytes)
-
+  case class BasicAuth(realm: String)(authenticate: (String, String) => Boolean) {
```
vkostyukov: I wonder if we need to support `(String, String) => Future[Boolean]` instead so we could, for example, call the database asynchronously.

core/src/main/scala/io/finch/Endpoints.scala
```diff
@@ -466,24 +466,52 @@ trait Endpoints {
...
+          case Array(username, password) => Some((username, password))
+          case _ => None
         }
 
-      override def toString: String = s"BasicAuth($e)"
+      override def toString: String = s"""BasicAuth(realm="$realm", $e)"""
     }
   }
 
+  object BasicAuth {
```
vkostyukov: I'm not 100% sure we need to provide these in `finch-core`. What do you think about just supporting `(String, String) =? Future[Boolean]` and let users to do the rest?

vkostyukov:
This is really interesting! Thanks a lot @PChambino! Just a couple of comments from me. Also, please don't forget to squash commits.

core/src/main/scala/io/finch/Endpoints.scala
PChambino: I was also unsure if `finch-core` should provide this or not. Since they are so simple I think it is reasonable to let users implement what they need. I'll remove them.

vkostyukov:
This is really nice. Thanks a ton @PChambino!

PR #594, opened 2016-05-19
Updated the main README.md file to inidcate that the finch-playjson d…

rampart81:
…ependency will be available for version 0.11.0


vkostyukov:
Sorry for the delay, @rampart81. I think this is good to go. Thanks!

PR #595, opened 2016-05-25
update cats, circe and catbird dependencies

crispywalrus:
Also updated without issue were finagle and shapeless. the only actual issue encountered was the change in cats dependency on algebra. the net effect was to move Eq from algebra to cats.kernel

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -34,7 +34,7 @@ val testDependencies = Seq(
   "org.scalacheck" %% "scalacheck" % "1.12.5",
   "org.scalatest" %% "scalatest" % "2.2.6",
   "org.typelevel" %% "cats-laws" % catsVersion,
-  "org.typelevel" %% "discipline" % "0.5"
+  "org.typelevel" %% "discipline" % "0.4"
 )
 
 val baseSettings = Seq(
--- a/build.sbt
+++ b/build.sbt
@@ -8,11 +8,11 @@ lazy val buildSettings = Seq(
   crossScalaVersions := Seq("2.10.6", "2.11.8")
 )
 
-lazy val finagleVersion = "6.34.0"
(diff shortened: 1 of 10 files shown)


ilya-murzinov:
```bash
[error] /home/travis/build/finagle/finch/core/src/test/scala/io/finch/FinchSpec.scala:158: could not find implicit value for parameter a: org.scalacheck.Arbitrary[io.finch.Input => A]
[error]       Arbitrary.arbitrary[Input => A].map { f =>
[error]                          ^
[error] one error found
```
@crispywalrus do you know what could cause this?

travisbrown:
I don't remember all the details of the situation, but Discipline 0.5 bumps the ScalaCheck version from 1.12 to 1.13, and there are still some questions about function instances in 1.13. The current versions of Cats and of all of the projects I maintain are still on Discipline 0.4, and I think that'd fix things here as well.

travisbrown:
Otherwise :+1: from me—thanks for doing this, @crispywalrus.

My one other question is about the `Eq` imports—specifically whether you think we should be using the `cats.Eq` alias. Before Cats 0.6 I stuck with the `algebra.Eq` because I don't particularly like the idea of aliases in the API, but that meant I had to make a lot of changes after the 0.6 arrangement, and I'm now using `cats.Eq`. I could go either way, though—I don't think it's a big deal.

crispywalrus:
As soon as I saw your question @travisbrown  about Eq I realized what I had felt was wrong about the change I made. It seems better to rely on cats.$stuff rather than cats.kernel just for clearness. I've rolled back the discipline version although I'm not seeing the reported failures.

vkostyukov:
Thanks @crispywalrus! LGTM!

PR #598, opened 2016-06-04
Add params example and clarify argument type

chbrown:
Since you have to provide types for each item in the argument tuple of the http verb method (e.g., `get`, `post`), and the error message you get when you try to use, e.g., `ids: List[String]` along with `:: params("ids")` takes a bit of scrutiny to figure out what it wants, I thought an example with the proper type would be helpful to newcomers like me.

Side question: I initially thought "multi-value param" would simply mean 'all the query parameters with that name will be aggregated and their values supplied in a single `Seq[String]`'. But it turns out finch also splits `params`-style query parameters on commas, and filters out empty strings. Which is handy in my current app, but more magical than I would have expected, and I'm not sure how I could get the pre-processed values if I wanted. Empty strings can be significant; I know of some client-side routing libraries (e.g., Angular's ui-router) that don't serialize query parameters with `null` or `undefined` values into the URL at all, but _do_ serialize `""` values as, well, empty strings. Anyways, the question was, is this behavior documented somewhere else, like maybe upstream in Finagle?

### DIFF
--- a/docs/endpoint.md
+++ b/docs/endpoint.md
@@ -169,7 +169,7 @@ Finch provides the following instances for reading HTTP params (evaluating endpo
 
 - `param("foo")` - required param "foo"
 - `paramOption("foo")` - optional param "foo"
-- `params("foos")` - multi-value param "foo" that might return an empty list
+- `params("foos")` - multi-value param "foo" that might return an empty sequence
 - `paramsNonEmpty("foos")` - multi-value param "foo" that fails when empty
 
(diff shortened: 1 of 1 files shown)


vkostyukov:
Thanks a lot @chbrown!

I just double checked the scaladoc for `params` and it seems we don't mention this behaviour there. I'm happy to fix that in the upcoming PR, but let me know if you want to take a stub on that as well.

Also, I think that's a valid concern that we don't always have to _filter_ empty strings. Maybe provide a new endpoint for that purpose? I filed #600. Thanks again!

PR #601, opened 2016-06-07
Add some community based finch templates to the best practices

tomjadams:


vkostyukov:
Thanks a lot @tomjadams!

PR #603, opened 2016-06-09
Add async authentication example to Cookbook

rf-:
The existing authentication example is helpful, but it's very common for auth to require external calls, so it seems worth including an example in the Cookbook.


vkostyukov:
This is really nice! Thanks a lot, Ryan @rf-!

PR #605, opened 2016-06-14
Fix the different typos in the doc

BenFradet:
I'm currently experimenting with Finch (having a great time btw, so thanks for that).
So, naturally, I went through the doc (which are also great :smile: ) and found a few typos.
This PR aims to fix those I found.


vkostyukov:
This looks fantastic! Thanks a ton @BenFradet!

BenFradet:
Glad I could help.

PR #607, opened 2016-06-28
Added typed services using finch talk.

tomjadams:


vkostyukov:
Thanks a lot @tomjadams!

PR #608, opened 2016-07-09
Update Finagle to 6.36.0 and use new catbird release

travisbrown:
I'm not sure when we're planning for this update to happen, but I just published a [new catbird.io release](https://github.com/travisbrown/catbird/releases/tag/v0.6.0) that depends on Finagle 6.36.0.


crispywalrus:
:+1: on pushing this into the 0.11 milestone

vkostyukov:
Thanks, Travis! I was hoping to get 0.11 released for both 2.10 and 2.11. That means staying on 6.35 for a while. Although, I'm not 100% sold on this idea yet so maybe we can (should?) drop 2.10 support in 0.11.

tomjadams:
FWIW I'm +1 on 2.11 only.

dave-bemiller-ck:
+1 on 2.11 only here, too

KadekM:
+1 on 2.11 only here

travisbrown:
Superseded by #629.

PR #609, opened 2016-07-10
Replace FlatSpec by Suite in ServiceSuite

BenFradet:
This PR replaces the self type annotation `FlatSpec` by `Suite` in `ServiceSuite` and `ServiceIntegrationSuite` so those can be mixed in with any scalatest specs (`WordSpec`, `FlatSpec`, etc).


vkostyukov:
Thanks @BenFradet! Merging this in - should be released as part of 0.11-M1.

BenFradet:
As a follow-up I was thinking we should add tests using those suites to the examples so people can refer to those to test their apps, what do you think?

BenFradet:
ping @vkostyukov , what do you think about ^?

PR #616, opened 2016-07-15
Fix priority of AnsyncStream encoders

vkostyukov:
This was preventing `toService` from compiling successfully.


vkostyukov:
I think I'm going to add some tests here.

PR #618, opened 2016-07-18
Fix documentation example

KadekM:
I believe it should be `Text.Plain`  insted of `Plain.Text`.


vkostyukov:
Thanks, @KadekM! Nice catch!

PR #619, opened 2016-07-22
Endpoint testing

### LINKED ISSUE #390: Testing Endpoints

BenFradet:
fixes #390

### DIFF
--- a/core/src/main/scala/io/finch/Input.scala
+++ b/core/src/main/scala/io/finch/Input.scala
@@ -1,7 +1,10 @@
 package io.finch
 
+import java.nio.charset.Charset
+
 import cats.Eq
-import com.twitter.finagle.http.Request
+import com.twitter.finagle.http.{Method, Request, RequestBuilder}
+import com.twitter.io.Buf
 
 /**
  * An input for [[Endpoint]].
@@ -10,6 +13,32 @@ final case class Input(request: Request, path: Seq[String]) {
   def headOption: Option[String] = path.headOption
   def drop(n: Int): Input = copy(path = path.drop(n))
   def isEmpty: Boolean = path.isEmpty
+
+  def withBody(buf: Buf, charset: Option[Charset] = None): Input = {
+    request.content = buf
+    request.contentLength = buf.length.toLong
+    charset.foreach(cs => request.charset = charset.get.displayName())
+    this
+  }
+
+  def withHeaders(headers: (String, String)*): Input = {
+    headers.foreach { case (k, v) => request.headerMap.set(k, v) }
+    this
+  }
+
+  def withForm(params: (String, String)*): Input = {
+    require(request.host.isDefined, "The host has to be defined")
+    require(params.nonEmpty, "Cannot create request without params")
+    val req = RequestBuilder()
+      .addFormElement(params: _*)
+      .url(request.host.get)
+      .buildFormPost()
+    request.method = req.method
+    request.content = req.content
+    req.contentLength.foreach(cl => request.contentLength = req.contentLength.get)
+    req.charset.foreach(cs => request.charset = req.charset.get)
+    this
+  }
 }
 
 /**
@@ -19,4 +48,22 @@ object Input {
   def apply(req: Request): Input = Input(req, req.path.split("/").toList.drop(1))
 
   implicit val inputEq: Eq[Input] = Eq.fromUniversalEquals
+
+  def get(path: String, params: (String, String)*): Input = Input(Request(path, params: _*))
+  def put(path: String, params: (String, String)*): Input = {
+    val req = Request(path, params: _*)
+    req.method = Method.Put
+    Input(req)
+  }
+  def patch(path: String, params: (String, String)*): Input = {
+    val req = Request(path, params: _*)
+    req.method = Method.Patch
+    Input(req)
+  }
+  def delete(path: String, params: (String, String)*): Input = {
+    val req = Request(path, params: _*)
+    req.method = Method.Patch
+    Input(req)
+  }
+  def post(path: String): Input = Input(Request(Method.Post, path))
 }
--- a/core/src/test/scala/io/finch/FinchSpec.scala
+++ b/core/src/test/scala/io/finch/FinchSpec.scala
@@ -225,4 +225,5 @@ trait FinchSpec extends FlatSpec with Matchers with Checkers with AllInstances
     Arbitrary(genPayloadOutput[A])
 
   implicit def arbitraryOutput[A: Arbitrary]: Arbitrary[Output[A]] = Arbitrary(genOutput[A])
+
 }
--- a/core/src/main/scala/io/finch/Output.scala
+++ b/core/src/main/scala/io/finch/Output.scala
@@ -168,10 +168,10 @@ object Output {
   }
 
   implicit class EndpointResultOps[A](val o: Endpoint.Result[A]) extends AnyVal {
-    private[finch] def poll: Option[Try[A]] = o.flatMap(_._2.run.poll.map(_.map(_.value)))
-    private[finch] def output: Option[Output[A]] = o.map({ case (_, oa) => Await.result(oa.run) })
-    private[finch] def value: Option[A] = output.map(oa => oa.value)
-    private[finch] def remainder: Option[Input] = o.map(_._1)
+    def poll: Option[Try[A]] = o.flatMap(_._2.run.poll.map(_.map(_.value)))
+    def output: Option[Output[A]] = o.map({ case (_, oa) => Await.result(oa.run) })
+    def value: Option[A] = output.map(oa => oa.value)
+    def remainder: Option[Input] = o.map(_._1)
   }
 
   implicit class OutputOps[A](val o: Output[A]) extends AnyVal {
--- a/core/src/test/scala/io/finch/InputSpec.scala
+++ b/core/src/test/scala/io/finch/InputSpec.scala
@@ -1,0 +1,63 @@
+package io.finch
+
+import java.nio.charset.Charset
+
+import com.twitter.finagle.http.Method
+import com.twitter.io.Buf
+import org.scalacheck.Prop
+import org.scalacheck.Prop.forAll
+
+class InputSpec extends FinchSpec {
+
+  behavior of "Input"
+
+  it should "properly construct Inputs using factories with params for the different methods" in {
+    def validateInput(
+        input: Input, method: Method, segments: Seq[String], params: Map[String, String]): Prop = {
+      input.request.method === method
+      input.request.path === (segments mkString "/")
+      input.request.params === params
+      input.path === segments
+    }
+    check { ps: Params =>
+      forAll(genPath) { p: String =>
+        val segments = p.split("/").toList.drop(1)
+        validateInput(Input.get(p, ps.p.toSeq: _*), Method.Get, segments, ps.p)
+        validateInput(Input.put(p, ps.p.toSeq: _*), Method.Put, segments, ps.p)
+        validateInput(Input.patch(p, ps.p.toSeq: _*), Method.Patch, segments, ps.p)
+        validateInput(Input.delete(p, ps.p.toSeq: _*), Method.Delete, segments, ps.p)
+      }
+    }
+  }
+
+  it should "add content through withContent" in {
+    check { (i: Input, b: Buf, cs: Charset) =>
+      i.withBody(b).request.content === b
+      i.withBody(b, Some(cs)).request.content === b
+    }
+  }
+
+  it should "add headers through withHeaders" in {
+    check { (i: Input, hs: Headers) =>
+      val hm = i.withHeaders(hs.m.toSeq: _*).request.headerMap
+      hs.m.forall { case (k, v) =>
+        hm.contains(k)
+        hm(k) === v
+      }
+    }
+  }
+
+  it should "add form elements through withForm" in {
+    check { (i: Input, ps: Params) =>
+      intercept[IllegalArgumentException](i.withForm(ps.p.toSeq: _*))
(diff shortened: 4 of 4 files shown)


core/src/main/scala/io/finch/Input.scala
```diff
@@ -10,6 +11,21 @@ final case class Input(request: Request, path: Seq[String]) {
   def headOption: Option[String] = path.headOption
   def drop(n: Int): Input = copy(path = path.drop(n))
   def isEmpty: Boolean = path.isEmpty
+
+  def withContent(s: String): Input = {
+    val req = request
+    req.contentString = s
+    req.contentLength = s.length.toLong
+    req.charset = "utf-8"
```
BenFradet: should we add `charset` as a param to `withContent`?

BenFradet:
I haven't yet found a clean way to do a `withHeaders`.

Regarding `withParams`, shouldn't the user just submit a new `Request` containing those params?

core/src/main/scala/io/finch/Input.scala
```diff
@@ -10,6 +11,21 @@ final case class Input(request: Request, path: Seq[String]) {
   def headOption: Option[String] = path.headOption
   def drop(n: Int): Input = copy(path = path.drop(n))
   def isEmpty: Boolean = path.isEmpty
+
+  def withContent(s: String): Input = {
```
BenFradet: Should we go with `withContent` or `withBody`?

core/src/main/scala/io/finch/Input.scala
```diff
@@ -19,4 +35,14 @@ object Input {
...
+
+  def get(path: String): Input = Input(Request(Method.Get, path))
+  def post(path: String): Input = Input(Request(Method.Post, path))
+  def put(path: String): Input = Input(Request(Method.Put, path))
+  def patch(path: String): Input = Input(Request(Method.Patch, path))
+  def delete(path: String): Input = Input(Request(Method.Delete, path))
+  def connect(path: String): Input = Input(Request(Method.Connect, path))
+  def head(path: String): Input = Input(Request(Method.Head, path))
+  def options(path: String): Input = Input(Request(Method.Options, path))
+  def trace(path: String): Input = Input(Request(Method.Trace, path))
```
BenFradet: I don't know if we should support all methods or a restricted set, what do you think?

core/src/main/scala/io/finch/Input.scala
vkostyukov: Let's do `withBody`. You can also make it take `Buf` instead.

core/src/main/scala/io/finch/Input.scala
vkostyukov: Yeah, Let's make it `withBody(buf: Buf, cs: Option[Charset] = None)`.

core/src/main/scala/io/finch/Input.scala
vkostyukov: Let's keep them all, it's fine.

vkostyukov:
Looks pretty great, @BenFradet! Thanks a lot! Mind squashing the commits into one, once you're done with changes?

BenFradet:
@vkostyukov do you have opinions/ideas regarding `withHeaders` and `withParams`?

vkostyukov:
I think `withParams(params: (Stirng, Stirng)*))` and `withHeaders(headers: (String, String)*)` should work fine here.

core/src/main/scala/io/finch/Input.scala
```diff
@@ -10,6 +13,14 @@ final case class Input(request: Request, path: Seq[String]) {
   def headOption: Option[String] = path.headOption
   def drop(n: Int): Input = copy(path = path.drop(n))
   def isEmpty: Boolean = path.isEmpty
+
+  def withBody(buf: Buf, charset: Option[Charset] = None): Input = {
+    val req = request
+    req.content = buf
+    req.contentLength = buf.length.toLong
+    if (charset.isDefined) req.charset = charset.get.displayName()
+    copy(request = req)
```
vkostyukov: Since request is mutable we can just return `this`

BenFradet:
@vkostyukov I implemented `withHeaders` but, after much fiddling around with trying to implement `withParams`, I haven't been able to. Do you have ideas on how to do so?

vkostyukov:
Ah right. I see the problem now. Let's think about that.

Looks like we can restrict to the following HTTP verbs for now: GET, PUT, PATCH, DELETE, POST. All of them, except for POST, usually are idempotent, which means users might want to use query string with them. POST is a special case that we should handle carefully and not motivate users to use odd practices/design decisions.

Here is what I think we should do.

* Let's support `Input.$method(path: String, params: (String, String)*)` for GET, PUT, PATCH, DELETE
* Let's support `Input.post(path: String)`
* Let's provide the following methods to specify the payload (on `Input`):
  * `withBody(body: Buf, charset: Option[Charset] = None)` - basic case
  * `withForm(params: (String, String)*)` - URL encoded params/form
  * `withMultipartForm(fileUpload: ???, params: (String, String)*)` - for multipart/form-data payloads

I think this covers enough cases and provides a reasonable API so users can easily test their endpoints. What do you think @BenFradet?

BenFradet:
Fine with me. I still don't see how to implement `withForm`/`withParams` since AFAICT you can't modify the params of an existing `Request`.

vkostyukov:
Yeah, but those go into a payload/body as `application/x-www-form-urlencoded` or `multipart/form-data`. You can take a look at Finagle's `RequestBuilder` on how it's implemented. I think we can either reuse `RequestBuilder` here or call N3 API directly. Ether way works for me.

BenFradet:
Ok, thanks, I'll investigate `RequestBuilder`.

core/src/main/scala/io/finch/Input.scala
```diff
@@ -10,6 +13,31 @@ final case class Input(request: Request, path: Seq[String]) {
...
+    if (params.nonEmpty) {
+      val req = RequestBuilder()
+        .addFormElement(params: _*)
+        .addHeaders(request.headerMap.toMap)
+        .url(request.host.get)
+        .buildFormPost()
+      copy(request = req)
+    } else {
+      this
+    }
```
BenFradet: I'm seeing a weird NPE from finagle's when `params` is an empty seq (cf [buildFormPost](https://github.com/twitter/finagle/blob/develop/finagle-http/src/main/scala/com/twitter/finagle/http/RequestBuilder.scala#L347-L383).

I guess netty's `HttpPostRequestEncoder` does not like having no body.

core/src/main/scala/io/finch/Input.scala
```diff
@@ -10,6 +13,31 @@ final case class Input(request: Request, path: Seq[String]) {
...
+    headers.foreach { case (k, v) => request.headerMap.set(k, v) }
+    this
+  }
+
+  def withForm(params: (String, String)*): Input = {
+    if (params.nonEmpty) {
+      val req = RequestBuilder()
+        .addFormElement(params: _*)
+        .addHeaders(request.headerMap.toMap)
+        .url(request.host.get)
```
BenFradet: How would you handle the case when host hasn't been set? Returning the same `Input`?

BenFradet:
@vkostyukov implemented what you advocated for, review welcome

core/src/main/scala/io/finch/Input.scala
```diff
@@ -10,6 +13,40 @@ final case class Input(request: Request, path: Seq[String]) {
...
+  def withMultipartForm(fileElements: FileElement*)(params: (String, String)*): Input = {
+    require(request.host.isDefined, "The host has to be defined")
+    require(fileElements.nonEmpty, "Cannot create request without file elements")
+    val req = RequestBuilder()
+      .addFormElement(params: _*)
+      .add(fileElements)
+      .url(request.host.get)
+      .buildFormPost(multipart = true)
+    copy(request = req)
+  }
```
BenFradet: I'm not sure `withForm` and `withMultipartForm` should be copy ctor, imo it'd be more appropriate to make them `post` factories, what do you think?

core/src/main/scala/io/finch/Input.scala
```diff
@@ -19,4 +56,26 @@ object Input {
   def apply(req: Request): Input = Input(req, req.path.split("/").toList.drop(1))
 
   implicit val inputEq: Eq[Input] = Eq.fromUniversalEquals
+
+  def get(path: String): Input = Input(Request(path))
```
vkostyukov: I don't think we need these methods given that `(String, String)*` defaults to empty list.

core/src/main/scala/io/finch/Input.scala
```diff
@@ -10,6 +13,40 @@ final case class Input(request: Request, path: Seq[String]) {
...
+
+  def withForm(params: (String, String)*): Input = {
+    require(request.host.isDefined, "The host has to be defined")
+    require(params.nonEmpty, "Cannot create request without params")
+    val req = RequestBuilder()
+      .addFormElement(params: _*)
+      .addHeaders(request.headerMap.toMap)
+      .url(request.host.get)
+      .buildFormPost()
+    copy(request = req)
```
vkostyukov: Let's just change the request's content, but keep the rest of the things. You can still use `RequestBuilder` to build the request and then copy its body over.

BenFradet: Will look into that, thx.

BenFradet: The problem with copying only the body of the request is that headers added by `buildFormPost` won't be preserved.

core/src/main/scala/io/finch/Input.scala
BenFradet: @vkostyukov do you any comment on this?

core/src/main/scala/io/finch/Input.scala
```diff
@@ -10,6 +13,48 @@ final case class Input(request: Request, path: Seq[String]) {
...
+      .url(request.host.get)
+      .buildFormPost()
+    request.method = req.method
+    request.content = req.content
+    if (req.contentLength.isDefined) request.contentLength = req.contentLength.get
+    if (req.charset.isDefined) request.charset = req.charset.get
+    this
+  }
+
+  def withMultipartForm(fileElements: FileElement*)(params: (String, String)*): Input = {
```
vkostyukov: How about using `c.t.f.http.exp.Multipart` here? I really don't want to use `FileElement` from the `RequestBuilder` API here, but we already use `Multipart.FileUpload` in Finch so we should be fine.

vkostyukov:
Thanks @BenFradet! This looks pretty great! Just one concern about `FileUpload` type.

core/src/main/scala/io/finch/Input.scala
BenFradet: I'm all for it, that will simplify things. I'll make the necessary changes later today.

BenFradet:
Still haven't figured out how to build a request from a multipart, any help would be appreciated.

core/src/main/scala/io/finch/Input.scala
```diff
@@ -10,6 +13,48 @@ final case class Input(request: Request, path: Seq[String]) {
   def headOption: Option[String] = path.headOption
   def drop(n: Int): Input = copy(path = path.drop(n))
   def isEmpty: Boolean = path.isEmpty
+
+  def withBody(buf: Buf, charset: Option[Charset] = None): Input = {
+    request.content = buf
+    request.contentLength = buf.length.toLong
+    if (charset.isDefined) request.charset = charset.get.displayName()
```
vkostyukov: You can do `charset.foreach(cs => request.charset = cs.displayName)`.

core/src/main/scala/io/finch/Input.scala
vkostyukov: I understand `Multipart` is super confusing and I'm not sure I have a good answer for it right now. How about we do everything else and leave multipart for the next attempt?

BenFradet: I'm ok with that, will remove that later today.

core/src/main/scala/io/finch/Input.scala
BenFradet: Thanks, will do.

core/src/test/scala/io/finch/FinchSpec.scala
```diff
@@ -130,6 +130,13 @@ trait FinchSpec extends FlatSpec with Matchers with Checkers with AllInstances
     )
   } yield b
 
+  def genFileElement: Gen[FileElement] = for {
```
vkostyukov: Do we still need those?

BenFradet: good point, forgot to remove them

BenFradet:
@vkostyukov good to go :+1:

vkostyukov:
Thanks @BenFradet! I'm going to merge that and wrap up a next release.

BenFradet:
great, looking forward to that!

PR #620, opened 2016-07-28
Add Sigma to Finch "Adopters" list in README.md

omerzach:
Sigma's entire API server is built with Finch


vkostyukov:
Thanks @omerzach! I'm really glad to see the adopter's list growing!

PR #625, opened 2016-08-14
Fix #600: Remove querystring params forced comma splitting

### LINKED ISSUE #600: Params endpoint that keeps empty string
As mentioned in #598 we need a simpler version of `params`:

1. Do not filter empty strings
2. Do not split values by `,`

chbrown:
Two commits:
* 590300f to establish previous behavior
* 0f92071 to fix it and test new behavior

I wasn't sure if there was any logic to the order of `it should ...`'s in `EndpointSpec.scala`, so I just put the new test near the other param/params test.


vkostyukov:
Perfect! Thanks a lot @chbrown!

PR #629, opened 2016-08-21
Update to Cats 0.7.0

travisbrown:
This (together with #595) supersedes #608, but since we don't have a catbird release for Cats 0.7.0 and Finagle 6.36.0, it'll need to wait until Finch is ready to drop 2.10 support.


crispywalrus:
So this looks 👍  but the travis definitions should be changed to drop the 2.10 and non-jdk8 builds as they'll always fail.

travisbrown:
Right, thanks @crispywalrus.

crispywalrus:
This change has led to a test failure in my project (which uses finch + circe + jackson)
```
SEVERE: Fatal exception propagated to the root monitor!
java.lang.NoClassDefFoundError: io/circe/jawn/package$
	at io.finch.circe.Decoders$$anonfun$decodeCirce$1.apply(Decoders.scala:14)
	at io.finch.circe.Decoders$$anonfun$decodeCirce$1.apply(Decoders.scala:13)
	at io.finch.LowPriorityDecodeInstances$$anon$1.apply(Decode.scala:21)
	at io.finch.Endpoint$StringEndpointOps$$anonfun$as$extension$3.apply(Endpoint.scala:447)
	at io.finch.Endpoint$StringEndpointOps$$anonfun$as$extension$3.apply(Endpoint.scala:447)
	at io.finch.Endpoint$$anon$3$$anonfun$apply$1$$anonfun$apply$2$$anonfun$apply$3.apply(Endpoint.scala:97)
	at io.finch.Endpoint$$anon$3$$anonfun$apply$1$$anonfun$apply$2$$anonfun$apply$3.apply(Endpoint.scala:97)
	at io.finch.Output$class.traverse(Output.scala:68)
	at io.finch.Output$Payload.traverse(Output.scala:142)
	at io.finch.Endpoint$$anon$3$$anonfun$apply$1$$anonfun$apply$2.apply(Endpoint.scala:97)
	at io.finch.Endpoint$$anon$3$$anonfun$apply$1$$anonfun$apply$2.apply(Endpoint.scala:97)
```
This code worked as expected (i.e. it used the jackson decoder/decode) before the upgrade (no class not found exception). It's kind of a kitchen sink wrt imports

```
import com.twitter.util.Await
import com.twitter.finagle.http._
import com.creditkarma.jarvis.model._
import com.creditkarma.jarvis.modules._
import com.creditkarma.jarvis.services._, modules._
import com.twitter.finagle.stats.NullStatsReceiver
import com.typesafe.config._
import io.circe.generic.decoding.DerivedDecoder._
import io.circe.jackson._
import io.finch.circe._
import org.scalatest._
```

travisbrown:
@crispywalrus If I remember correctly (or rather if this hasn't changed in the past few months), Finch only supports using Jackson for printing, not parsing.

The `NoClassDefFoundError` found is really odd, since finch-circe definitely has a circe-jawn dependency. You've tried cleaning?

vkostyukov:
Thanks @travisbrown! Merging that.

PR #632, opened 2016-08-27
Add circe JSON examples

NeQuissimus:
I was just looking into finch in combination with circe.
Since I created a small sample application, I thought it may make for a good addition to the examples in the repo.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -230,6 +230,7 @@ lazy val examples = project
       |io\.finch\.streaming\..*;
       |io\.finch\.oauth2\..*;
       |io\.finch\.wrk\..*;
+      |io\.finch\.circe\..*;
     """.stripMargin)
   .settings(
     libraryDependencies ++= Seq(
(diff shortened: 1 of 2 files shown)


vkostyukov:
@NeQuissimus Have you seen [oauth2](https://github.com/finagle/finch/tree/master/examples/src/main/scala/io/finch/oauth2) and [todo](https://github.com/finagle/finch/tree/master/examples/src/main/scala/io/finch/todo) examples? They are both based on Finch and Circe.

NeQuissimus:
I had not gone through the todo one. Do with this PR what you want really :D I figured since I did the work, why not send it in. You cannot really have too many examples...
If there is too much overlap, just close the PR :) I don't mind!

vkostyukov:
I think having a Todo example written with the most recent best-practices is quite sufficient (it covers a lot of use cases). Please feel free to open a new PR if you think a Todo example requires some improvements.

PR #639, opened 2016-09-04
Update Cats and Circe versions

travisbrown:
Circe 0.5.0 is identical to 0.5.0-M3, and Cats 0.7.2 is binary-compatible with 0.7.0, but updating lets us avoid exposing users to [this bug](https://github.com/typelevel/cats/issues/1346) in Cats 0.7.0.


vkostyukov:
Thanks @travisbrown!

PR #642, opened 2016-09-13
Fix #641: Treat empty params, headers, and bodies as Some("")

### LINKED ISSUE #641: Support empty strings in optional params and headers
When using `paramOption("key")` where the `key` param is present but has an empty string for its value Finch gives us `None` instead of `Some("")`.

I believe just removing the `noneIfEmpty` call solves the issue
```
  def paramOption(name: String): Endpoint[Option[String]] =
    option(items.ParamItem(name))(requestParam(name)).noneIfEmpty
```

Is there a reason we'd want to keep that? If not I'll submit a pull request to remove it for `paramOption` and `headerOption`. If there is a reason to keep the existing behavior is there some obvious workaround I'm missing or is it okay for me to submit a pull request that adds a new endpoint with my desired behavior?

omerzach:


core/src/main/scala/io/finch/Endpoints.scala
```diff
@@ -282,14 +282,15 @@ trait Endpoints {
 
   private[this] def requestBody(req: Request): Option[String] =
     req.contentLength match {
-      case Some(n) if n > 0 =>
+      case Some(0) => Some("")
```
vkostyukov: Let's preallocate `Some("")` since it's immutable.

core/src/main/scala/io/finch/Endpoints.scala
```diff
@@ -282,14 +282,15 @@ trait Endpoints {
...
+      case Some(n) =>
```
vkostyukov: We don't use `n` so feel free to `Some(_)`.

core/src/main/scala/io/finch/Endpoints.scala
```diff
@@ -403,8 +404,9 @@ trait Endpoints {
   val binaryBodyOption: Endpoint[Option[Array[Byte]]] =
     matches(items.BodyItem)(!_.isChunked)(req =>
       req.contentLength match {
-        case Some(n) if n > 0 => Some(Buf.ByteArray.Shared.extract(req.content))
-        case _ => None
+        case Some(0) => Some(Array.empty)
```
vkostyukov: Let's preallocate `Some(Array.empty[Byte])` since it's effectively immutable (there is nothing a user can do to change that array: length's is fixed and it's zero).

core/src/main/scala/io/finch/Endpoints.scala
```diff
@@ -403,8 +404,9 @@ trait Endpoints {
...
+        case Some(n) => Some(Buf.ByteArray.Shared.extract(req.content))
```
vkostyukov: We don't use `n` here so feel free to `Some(_)`.

vkostyukov:
Thanks @omerzach! This looks good!

Can we also remove `noneIfEmpty` from [here](https://github.com/finagle/finch/blob/82f2f4240805c1bc63f5ed16fb402e01764d32dc/core/src/main/scala/io/finch/Endpoint.scala#L549-L553)?

core/src/main/scala/io/finch/Endpoints.scala
omerzach: Sure. You mean I should add a
```
private val emptyBinaryBodyOption = Some(Array.empty[Byte])
```
above and just use that in the `Some(0)` case, right?

vkostyukov: Yeah, something like that works for me.

omerzach:
Hmm looks like I messed something up with the PR, looking into it

omerzach:
All set now, I think

core/src/main/scala/io/finch/Endpoints.scala
```diff
@@ -280,16 +280,18 @@ trait Endpoints {
   private[this] def requestCookie(cookie: String)(req: Request): Option[Cookie] =
     req.cookies.get(cookie)
 
+  private[this] val emptyRequestBody = Some("")
   private[this] def requestBody(req: Request): Option[String] =
     req.contentLength match {
-      case Some(n) if n > 0 =>
+      case Some(0) => emptyRequestBody
```
vkostyukov: Sorry to be picky, but can we make those names symmetric? How about

```scala
private[this] val someEmptyString = Some("")
private[this] val someEmptyByteArray = Some(Array.empty[Byte]) 
```

vkostyukov:
Just one nit, otherwise LGTM!

core/src/main/scala/io/finch/Endpoints.scala
omerzach: Yeah those sound better. Picky is good :)

vkostyukov:
Thanks again @omerzach! Merging this.

PR #643, opened 2016-09-13
Make Input accept json

### LINKED ISSUE #635: Make Input accept JSON payload
It would easier for users if they can shove case classes into an `Input` payload as JSON objects. I'm having something like this in mind (pseudocode):

```scala
def withJsonBody[A](a: A)(implicit e: Encode.Json[A], cs: Option[Charset] = Som("utf8")): Input =
  withBody(Buf.Text(e(a, cs)), Some("application/json;charset=" + cs))
```

BenFradet:
fixes #635

### DIFF
--- a/core/src/main/scala/io/finch/Input.scala
+++ b/core/src/main/scala/io/finch/Input.scala
@@ -1,9 +1,11 @@
 package io.finch
 
+import java.nio.charset.Charset
+
 import cats.Eq
 import com.twitter.finagle.http.{MediaType, Method, Request}
 import com.twitter.finagle.netty3.ChannelBufferBuf
-import com.twitter.io.{Buf, ConcatBuf}
+import com.twitter.io.{Buf, Charsets, ConcatBuf}
 import org.jboss.netty.handler.codec.http.{DefaultHttpRequest, HttpMethod, HttpVersion}
 import org.jboss.netty.handler.codec.http.multipart.{DefaultHttpDataFactory, HttpPostRequestEncoder}
 
@@ -30,6 +32,20 @@ final case class Input(request: Request, path: Seq[String]) {
 
     this
   }
+
+  /**
+   * Overrides (mutates) the payload of this input in order to reflect `a` as json and returns
+   * `this`.
+   *
+   * @note application/json will be passed as content type as well as the charset used to encode
+   *       the json
+   */
+  def withJson[A](a: A, charset: Option[Charset] = Some(Charsets.Utf8))(
+      implicit e: Encode.Json[A]): Input =
+    charset match {
+      case Some(cs) => withBody(e(a, cs), Some(s"application/json;charset=$cs"))
+      case None => withBody(e(a, Charsets.Utf8), Some("application/json"))
+    }
 
   /**
    * Adds (mutates) new `headers` to this input and returns `this`.
--- a/docs/endpoint.md
+++ b/docs/endpoint.md
@@ -632,18 +632,19 @@ val bar: Input = Input.post("/bar")
   .withHeaders("X-Header" -> "Y")
 ```
 
-At this point, there is no JSON-specific support encoded in the `Input` API (but will be in 0.11)
-so a generic `withBody` method may be used instead.
+Additionally, there is JSON-specific support in the `Input` API through `withJson`.
 
 ```scala
-import io.circe._
-import io.circe.syntax._
+import io.circe.generic.auto._
 import io.finch._
-import com.twitter.io.Buf
+import io.finch.circe._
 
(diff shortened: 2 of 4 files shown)


core/src/main/scala/io/finch/Input.scala
```diff
@@ -32,6 +34,19 @@ final case class Input(request: Request, path: Seq[String]) {
   }
 
   /**
+   * Overrides (mutates) the payload of this input in order to reflect `a` as json and returns
+   * `this`.
+   *
+   * @note application/json will be passed as content type as well as the charset used to encode
+   *       the json
+   */
+  def withJson[A](a: A)(implicit e: Encode[A], cs: Option[Charset] = Some(Charsets.Utf8)): Input =
```
vkostyukov: It should be `Encode.Json[A]` so we provide compile-time guarantee on using JSON encoders here.

core/src/main/scala/io/finch/Input.scala
```diff
@@ -32,6 +34,19 @@ final case class Input(request: Request, path: Seq[String]) {
...
+    cs match {
+      case Some(charset) => withBody(e(a, charset), Some(s"application/json;charset=$charset"))
+      case None => withBody(e(a, Charsets.Utf8), Some(s"application/json;charset=utf-8"))
```
vkostyukov: Let's not set `charset=` when it's `None`.

docs/endpoint.md
```diff
@@ -632,19 +632,20 @@ val bar: Input = Input.post("/bar")
...
-import io.circe.syntax._
+import io.circe.generic.auto._
 import io.finch._
-import com.twitter.io.Buf
+import io.finch.circe._
 
-val baz: Input = Input.put("/baz")
-  .withBody(Buf.Utf8(Map("a" -> "b").asJson.toString), Some("application/json;charset=utf8"))
+case class Baz(m: Map[String, String])
+val baz: Input = Input.put("/baz").withJson(Baz(Map("a" -> "b")))
```
vkostyukov: This is pretty sweet!

BenFradet: yup!

core/src/main/scala/io/finch/Input.scala
```diff
@@ -32,6 +34,20 @@ final case class Input(request: Request, path: Seq[String]) {
...
 
   /**
+   * Overrides (mutates) the payload of this input in order to reflect `a` as json and returns
+   * `this`.
+   *
+   * @note application/json will be passed as content type as well as the charset used to encode
+   *       the json
+   */
+  def withJson[A](a: A)(
+      implicit e: Encode.Json[A], cs: Option[Charset] = Some(Charsets.Utf8)): Input =
```
vkostyukov: `cs: Option[Charset]` should not be implicit. You want to get something like:

```scala
def withJson[A](a: A, cs: Option[Charset] = Some(Charsets.Utf8))(implicit e: Encode.Json[A]): Input =
```

core/src/main/scala/io/finch/Input.scala
```diff
@@ -32,6 +34,20 @@ final case class Input(request: Request, path: Seq[String]) {
...
+    cs match {
+      case Some(charset) => withBody(e(a, charset), Some(s"application/json;charset=$charset"))
+      case None => withBody(e(a, Charsets.Utf8), Some(s"application/json"))
```
yanana: You don't need string interpolation here.

BenFradet:
updated :+1: thanks for the review @vkostyukov @yanana

core/src/main/scala/io/finch/Input.scala
```diff
@@ -32,6 +34,20 @@ final case class Input(request: Request, path: Seq[String]) {
...
+   * Overrides (mutates) the payload of this input in order to reflect `a` as json and returns
+   * `this`.
+   *
+   * @note application/json will be passed as content type as well as the charset used to encode
+   *       the json
+   */
+  def withJson[A](a: A, charset: Option[Charset] = Some(Charsets.Utf8))(
+      implicit e: Encode.Json[A]): Input =
+    charset match {
+      case Some(cs) => withBody(e(a, cs), Some(s"application/json;charset=$cs"))
```
yanana: Should we put a space after the semicolon?

BenFradet: it's not done anywhere else in the code base so I don't think so, no.

yanana: I see.

vkostyukov:
Not sure why Codecov complains (-0.1% coverage?), but going to go ahead and merge it. Thanks @BenFradet!

Issue #644, opened 2016-09-15
Serving static content - example fails to compile


chbrown:
The example at https://github.com/finagle/finch/blob/master/docs/cookbook.md#serving-static-content isn't compiling for me: 

    [error]  found   : io.finch.Endpoint[com.twitter.util.Future[com.twitter.io.Buf]]
    [error]  required: io.finch.Endpoint[com.twitter.io.Buf]
    [error]   val file: Endpoint[Buf] = get("file") {
    [error]                                         ^

To replicate:

    git clone https://github.com/chbrown/finch-examples
    cd finch-examples && sbt compile

What's the modern (0.11.0-M3) way to do this?

A couple other documentation notes:
* `import com.twitter.finagle.Http` is missing from that example
* As you point out, serving static files isn't really the point of a finch app, but a fallback that makes development easier until you put the app behind a proxy, e.g., nginx, in production. So turning that endpoint into a Text.Plain service is kind of strange -- in my case, I'm (trying to) coproduct the static file endpoint with a lot of other actually-dynamic endpoints (which don't result in Text).

rpless:
Hi @chbrown try changing the `file` endpoint to:
```
  val file: Endpoint[Buf] = get("file") {
    Reader.readAll(reader).map(Ok)
  }
```
That endpoint needs to return a `Future[Output[Buf]]` instead of a `Output[Future[Buf]]`.

I also noticed that the server hangs when trying to `readAll` from `/dev/urandom`, so might want to serve another file (I think you need to stream `/dev/urandom`, not readAll from it).

As for coproducting the static file endpoint with other non-text endpoints, try this out: https://github.com/finagle/finch/blob/master/docs/cookbook.md#serving-multiple-content-types
Hope that helps.

vkostyukov:
Thanks @rpless! Any chance you want to open a PR with these fixes (including the missing import)?

Also thanks @chbrown for pointing this out.

vkostyukov:
Just merged #645.

PR #645, opened 2016-09-16
Update Static File doc

rpless:
based on feedback from @chbrown in #644 . Let me know if there are issues.


vkostyukov:
Thanks @rpless!

PR #646, opened 2016-09-16
Generalize Input.withBody

vkostyukov:
I figured since we have this powerful concept of content-type-as-a-type, why not to use it in `withBody` and make it generic? Basically, instead of duplicating logic in `withBody`, `withJson`, `withText`, `withXml`, `withWhatever`, we can just have one.

Before:
```scala
Input.post("/").withBody(Buf.Utf8("text"), Some("text/plain;charset=utf8"))
Input.post("/").withJson(Map("a" -> "b"), Some(Charsets.Utf8))
```

Now:

```scala
// Now:
Input.post("/").withBody[Text.Plain]("text", Some(Charsets.Utf8))
Input.post("/").withBody[Application.Json](Map("a" -> "b"), Some(Charsets.Utf8))

// We get the following for free:
import cats.instances.int._
Input.post("/").withBody[Text.Plain](42)
```

This not only looks nice but also compile-safe: there is no way to make a typo in `Application.Json` (it won't compile).

What do people think? /cc @BenFradet

### DIFF
--- a/benchmarks/src/main/scala/io/finch/benchmarks/BodyBenchmark.scala
+++ b/benchmarks/src/main/scala/io/finch/benchmarks/BodyBenchmark.scala
@@ -7,7 +7,7 @@ import org.openjdk.jmh.annotations.{Benchmark, Scope, State}
 @State(Scope.Benchmark)
 class BodyBenchmark extends FinchBenchmark {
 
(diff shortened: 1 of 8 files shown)


BenFradet:
This looks really nice :+1:

Issue #647, opened 2016-09-20
Use cats `show` to display circe decoding errors


clhodapp:
Right now, finch-circe calls `getMessage` on errors it gets from Circe in https://github.com/finagle/finch/blob/c447fb038943deac10d4b6137d13b345947ae920/circe/src/main/scala/io/finch/circe/Decoders.scala.

This produces really ugly error messages, such as `"body cannot be converted to Post: Attempt to decode value on failed cursor: El(DownField(systemRoles),false,false)."`

It really ought to use `Show` from cats to get something like `"body cannot be converted to Post: DecodingFailure at .systemRoles: Attempt to decode value on failed cursor."`, which is much more readable.

The change is to import `cats.syntax.show._` and call `.show` instead of `.getMessage` in https://github.com/finagle/finch/blob/c447fb038943deac10d4b6137d13b345947ae920/circe/src/main/scala/io/finch/circe/Decoders.scala.

vkostyukov:
Thanks @clhodapp! This is a great idea! Do you want to open a PR?

clhodapp:
Will do.

travisbrown:
👍 to this. In 0.5 it's also now possible to use the new [`HistoryOp.opsToPath`](https://github.com/travisbrown/circe/pull/267) to get the path string without the `DecodingFailure at` part, which might be a little cleaner here.

vkostyukov:
FWIW, there is an old ticket to make any JSON error more readable (pattern-matcheable): https://github.com/finagle/finch/issues/264

clhodapp:
I opted for the lazy takes-10-minutes, can-do-it-today solution that I originally suggested in my PR. I think more can be done to improve the display of decoding errors both on the finch side and the circe side at some point, though.

vkostyukov:
Fixed in #648.

joroKr21:
I'm not sure when but this has regressed at some point.

joroKr21:
The second commit here should address this: #1179

PR #648, opened 2016-09-22
Improve decoding errors from finch-circe

clhodapp:
Addresses #647

Ping @vkostyukov

### DIFF
--- a/circe/src/main/scala/io/finch/circe/Decoders.scala
+++ b/circe/src/main/scala/io/finch/circe/Decoders.scala
@@ -1,5 +1,6 @@
 package io.finch.circe
 
+import cats.syntax.show._
 import com.twitter.util.{Return, Throw, Try}
 import io.circe.Decoder
 import io.circe.jawn.decode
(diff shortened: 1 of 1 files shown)


circe/src/main/scala/io/finch/circe/Decoders.scala
```diff
@@ -4,6 +4,7 @@ import com.twitter.util.{Return, Throw, Try}
 import io.circe.Decoder
 import io.circe.jawn.decode
 import io.finch.{Decode, Error}
+import cats.syntax.show._
```
vkostyukov: I think these need to be sorted to make `sbt scalastyle` happy.

clhodapp:
Oh no! I'm not stylish enough!

circe/src/main/scala/io/finch/circe/Decoders.scala
clhodapp: Yes. I just came to exactly the same conclusion. I'll just fix that...

vkostyukov:
Thanks @clhodapp! Merging this.

PR #651, opened 2016-09-30
Upgrade scoverage

BenFradet:
Since scoverage/sbt-scoverage#146 is fixed we might as well update scoverage.


vkostyukov:
Thanks @BenFradet!

Issue #593, opened 2016-05-13
Add support for Server Sent Events (SSE)


Tzbob:
It would be nice to have support for Server Sent Events (https://www.w3.org/TR/eventsource/).

I'm not that familiar with Finch but I'm pretty sure it's already possible to comply with SSE in its current state, perhaps adding explicit support to something like `Endpoint[AsyncStream[ServerSentEvent]` would be sufficient?

rpless:
I've started a little sketch of how this could work here: https://github.com/rpless/finch/tree/wip-sse

vkostyukov:
This is done in #655.

PR #655, opened 2016-10-18
Add basic support for Server Sent Events.

rpless:
Proposed fix for #593 
Adds a ToResponse to create to hook into the asyncBodyResponse
Adds an example of SSE
Adds a section to the cookbook about SSE

Two things to call out:
- This support is very bare-bones and makes it the responsibility of the user to provide properly formatted events
- The [EventSource Standards](https://html.spec.whatwg.org/multipage/comms.html#authoring-notes) advises against using chunked requests, but `asyncResponseBuilder` does use chunked requests. When I tried implementing my own version of `asyncResponseBuilder` that didn't use chunking the `writer` would not write any data into the event. Not sure if it just isn't possible or if there was some method on the writer that I missed that would get it to write.

### DIFF
--- a/core/src/main/scala/io/finch/internal/ToResponse.scala
+++ b/core/src/main/scala/io/finch/internal/ToResponse.scala
@@ -38,7 +38,7 @@ trait LowPriorityToResponseInstances {
     def apply(a: A, cs: Charset): Response = fn(a, cs)
   }
 
-  protected def asyncResponseBuilder[A, CT <: String](writer: (A, Charset) => Buf)(implicit
+  protected[finch] def asyncResponseBuilder[A, CT <: String](writer: (A, Charset) => Buf)(implicit
     w: Witness.Aux[CT]
   ): Aux[AsyncStream[A], CT] = instance { (as, cs) =>
     val rep = Response()
--- a/core/src/main/scala/io/finch/package.scala
+++ b/core/src/main/scala/io/finch/package.scala
@@ -39,5 +39,6 @@ package object finch extends Endpoints with Outputs with ValidationRules {
   object Text {
     type Plain = Witness.`"text/plain"`.T
     type Html = Witness.`"text/html"`.T
+    type EventStream = Witness.`"text/event-stream"`.T
   }
 }
--- a/docs/cookbook.md
+++ b/docs/cookbook.md
@@ -9,6 +9,7 @@
 * [Defining custom endpoints](cookbook.md#defining-custom-endpoints)
 * [CORS in Finch](cookbook.md#cors-in-finch)
 * [Converting between Scala futures and Twitter futures](cookbook.md#converting-between-scala-futures-and-twitter-futures)
+* [Server Sent Events](cookbook.md#server-sent-events)
 
 This is a collection of short recipes of "How to X in Finch".
 
@@ -459,6 +460,36 @@ implicit class RichSFuture[A](val f: SFuture[A]) extends AnyVal {
   }
 }
 ```
+
+### Server Sent Events
+
+Finch offers support for Server Sent Events through the `finch-sse` sub-project.
+Server Sent Events are represented as `AsyncStream`s:
+```scala
+import com.twitter.concurrent.AsyncStream
+import com.twitter.conversions.time._
+import com.twitter.finagle.Http
+import com.twitter.util.Future
+import io.finch._
+import io.finch.sse.ServerSentEvent
+
+def stream[A](sse: AsyncStream[ServerSentEvent[A]]): Output[AsyncStream[ServerSentEvent[A]]] = {
+  Ok(sse).withHeader("Cache-Control" -> "no-cache").withHeader("Connection" -> "keep-alive")
+}
+
+def currentTime: String = java.util.Calendar.getInstance().getTime.toString
+
+def time: AsyncStream[ServerSentEvent[String]] = {
+  ServerSentEvent(currentTime) +::
+    AsyncStream.fromFuture(Future.sleep(1000.millis)).flatMap(_ => time)
+}
+
+val sse: Endpoint[AsyncStream[ServerSentEvent[String]]] = get("sse")(stream(time))
+Http.server.serve(":8081", sse.toServiceAs[Text.EventStream])
+```
+In this example, the `stream` method sets up an `Output` that returns a stream with the necessary headers.
+It is then used as one uses anyh other Finch endpoint.
+To serve it as a Server Sent Event, the service creation is parameterized over `Text.EventStream`.
 
 --
 Read Next: [Best Practices](best-practices.md)
--- a/sse/src/main/scala/io/finch/sse/ServerSentEvent.scala
+++ b/sse/src/main/scala/io/finch/sse/ServerSentEvent.scala
@@ -1,0 +1,46 @@
+package io.finch.sse
+
+import java.nio.charset.Charset
+
+import cats.Show
+import com.twitter.concurrent.AsyncStream
+import com.twitter.io.Buf
+import io.finch.{Encode, Ok, Output, Text}
+import io.finch.internal.{BufText, ToResponse}
+
+case class ServerSentEvent[A](
+  data: A,
+  id: Option[String] = None,
+  event: Option[String] = None,
+  retry: Option[Long] = None
+)
+
+object ServerSentEvent {
+
+  implicit def sseAsyncToResponse[A](implicit
+     e: Encode.Aux[A, Text.EventStream]
+  ): ToResponse.Aux[AsyncStream[A], Text.EventStream] =
+    ToResponse.asyncResponseBuilder((a, cs) => e(a, cs).concat(Buf.Utf8("\n")))
+
+  implicit def encodeEventStream[A](implicit s: Show[A]): Encode.Aux[ServerSentEvent[A], Text.EventStream] = {
+    Encode.instance[ServerSentEvent[A], Text.EventStream]({ (event: ServerSentEvent[A], c: Charset) =>
+      encodeEvent[A](event, c, { (a: A, cs: Charset) => BufText(s.show(a), cs) })
+    })
+  }
+
+  private[sse] def encodeEvent[A](sse: ServerSentEvent[A], cs: Charset, e: (A, Charset) => Buf): Buf = {
+    val dataBuf = BufText("data:", cs).concat(e(sse.data, cs)).concat(BufText("\n", cs))
+    val eventType = sse.event.map(e => s"event:$e\n").getOrElse("")
+    val id = sse.id.map(id => s"id:$id\n").getOrElse("")
+    val retry = sse.retry.map(retry => s"retry:$retry\n").getOrElse("")
+    val restBuf = BufText(eventType + id + retry, cs)
+    dataBuf.concat(restBuf)
+  }
+
+  /** By default Finch encodes an exception as a Comment */
+  implicit val encodeSseException: Encode.Aux[Exception, Text.EventStream] = {
+    Encode.instance[Exception, Text.EventStream]({ (e: Exception, c: Charset) =>
+      BufText(s":${Option(e.getMessage).getOrElse("Unknown Error")}\n", c)
+    })
+  }
+}
--- a/build.sbt
+++ b/build.sbt
@@ -138,7 +138,9 @@ lazy val finch = project.in(file("."))
     "io.circe" %% "circe-generic" % circeVersion,
     "io.spray" %%  "spray-json" % sprayVersion
   ))
-  .aggregate(core, argonaut, jackson, json4s, circe, playjson, sprayjson, benchmarks, test, jsonTest, oauth2, examples)
+  .aggregate(
+    core, argonaut, jackson, json4s, circe, playjson, sprayjson, benchmarks, test, jsonTest, oauth2, examples, sse
+  )
   .dependsOn(core, circe)
 
 lazy val core = project
@@ -221,6 +223,11 @@ lazy val oauth2 = project
   ))
   .dependsOn(core)
 
+lazy val sse = project
+  .settings(moduleName := "finch-sse")
+  .settings(allSettings)
+  .dependsOn(core)
+
 lazy val examples = project
   .settings(moduleName := "finch-examples")
   .settings(allSettings)
@@ -234,6 +241,7 @@ lazy val examples = project
       |io\.finch\.streaming\..*;
       |io\.finch\.oauth2\..*;
       |io\.finch\.wrk\..*;
+      |io\.finch\.sse\..*;
     """.stripMargin)
   .settings(
     libraryDependencies ++= Seq(
--- a/sse/src/test/scala/io/finch/sse/ServerSentEventSpec.scala
+++ b/sse/src/test/scala/io/finch/sse/ServerSentEventSpec.scala
@@ -1,0 +1,110 @@
+package io.finch.sse
+
+import java.nio.charset.Charset
+
+import cats.Show
+import com.twitter.concurrent.AsyncStream
+import com.twitter.io.{Charsets, ConcatBuf}
+import io.finch.internal.BufText
+import org.scalacheck.Gen.Choose
+import org.scalacheck.Prop.BooleanOperators
+import org.scalacheck.{Arbitrary, Gen}
+import org.scalatest.prop.Checkers
+import org.scalatest.{FlatSpec, Matchers}
+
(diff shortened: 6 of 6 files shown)


rpless:
Hmmm, looks like I forgot to exclude the new example from coverage. Let me fix that.

vkostyukov:
Thanks @rpless!

- I guess it's always a user responsibility to provide domain types and their encoders for SSE. Not sure if there is anything we can do to make it simpler.

- Not sure I follow, if you don't need streaming, why using `AsyncStream` in the first place?

vkostyukov:
Actually, looking at [Akka's implementation](https://github.com/hseeberger/akka-sse/blob/master/akka-sse/src/main/scala/de/heikoseeberger/akkasse/ServerSentEvent.scala), there is a protocol we can follow, right?

Maybe we should build protocol types (eg: messages) as part of a separate package (i.e., finch-sse)?

I'm not super familiar with SSE and I guess I have to spend some time trying figure out what exactly it is. But it feels we can do more here.

rpless:
I'm always hesitant to just create new sub projects when there is no initial indication that we need one, but if you think that's the way to go, it should be fairly easy to do given that there is a BNF for messages. I'm fine with moving this PR toward that.

To clarify my first comment, the potential issue is [Chunking the Request](https://github.com/finagle/finch/blob/master/core/src/main/scala/io/finch/internal/ToResponse.scala#L45). The spec I linked to says: `Authors are also cautioned that HTTP chunking can have unexpected negative effects on the reliability of this protocol...`.

vkostyukov:
Thanks @rpless! I think I get it now.

Given that SSE are about polling and sending "push notifications" to the subscribed client (a browser), there is no way we can achieve that w/o streaming in Finagle. Normally (eg: if we would have a fine-grained control over the HTTP connection), we can keep the connection alive and shove our events there until there is something to send.

Akka's SSE implementation uses something called closed-delimited streaming that indicates the end of the stream by closing the connection. Technically, this is exactly what we want.

In Finagle, there is no notion of connection exposed to the end-user and the whole RPC story is encoded in `Req => Future[Rep]`. Because of that limitation, the only way we can achieve something similar is to use chunked responses (`AsyncStream` in Finch). I understand that the authours of the SSE spec do not recommend that, but this is the only option we have and after searching the web yesterday I didn't manage to find any prove that chunked-responses don't work well with SSE. In fact, most of the server-side implementation utilise them.

So I think our initial idea about using `AsyncStream` is right. Although I still think we deserves its own package given that we'd need to provide a data-type for SSE (i.e., `ServerSentEvent`). I can imagine this to be a case class that also supports an arbitrary `data` field for which there is an `Encode` method available.

What do you think @rpless?

rpless:
That all makes sense to me. Its been awhile since I've used SSE and when I saw that note in the spec, I figured I'd call it out.
I'll start working on making a `finch-sse` package with a data type for Server Sent Events.

vkostyukov:
Here is an example of the [data type](https://github.com/http4s/http4s/blob/master/core/src/main/scala/org/http4s/ServerSentEvent.scala) from http4s. I just talked to @bryce-anderson to confirm that http4s also uses chunked responses for that.

rpless:
Not sure why that failed on Travis. The same commit passes on my fork's travis setup: https://travis-ci.org/rpless/finch/builds/170636276

core/src/main/scala/io/finch/Encode.scala
```diff
@@ -23,6 +23,7 @@ trait LowPriorityEncodeInstances {
 
   type Json[A] = Aux[A, Application.Json]
   type Text[A] = Aux[A, Text.Plain]
+  type EventStream[A] = Aux[A, Text.EventStream]
```
vkostyukov: I'm not 100% sure we actually need these aliases here. I agree they are convenient, but given that users won't need to define instances by their own, I think, we can totally omit them here. What do you think?

core/src/main/scala/io/finch/Encode.scala
```diff
@@ -36,6 +37,9 @@ trait LowPriorityEncodeInstances {
   def text[A](fn: (A, Charset) => Buf): Text[A] =
     instance[A, Text.Plain](fn)
 
+  def eventStream[A](fn: (A, Charset) => Buf): EventStream[A] =
```
vkostyukov: Also this.

core/src/main/scala/io/finch/internal/ToResponse.scala
```diff
@@ -71,6 +71,11 @@ trait HighPriorityToResponseInstances extends LowPriorityToResponseInstances {
   ): Aux[AsyncStream[A], Text.Plain] =
     asyncResponseBuilder((a, cs) => e(a, cs).concat(newLine))
 
+  implicit def sseAsyncToResponse[A](implicit
```
vkostyukov: I think we can define this under `ServerSentEvent` companion object and it should be resolved just fine. Any chance you can try this?

vkostyukov: I just want to make sure we separate SSE from finch-core, because otherwise it doesn't actually make a lot of sense to extract that into t a separate package. We can try, if it's now working, I'm happy to go back to an initial idea where SSE is defined as part of finch-core.

core/src/main/scala/io/finch/package.scala
```diff
@@ -39,5 +39,6 @@ package object finch extends Endpoints with Outputs with ValidationRules {
   object Text {
     type Plain = Witness.`"text/plain"`.T
     type Html = Witness.`"text/html"`.T
+    type EventStream = Witness.`"text/event-stream"`.T
```
vkostyukov: That's totally fine to keep this in core.

core/src/main/scala/io/finch/sse/ServerSentEvent.scala
```diff
@@ -0,0 +1,34 @@
...
+)
+
+object ServerSentEvent {
+  type EventStream[A] = AsyncStream[ServerSentEvent[A]]
+
+  def stream[A](sse: EventStream[A]): Output[EventStream[A]] = {
+    Ok(sse).withHeader("Cache-Control" -> "no-cache").withHeader("Connection" -> "keep-alive")
+  }
+
+  implicit def encodeEventStream[A <: AnyRef](implicit s: Show[A]): Encode.EventStream[ServerSentEvent[A]] = {
```
vkostyukov: I love this!

vkostyukov: Do we actually need `<: AnyRef`?

docs/cookbook.md
```diff
@@ -460,6 +461,36 @@ implicit class RichSFuture[A](val f: SFuture[A]) extends AnyVal {
...
+Finch has basic support for Server Sent Event using `AsyncStream` that has a content type of `Text.EventStream`.
+```scala
+import com.twitter.concurrent.AsyncStream
+import com.twitter.finagle.http.Http
+import io.finch._
+
+case class Example(x: Int)
+def exampleStream: AsyncStream[Example] = ???
+
+implicit def encodeEventStream[A <: AnyRef]: Encode.EventStream[AsyncStream[A]] = ???
```
vkostyukov: I don't think users need to provide that. Given that there is `Show[Example]`, everything should just work, right?

examples/src/main/scala/io/finch/sse/Main.scala
```diff
@@ -0,0 +1,38 @@
...
+
+case class Example(x: Int)
+object Example {
+  implicit val show: Show[Example] = Show.fromToString
+}
+
+object Main {
+  import Example._
+
+  implicit val encodeExceptionAsTextPlain: Encode.EventStream[Exception] = {
```
vkostyukov: I wish we could provide that out of the box, but I'm not sure where it could be defined.

rpless:
Thanks for the feedback @vkostyukov!
I have a couple answers:
1. I agree. I've used JSON in SSE before and I've seen plenty of other people do it too, so I suspect there is a strong case for having support for that. I have a couple of thoughts on how to achieve that, they'd I'd like to play with.
2. I agree. It is leftover from before we decided to provide the case class and other helpers. My current think is that we actually drop the example and either clean up the snippet in the cookbook or put something in the Endpoint `docs` about it.
3. To clarify to, do you want this in its own sub-project that ships a separate artifact and needs to be included as a separate dependency or just have it in a separate package in `core`.

examples/src/main/scala/io/finch/sse/Main.scala
rpless: One thing SSE allows is for "comments" to be sent. Perhaps the default serialization of an error could be a comment.

vkostyukov:
Thanks @rpless! Sorry for the confusion around package vs. project! I meant `finch-sse` project that only depends on `finch-core`. But again, if that means we can not provide default instances for `Encode` (both `Show` and `Exception`), I'm happy to drop this idea and make SEE a part of `finagle-core`.

rpless:
Ok cool. I'll try to make a another project. If for whatever reason it does not work, I'll let you know and go back to the package in core.

vkostyukov:
@rpless, how are things going with SSEs? Is there anything I can help with?

rpless:
Hi @vkostyukov. I haven't had as much time to spend on this as I'd like. I've pushed up my latest work.
It does the following:
- backs out both the cookbook example and the examples file. (here's one I've been using for testing with a browser though: https://gist.github.com/rpless/77d995c3f41c59c3d58094f7880bdb70)
- Creates a finch sse project

The only thing I haven't moved is the `ToResponse` instance for EventStream. I was having issues when I moved it into the sse project. I think it was either an implicit resolution issue or an issue with `Buf` not flushing but I haven't yet tracked it down. 

My plan for encoding JSON in the payload also didn't work, but that may be due to the previous issue. The gist is that you can implicitly pull in one of the the Json encoders and use that with `ServerSentEvent.encodeEvent`.
I should have some time to work on this tonight and I'm hoping to get some `docs` in and move the `ToResponse` instance.

rpless:
I just figured out the build issue, looks like I started by fork before 9a73f73737bf1af0cd519f7cc6c5b26fb3a3fa08, which changes where `newLine` is defined. I'll pull in the latest to fix it.

vkostyukov:
Thanks @rpless! Let's not block ourselves on JSON support yet.

I think there is a path forward with the current solution. One way to do JSON is to provide `Show[Json]` (Circe does that out of the box). I think it's totally fine solution for now.

sse/src/main/scala/io/finch/sse/ServerSentEvent.scala
```diff
@@ -0,0 +1,51 @@
...
+import io.finch.internal.{BufText, LowPriorityToResponseInstances}
+
+case class ServerSentEvent[A](
+  data: A,
+  id: Option[String] = None,
+  event: Option[String] = None,
+  retry: Option[Long] = None
+)
+
+object ServerSentEvent extends LowPriorityToResponseInstances {
```
vkostyukov: Why do we need this `LowPriorityToResponseInstances`? I'm not sure I'm happy about this inheritance.

sse/src/main/scala/io/finch/sse/ServerSentEvent.scala
```diff
@@ -0,0 +1,51 @@
...
+  type EventStream[A] = AsyncStream[ServerSentEvent[A]]
+
+  def stream[A](sse: EventStream[A]): Output[EventStream[A]] = {
+    Ok(sse).withHeader("Cache-Control" -> "no-cache").withHeader("Connection" -> "keep-alive")
```
vkostyukov: Can we move that part into a cookbook? I'm not sure we need an explicit API method for this.

sse/src/main/scala/io/finch/sse/ServerSentEvent.scala
```diff
@@ -0,0 +1,51 @@
...
+  ): Aux[AsyncStream[A], Text.EventStream] =
+    asyncResponseBuilder((a, cs) => e(a, cs).concat(Buf.Utf8("\n")))
+
+  implicit def encodeEventStream[A](implicit s: Show[A]): Encode.Aux[ServerSentEvent[A], Text.EventStream] = {
+    Encode.instance[ServerSentEvent[A], Text.EventStream]({ (event: ServerSentEvent[A], c: Charset) =>
+      encodeEvent[A](event, c, { (a: A, cs: Charset) => BufText(s.show(a), cs) })
+    })
+  }
+
+  def encodeEvent[A](sse: ServerSentEvent[A], cs: Charset, e: (A, Charset) => Buf): Buf = {
```
vkostyukov: `private[this]`?

sse/src/main/scala/io/finch/sse/ServerSentEvent.scala
```diff
@@ -0,0 +1,51 @@
...
+    val dataBuf = BufText("data:", cs).concat(e(sse.data, cs)).concat(BufText("\n", cs))
+    val eventType = sse.event.map(e => s"event:$e\n").getOrElse("")
+    val id = sse.id.map(id => s"id:$id\n").getOrElse("")
+    val retry = sse.retry.map(retry => s"retry:$retry\n").getOrElse("")
+    val restBuf = BufText(eventType + id + retry, cs)
+    dataBuf.concat(restBuf)
+  }
+
+  /** By default Finch encodes an exception as a Comment */
+  implicit val encodeSseException: Encode.Aux[Exception, Text.EventStream] = {
```
vkostyukov: Does it actually work?

sse/src/main/scala/io/finch/sse/ServerSentEvent.scala
rpless: Works for me.

sse/src/main/scala/io/finch/sse/ServerSentEvent.scala
rpless: I left this available with the thought that it can be used by consumers of the library to construct their own server sent events. But I can make this private.

sse/src/main/scala/io/finch/sse/ServerSentEvent.scala
rpless: This is because `asyncResponseBuilder` is inaccessible otherwise. Thinking more about it, I can probably add a `io.finch.sse.internal` package with a `SSEToResponse` that extends this and provides `asyncResponseBuilder`.

vkostyukov: Or you could make the original `asyncResponseBuilder` `private[finch]` and use it directly.

sse/src/main/scala/io/finch/sse/ServerSentEvent.scala
rpless: It does, but its not great. The server does produce a new event, but it looks like browsers discard comments. We might want to rethink how we treat errors.

sse/src/main/scala/io/finch/sse/ServerSentEvent.scala
```diff
@@ -0,0 +1,47 @@
...
+
+case class ServerSentEvent[A](
+  data: A,
+  id: Option[String] = None,
+  event: Option[String] = None,
+  retry: Option[Long] = None
+)
+
+object ServerSentEvent {
+  type EventStream[A] = AsyncStream[ServerSentEvent[A]]
```
vkostyukov: I'd vote for not introducing this type. `AsyncStean[ServerSentEvent[A]]` isn't that verbose and quite self-documenting.

docs/cookbook.md
```diff
@@ -460,6 +461,26 @@ implicit class RichSFuture[A](val f: SFuture[A]) extends AnyVal {
 }
 ```
 
+### Server Sent Events
+
+Finch offers support for Server Sent Events through the `finch-sse` sub-project.
+Server Sent Events are represented as `AsyncStream`s:
+```scala
+import io.finch.sse.ServerSentEvent
```
vkostyukov: Let's make sure all the imports present. (`io.finch._` is missing).

docs/cookbook.md
```diff
@@ -460,6 +461,26 @@ implicit class RichSFuture[A](val f: SFuture[A]) extends AnyVal {
...
+import ServerSentEvent._
+
+def stream[A](sse: EventStream[A]): Output[EventStream[A]] = {
```
vkostyukov: What do you think about writing an example that streams current local time, every second? I'm happy to alter this example in the upcoming commit if you don't feel like doing it.

vkostyukov:
Thanks @rpless! I going to go ahead and merge this in. I'd like us to rethink the encoding-exception behaviour, but let's do that separate.

PR #658, opened 2016-10-20
Invalid class path reference in Documentation

wennergr:
com.twitter.finagle.Response doesn't exists.


ilya-murzinov:
We should probably use https://github.com/tpolecat/tut to avoid such errors.

vkostyukov:
@ilya-murzinov I was thinking about tut for a long time. Would be totally awesome to get it at some point. Mind filling a ticket?

vkostyukov:
Thanks @wennergr! Merging this.

PR #661, opened 2016-11-02
Update Cats, circe, ScalaCheck, and ScalaTest versions

travisbrown:
I'm not sure we want to merge this yet, but I wanted to show that it works. :smile: 

* `Xor` is gone.
* Cats provides its own `UUID` instances.
* ScalaTest has moved a couple of things around.
* `Arbitrary` instances for functions now require `Cogen` instances for the argument types.

I need to take a closer look at the `Cogen` instances, but I think they make sense.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,7 +2,7 @@ language: scala
 sudo: false
 scala:
   - 2.10.6
-  - 2.11.7
+  - 2.11.8
 jdk:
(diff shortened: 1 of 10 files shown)


vkostyukov:
Thanks @travisbrown! We totally want these upgrades in the 0.11 release so I'm happy to merge this in.

PR #662, opened 2016-11-05
Increase test coverage of the Error classes

DieBauer:
The RequestError class accepts a Seq[Throwable], in case this is empty we should handle this accordingly and print a better message.

### DIFF
--- a/core/src/main/scala/io/finch/Endpoint.scala
+++ b/core/src/main/scala/io/finch/Endpoint.scala
@@ -1,6 +1,9 @@
 package io.finch
 
-import cats.Alternative
+import java.nio.charset.Charset
+import scala.reflect.ClassTag
+
+import cats.{Alternative, Functor}
 import cats.data.NonEmptyList
 import com.twitter.finagle.Service
 import com.twitter.finagle.http.{Cookie, Request, Response, Status}
@@ -8,8 +11,6 @@ import com.twitter.io.Buf
 import com.twitter.util.{Future, Return, Throw, Try}
 import io.catbird.util.Rerunnable
 import io.finch.internal._
-import java.nio.charset.Charset
-import scala.reflect.ClassTag
 import shapeless._
 import shapeless.ops.adjoin.Adjoin
 import shapeless.ops.hlist.Tupler
@@ -162,13 +163,13 @@ trait Endpoint[A] { self =>
         case (_, Throw(e)) => Future.exception(e)
       }
 
-    private[this] def collectExceptions(a: Throwable, b: Throwable): Error.RequestErrors = {
-      def collect(e: Throwable): Seq[Throwable] = e match {
-        case Error.RequestErrors(errors) => errors
-        case rest => Seq(rest)
+    private[this] def collectExceptions(a: Throwable, b: Throwable): Error.Multiple = {
+      def collect(e: Throwable): NonEmptyList[Throwable] = e match {
+        case Error.Multiple(errors) => errors
+        case rest => NonEmptyList.of(rest)
       }
 
-      Error.RequestErrors(collect(a) ++ collect(b))
+      Error.Multiple(collect(a).concat(collect(b)))
     }
 
     def apply(input: Input): Endpoint.Result[(A, B)] =
@@ -505,16 +506,16 @@ object Endpoint {
           case Throw(e) => Error.NotParsed(self.item, tag, e)
         }
 
-        if (errors.isEmpty)
-          Future.const(Try.collect(decoded).map(seq => NonEmptyList(seq.head, seq.tail.toList)))
-        else
-          Future.exception(Error.RequestErrors(errors))
+        NonEmptyList.fromList(errors) match {
+          case None => Future.const(Try.collect(decoded).map(seq => NonEmptyList(seq.head, seq.tail.toList)))
+          case Some(err) => Future.exception(Error.Multiple(Functor[NonEmptyList].widen(err)))
+        }
       }
   }
 
   /**
     * Implicit conversion that allows to call `as[A]` on any `Endpoint[Seq[String]]` to perform a
-    * type conversion based on an implicit ``DecodeRequest[A]` which must be in scope.
+    * type conversion based on an implicit `DecodeRequest[A]` which must be in scope.
     *
     * The resulting endpoint will fail when the result is non-empty and type conversion fails on one
     * or more of the elements in the `Seq`. It will succeed if the result is empty or type conversion
@@ -539,8 +540,10 @@ object Endpoint {
           case Throw(e) => Error.NotParsed(self.item, tag, e)
         }
 
-        if (errors.isEmpty) Future.const(Try.collect(decoded))
-        else Future.exception(Error.RequestErrors(errors))
+        NonEmptyList.fromList(errors.toList) match {
+          case None => Future.const(Try.collect(decoded))
+          case Some(err) => Future.exception(Error.Multiple(Functor[NonEmptyList].widen(err)))
+        }
       }
   }
 
--- a/core/src/main/scala/io/finch/Endpoint.scala
+++ b/core/src/main/scala/io/finch/Endpoint.scala
@@ -164,9 +164,11 @@ trait Endpoint[A] { self =>
       }
 
     private[this] def collectExceptions(a: Throwable, b: Throwable): Error.Multiple = {
-      def collect(e: Throwable): NonEmptyList[Throwable] = e match {
+      def collect(e: Throwable): NonEmptyList[Error] = e match {
         case Error.Multiple(errors) => errors
-        case rest => NonEmptyList.of(rest)
+        case rest => NonEmptyList.of(
+          Error(Try(rest.getMessage).getOrElse("Exception message was null"))
+        )
       }
 
       Error.Multiple(collect(a).concat(collect(b)))
--- a/core/src/main/scala/io/finch/Error.scala
+++ b/core/src/main/scala/io/finch/Error.scala
@@ -1,6 +1,10 @@
 package io.finch
 
+import scala.compat.Platform.EOL
 import scala.reflect.ClassTag
+
+import cats.data.NonEmptyList
+import io.finch.items.RequestItem
 
 /**
  * A basic error from a Finch application.
@@ -9,10 +13,13 @@ trait Error extends Exception
 
 object Error {
 
-  import items._
-
   def apply(message: String): Error = new Error {
     override def getMessage: String = message
+  }
+
+  object RequestErrors {
+    @deprecated("Use Multiple instead", "0.11")
+    def unapply(e: Throwable): Option[Seq[Throwable]] = Some(Seq(e))
   }
 
   /**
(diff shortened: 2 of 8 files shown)


vkostyukov:
Really nice @DieBauer! Since you're here, do you think we should refactor `RequestErrors` such that it constains a non-empty list (which it is) (i.e., Cat's `Nel`)?

DieBauer:
Thanks @vkostyukov :) It seems indeed weird to accept `Nil`, if we expect (and there will always be) at least 1 element in there.
I'm not really familiar with the cats-library, what would the implications be to use the `Nel` construction, runtimewise?
I do think that it is better to be explicit about the nonEmptyness of this parameter, so indeed refactor to using that `Nel`-construct.

DieBauer:
I was fiddling around with `Nel`, however this class is invariant.
```
 found   : cats.data.NonEmptyList[io.finch.Error.NotParsed]
 required: cats.data.NonEmptyList[Throwable]
Note: io.finch.Error.NotParsed <: Throwable, but class NonEmptyList is invariant in type A.
You may wish to define A as +A instead. (SLS 4.5)
          case Some(err) => Future.exception(Error.RequestErrors(err))
```
RequestErrors looks like this:
```scala
  final case class RequestErrors(errors: NonEmptyList[Throwable]) extends Error
```
Is this the way we want to take the solution, or do I have to roll my own `Nel`?

DieBauer:
Implemented RequestErrors with `Nel`, updated tests and docs, what do you think @vkostyukov?

vkostyukov:
Sorry for the delay on this @DieBauer! I will have a closer look today.

core/src/main/scala/io/finch/Endpoint.scala
```diff
@@ -163,12 +164,12 @@ trait Endpoint[A] { self =>
...
     private[this] def collectExceptions(a: Throwable, b: Throwable): Error.RequestErrors = {
-      def collect(e: Throwable): Seq[Throwable] = e match {
+      def collect(e: Throwable): NonEmptyList[Throwable] = e match {
         case Error.RequestErrors(errors) => errors
-        case rest => Seq(rest)
+        case rest => NonEmptyList.of(rest)
       }
 
-      Error.RequestErrors(collect(a) ++ collect(b))
+      Error.RequestErrors(collect(a) concat collect(b))
```
vkostyukov: Let's do `collect(a).concat(collect(b))`.

core/src/main/scala/io/finch/Endpoint.scala
```diff
@@ -503,10 +504,10 @@ object Endpoint {
           case Throw(e) => Error.NotParsed(self.item, tag, e)
         }
 
-        if (errors.isEmpty)
-          Future.const(Try.collect(decoded).map(seq => NonEmptyList(seq.head, seq.tail.toList)))
-        else
-          Future.exception(Error.RequestErrors(errors))
+        NonEmptyList.fromList(errors) match {
+          case None => Future.const(Try.collect(decoded).map(seq => NonEmptyList(seq.head, seq.tail.toList)))
+          case Some(err) => Future.exception(Error.RequestErrors(Functor[NonEmptyList].widen(err)))
```
vkostyukov: Why `Functor[...].widen` is necessary?

vkostyukov:
Thinking about this more. What's your feeling on trying to keep this source compatible?

I thnk since we break the API, we might want to change the name as well. I'm not a huge fan of `RequestErrors` so why don't we change it something like `Error.Multiple` and also provide a backward-compatible extractor (marked as `@deprecated`) for `RequestErrors(list: List[Throwable])`.

Also since we're here. I'm thinking about using changing `Multiple(errors: Nel[Throwable])` to `Multiple(errors: Nel[Error])`.

@DieBauer What do you think about this?

core/src/main/scala/io/finch/Endpoint.scala
DieBauer: This is because `Nel` is invariant in `A`. 

```
 found   : cats.data.NonEmptyList[io.finch.Error.NotParsed]
 required: cats.data.NonEmptyList[Throwable]
Note: io.finch.Error.NotParsed <: Throwable, but class NonEmptyList is invariant in type A.
You may wish to define A as +A instead. (SLS 4.5)
          case Some(err) => Future.exception(Error.RequestErrors(err))
```

Problably when we change the signature of `RequestErrors` from taking a `Throwable` to an `Error` this widen will be unneccesary, but I'm not entirely sure.

DieBauer:
I think source compatibility is a good idea, and renaming the instance is also fine by me, I don't have strong feelings towards the naming of this case class ;)
Since it's only used as `bag` in case of Endpoint.product throws errors, `Multiple`, or `Aggregate` seems to fit? 

By making the Nel only take `Error` instead of throwable, I'm not sure if `Rerunnable[Output[?]].liftToTry` can always be converted to an `Error`, or what Error will make sense, can you comment on this?

What do you mean by "backward-compatible extractor"?

vkostyukov:
Thanks @DieBauer!

What I mean is to 

1. Rename `case class RequestErrors` to `case class Multiple` and then
2. Define `object RequestErrorrs` with a custom extractor (`def unapply(e: Exception): Option[Seq[Throwable]]`)

This way users would still be able to keep their existing pattern matching expressions that matched on `RequestErrors`.

DieBauer:
Thanks for the clarification @vkostyukov! 
Now I run into an issue by changing the signature of `Multiple(errors: Nel[Throwable])` to accept `Nel[Error]`. The applicativelaws are failing in EndpointSpec and it seems to have to do with the change of the collect method at line 166 in Endpoint.

in `collectExceptions` there is a method collect defined which I implemented like this:
```scala
def collect(e: Throwable): NonEmptyList[Throwable] = e match {
  case Error.Multiple(errors) => errors
  case rest => NonEmptyList.of(rest)
}
```
but by changing the signature, it should be implemented more like this:
```scala
def collect(e: Throwable): NonEmptyList[Error] = e match {
  case Error.Multiple(errors) => errors
  case rest => NonEmptyList.of(
    Error(Try(rest.getMessage).getOrElse("Exception message was null"))
  )
}
```
I'm not entirely sure why this fails, maybe because we are instantiating an anonymous Error class and equality of these classes is not implemented(?). 

What do you think, I'm not an expert in scalacheck, but it seems to point to the arbitraryEndpoint generator in FinchSpec

DieBauer:
I think I found what's wrong with this implementation of a trait extending Exception.

consider this test-case: 
```scala
trait Error extends Exception

it should "equal anonymous Error classes" in {
    val error = Error("message")
    val e = Error("message")
    val faulty = Error("other message")
    e.equals(error) shouldBe true
    error.equals(faulty) shouldBe false
}
```
since the `Error.apply(m:String)` method instantiates an anonymous `Error` class that only overrides the `getMessage` method, the `equals` method returns true the, Errors are equal. 
So `error.equals(faulty) shouldBe false` yields true and the test fails in the current implementation.
And the `AlternativeLaws` fail as well (I'm still not sure why). 

However, if I add an `override equals` method to the `Error` trait, everything seems fine. 

```scala
trait Error extends Exception {
  def canEqual(other: Any): Boolean = other.isInstanceOf[Error]
  override def hashCode(): Int = (super.hashCode, getMessage).##
  override def equals(other: Any): Boolean = other match {
    case that: Error => that.canEqual(this) && that.getMessage == this.getMessage
    case _ => false
  }
}
```

@vkostyukov what do you think about this solution? Otherwise we should pass the `message` to the constructor of the `Exception` and/or `Throwable` which seems to complicate the instantiation of our custom Error subclasses (but gives us the Java equals() method).

vkostyukov:
Thanks @DieBauer! 

I wonder if we should provide `Eq[Error]` instead to make applicative laws happy.

DieBauer:
What would be the benefits of that approach over the `equals` method in the trait?

As far as I can see, if we want to introduce an `implicit eqError` method, we also need an `unapply` method in the Error object, only for supporting equivalence of the Error trait. 

The subclasses are case-classes, so they have a proper equals method implemented. 

Let me know what you think, @vkostyukov!

vkostyukov:
Thanks @DieBauer! I think you're right this doesn't help a lot in this case. Although, I think we should add `Eq[Error]` and `Show[Error]` regardless (perhaps as a separate step).

Looking at the usages of `Error.apply`, we only use it in Argonaut since it doesn't have its own exception type for decoding failures. I think we don't have to stick with `Error.apply` in that case and can easily return something generic as `Exception(msg)`. One way or another it will be wrapped with `Error.NotParsed` at the end of the day.

With that said, how do you feel about removing anonymous instantiation from `Error.apply` and use just `Exception` in Argonaut?

core/src/main/scala/io/finch/Error.scala
```diff
@@ -1,28 +1,44 @@
...
-  import items._
+object Error {
 
   def apply(message: String): Error = new Error {
     override def getMessage: String = message
   }
 
+  object RequestErrors {
+    @deprecated("Use Multiple instead", "0.11")
+    def unapply(e: Throwable): Option[Seq[Throwable]] = Some(Seq(e))
```
vkostyukov: I'm not sure this is correct. What we want to do here is to make sure current users' code is still working. So ideally. `case RequestErros(seq)` should be backed by `Multiple(nel)`.

DieBauer: As far as I understand, if someone is using a pattern match on `RequestErrors(_)` it returns the `Seq[Throwable]` that was wrapped in it (and it will show a deprecated warning) which will still be the type of Throwable right? otherwise the signature changes -> no backward compatibility. 

Should something like this suffice then?
```scala
e match {
      case Multiple(err) => Some(err.toList)
      case t: Throwable => Some(Seq(t))
      case _ => None
    }
```

or am I misunderstanding your "should be backed by Multiple(nel)"?

vkostyukov: I think we need to return `None` for the `t: Throwable` case. Otherwise your reasoning about this is right.

DieBauer: and what if there is a different Error wrapped in there?
Do we need to catch `case e: Error => Some(Seq(e))`?

DieBauer:
> With that said, how do you feel about removing anonymous instantiation from Error.apply and use just Exception in Argonaut?

Right now I've added one anonymous `Error` instantiation to the Endpoint class.

```scala
      def collect(e: Throwable): NonEmptyList[Error] = e match {
        case Error.Multiple(errors) => errors
        case rest => NonEmptyList.of(
          Error(Try(rest.getMessage).getOrElse("Exception message was null"))
        )
      }
```
How would we tackle that? At this moment I can't see how to wrap that Throwable to conform the method signature. 
(btw this method is also the reason the applicativelaws failed in first place, I introduced the anonymous error class here).

core/src/main/scala/io/finch/Error.scala
DieBauer: Ah wait, I was confused for a 'generic' Error unapply. But this should only handle Multiple then. Don't mind my previous comment..

vkostyukov:
This looks great! Thanks again @DieBauer for the hard work and been so patient with all my nitpicks!

Issue #528, opened 2016-02-17
Add new package finch-generic


vkostyukov:
I think it makes sense to provide `finch-generic` covering generic derivations on endpoints in Finch. As for today we can build endpoints from query string params (we used to build request readers in the same way):

``` scala
val e = Endpoint.derive[User].fromParams
```

I think I'm in favor of Circe-style API for this:

``` scala
import io.finch.generic._

val e: Endpoint[User] = deriveFromParams[User] // or simiar
```

Would be also nice to support derivation from path segments.

teodimoff:
+1

PR #663, opened 2016-11-07
Decoding overhaul

### LINKED ISSUE #562: Refactor Decode type-class

vkostyukov:
Fixes #562.

This PR revisits the decoding machinery in Finch, which had some broken windows waiting to be fixed (some of them are fixed in PR).

1. `Buf` to `String` is not absolutely necessary when decoding an HTTP payload, although our main decoding abstraction takes `String` as an input (see #511)
2. Generic derivation of endpoints was broken since day one. Now it's fixed and introduces a path forward to a standalone `finch-generic` package (see #528)
3. Picking the right decoder at compile time was a matter a luck and a proper ordering of implicits. Now it's as explicit as it could be.

Basically, this RB restructures the `Decode` types class such that it also takes a content-type (a type-level string) and a charset. This makes the decoding of HTTP payloads quite explicit (not wired it yet). An original `Decode` (operating on strings) was renamed to `DecodeEntity` so it still can be used to decode headers, params, etc (`Decode.Aux[]` on the other hand, designed for payloads).

Everything was possible with a single API change that most of the users shouldn't notice: `body` now `Endpoint[Buf]` (previously `Endpoint[String]`). That said, `body.as[User]` should work as expected.

I plan to 
1. wire in the new API as a next step (also revisit the API around `.as`) and then
2. eliminate extra to-string conversion, which should give some nice performance improvements.

I want to make sure we carefully review each step (updated API, perf improvements) and measure its impact so I decided to minimize this change and only provide the building blocks we would use later.


yanana:
Great work!

PR #671, opened 2016-11-19
Remove String conversion in decoders.

rpless:
Proposed fix for #511. The only thing I was unsure about was whether to use `Buf.ByteArray.Owned` or `Buf.ByteArray.Shared` for the libraries that can convert `Array[Byte]`. I wound up using `Shared`, but I'm not sure this is necessary.

### DIFF
--- a/argonaut/src/main/scala/io/finch/argonaut/Decoders.scala
+++ b/argonaut/src/main/scala/io/finch/argonaut/Decoders.scala
@@ -1,9 +1,11 @@
 package io.finch.argonaut
 
 import argonaut.{CursorHistory, DecodeJson, Json}
+import com.twitter.finagle.netty3.ChannelBufferBuf
 import com.twitter.util.{Return, Throw, Try}
 import io.finch._
 import io.finch.internal.BufText
+import java.nio.charset.StandardCharsets
 import jawn.Parser
 import jawn.support.argonaut.Parser._
 import scala.util.{Failure, Success}
@@ -16,12 +18,12 @@ trait Decoders {
   implicit def decodeArgonaut[A](implicit d: DecodeJson[A]): Decode.Json[A] =
     Decode.json { (b, cs) =>
       val err: (String, CursorHistory) => Try[A] = { (str, hist) => Throw[A](Error(str)) }
-
-      // TODO: Eliminate toString conversion
-      // See https://github.com/finagle/finch/issues/511
-      // Jawn can parse from ByteBuffer's if they represent UTF-8 strings.
-      // We can check the charset here and do parsing w/o extra to-string conversion.
-      Parser.parseFromString[Json](BufText.extract(b, cs))(facade) match {
+      val attemptJson = cs match {
+        case StandardCharsets.UTF_8 =>
+          Parser.parseFromByteBuffer[Json](ChannelBufferBuf.Owned.extract(b).toByteBuffer())(facade)
+        case _ => Parser.parseFromString[Json](BufText.extract(b, cs))(facade)
+      }
+      attemptJson match {
         case Success(value) => d.decodeJson(value).fold[Try[A]](err, Return(_))
         case Failure(error) => Throw[A](Error(error.getMessage))
       }
--- a/circe/src/main/scala/io/finch/circe/Decoders.scala
+++ b/circe/src/main/scala/io/finch/circe/Decoders.scala
@@ -1,26 +1,25 @@
 package io.finch.circe
 
 import cats.syntax.show._
+import com.twitter.finagle.netty3.ChannelBufferBuf
 import com.twitter.util.{Return, Throw, Try}
 import io.circe.Decoder
-import io.circe.jawn.decode
+import io.circe.jawn._
 import io.finch.{Decode, Error}
 import io.finch.internal.BufText
+import java.nio.charset.StandardCharsets
 
 trait Decoders {
 
   /**
    * Maps a Circe's [[Decoder]] to Finch's [[Decode]].
    */
-  implicit def decodeCirce[A: Decoder]: Decode.Json[A] = Decode.json((b, cs) =>
-
-    // TODO: Eliminate toString conversion
-    // See https://github.com/finagle/finch/issues/511
-    // Jawn can parse from ByteBuffer's if they represent UTF-8 strings.
-    // We can check the charset here and do parsing w/o extra to-string conversion.
-    decode[A](BufText.extract(b, cs)).fold[Try[A]](
-      error => Throw[A](Error(error.show)),
-      value => Return(value)
-    )
-  )
+  implicit def decodeCirce[A: Decoder]: Decode.Json[A] = Decode.json { (b, cs) =>
+    val attemptJson = cs match {
+      case StandardCharsets.UTF_8 =>
+        parseByteBuffer(ChannelBufferBuf.Owned.extract(b).toByteBuffer()).right.flatMap(_.as[A])
+      case _ => decode[A](BufText.extract(b, cs))
+    }
+    attemptJson.fold[Try[A]](error => Throw[A](Error(error.show)), value => Return(value))
+  }
 }
--- a/jackson/src/main/scala/io/finch/jackson/package.scala
+++ b/jackson/src/main/scala/io/finch/jackson/package.scala
@@ -1,21 +1,20 @@
 package io.finch
 
-import scala.reflect.ClassTag
-
 import com.fasterxml.jackson.databind.ObjectMapper
+import com.twitter.finagle.netty3.ChannelBufferBuf
 import com.twitter.util.Try
 import io.finch.internal.BufText
+import scala.reflect.ClassTag
 
 package object jackson {
 
   implicit def decodeJackson[A](implicit
     mapper: ObjectMapper, ct: ClassTag[A]
-  ): Decode.Json[A] = Decode.json((b, cs) =>
-    // TODO: Eliminate toString conversion
-    // See https://github.com/finagle/finch/issues/511
-    // Jackson can parse from byte[] automatically detecting the encoding.
-    Try(mapper.readValue(BufText.extract(b, cs), ct.runtimeClass.asInstanceOf[Class[A]]))
-  )
+  ): Decode.Json[A] = Decode.json { (b, cs) =>
+    val buf = ChannelBufferBuf.Owned.extract(b)
+    if (buf.hasArray) Try(mapper.readValue(buf.array(), 0, buf.readableBytes(), ct.runtimeClass.asInstanceOf[Class[A]]))
+    else Try(mapper.readValue(buf.toString(cs), ct.runtimeClass.asInstanceOf[Class[A]]))
+  }
 
   implicit def encodeJackson[A](implicit mapper: ObjectMapper): Encode.Json[A] =
     Encode.json((a, cs) => BufText(mapper.writeValueAsString(a), cs))
--- a/playjson/src/main/scala/io/finch/playjson/package.scala
+++ b/playjson/src/main/scala/io/finch/playjson/package.scala
@@ -1,7 +1,8 @@
 package io.finch
 
+import com.twitter.finagle.netty3.ChannelBufferBuf
 import com.twitter.util.Try
-import io.finch.internal.BufText
+import io.finch.internal.{extractBytesFromArrayBackedBuf, BufText}
 import play.api.libs.json._
 
 package object playjson {
@@ -11,10 +12,11 @@ package object playjson {
    * @tparam A the type of the data to decode into
    */
   implicit def decodePlayJson[A](implicit reads: Reads[A]): Decode.Json[A] =
-    // TODO: Eliminate toString conversion
-    // See https://github.com/finagle/finch/issues/511
-    // PlayJson can parse from byte[] automatically detecting the charset.
-    Decode.json((b, cs) => Try(Json.parse(BufText.extract(b, cs)).as[A]))
+    Decode.json { (b, cs) =>
+      val buf = ChannelBufferBuf.Owned.extract(b)
+      if (buf.hasArray) Try(Json.parse(extractBytesFromArrayBackedBuf(buf)).as[A])
+      else Try(Json.parse(buf.toString(cs)).as[A])
+    }
 
   /**
    * @param writes Play JSON `Writes` to use for encoding
--- a/sprayjson/src/main/scala/io/finch/sprayjson/package.scala
+++ b/sprayjson/src/main/scala/io/finch/sprayjson/package.scala
@@ -1,20 +1,27 @@
 package io.finch
 
+import com.twitter.finagle.netty3.ChannelBufferBuf
 import com.twitter.util.Try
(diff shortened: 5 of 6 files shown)


vkostyukov:
@rpless Thanks a lot for looking into this!

Some suggestions:

 - Instead of `Buf.ByteArray.Shared` (which always copy), let's use [this trick](https://github.com/finagle/finch/blob/master/core/src/main/scala/io/finch/Endpoints.scala#L287-L292) that allows retrieving a byte array from Netty's channel buffer. This heavily depends on the underlying implementation, but I'm fine with that for now.
 - Let run some benchmarks (both JMH and wrk) to see how this affects throughput (I'm excited to see the numbers!)

vkostyukov:
Trying to be more specific here.

If you need `ByteBuffer`, the cheapest way (at least for Netty 3) to get it is:

```scala
val byteBuffer = ChannelBufferBuf.Owned.extract(buf).toByteBuffer()
```

If you need `Array[Byte]`, the cheapest way to get it is:

```scala
val channelBuffer = ChannelBufferBuf.Owned.extract(buf)
val (byteArray, offset, length) = 
  // assert channelBuffer.hasArray()
  (channelBuffer.array(), channelBuffer.readerIndex(), channelBuffer.readableBytes()) 
```

rpless:
I ran the wrk benchmarks for Finch + circe. The results are [here](https://gist.github.com/rpless/7149d0f110e25a9885be4fa8582c6a55).
As for jmh benchmarks, unless I missed something, it doesn't look like we have ones that exercise the the json decoders anymore.

vkostyukov:
Nice! 

@rpless Do you mind keep the server running and run `wrk` for 3 times. Usually, it takes some time for JVM to warmup so using the results from 3rd run seems like the closest we can get to the steady state.

argonaut/src/main/scala/io/finch/argonaut/Decoders.scala
```diff
@@ -1,6 +1,8 @@
 package io.finch.argonaut
 
 import argonaut.{CursorHistory, DecodeJson, Json}
+import com.twitter.finagle.netty3.ChannelBufferBuf
+import com.twitter.io.Charsets
```
vkostyukov: Please, use `StandardCharsets` from JDK. `c.t.i.Charsets` are deprecated in the most recent Finagle release.

circe/src/main/scala/io/finch/circe/Decoders.scala
```diff
@@ -12,15 +14,11 @@ trait Decoders {
...
-    // Jawn can parse from ByteBuffer's if they represent UTF-8 strings.
-    // We can check the charset here and do parsing w/o extra to-string conversion.
-    decode[A](BufText.extract(b, cs)).fold[Try[A]](
-      error => Throw[A](Error(error.show)),
-      value => Return(value)
-    )
-  )
+  implicit def decodeCirce[A: Decoder]: Decode.Json[A] = Decode.json({ (b, cs) =>
+    val attemptJson = cs match {
+      case Charsets.Utf8 => parseByteBuffer(ChannelBufferBuf.Owned.extract(b).toByteBuffer()).right.flatMap(_.as[A])
```
vkostyukov: Is scalacheck happy with that line length?

vkostyukov: Nah, it's fine. Thanks!

rpless:
Updated benckmarks, results after 3 runs:
## Without String Conversion (this PR)
```
$ wrk -t4 -c24 -d30s -s examples/src/main/scala/io/finch/wrk/wrk.lua http://localhost:8081/
Running 30s test @ http://localhost:8081/
  4 threads and 24 connections
  Thread Stats   Avg      Stdev     Max   +/- Stdev
    Latency     1.68ms    3.67ms  52.43ms   90.18%
    Req/Sec    12.20k     2.26k   23.58k    67.25%
  1458406 requests in 30.09s, 137.69MB read
Requests/sec:  48471.67
Transfer/sec:      4.58MB
```

## With String Conversion (Current Finch)
```
$ wrk -t4 -c24 -d30s -s examples/src/main/scala/io/finch/wrk/wrk.lua http://localhost:8081/
Running 30s test @ http://localhost:8081/
  4 threads and 24 connections
  Thread Stats   Avg      Stdev     Max   +/- Stdev
    Latency     1.62ms    3.22ms  62.64ms   89.91%
    Req/Sec    10.84k     2.44k   22.20k    68.33%
  1296147 requests in 30.10s, 122.37MB read
Requests/sec:  43067.06
Transfer/sec:      4.07MB
```

vkostyukov:
Looks like **12.5%** improvement in the throughput! Sweet!

playjson/src/main/scala/io/finch/playjson/package.scala
```diff
@@ -11,10 +13,11 @@ package object playjson {
...
   implicit def decodePlayJson[A](implicit reads: Reads[A]): Decode.Json[A] =
-    // TODO: Eliminate toString conversion
-    // See https://github.com/finagle/finch/issues/511
-    // PlayJson can parse from byte[] automatically detecting the charset.
-    Decode.json((b, cs) => Try(Json.parse(BufText.extract(b, cs)).as[A]))
+    Decode.json({ (b, cs) => Try(Json.parse(Buf.ByteArray.Shared.extract(b)).as[A])
+      val buf = ChannelBufferBuf.Owned.extract(b)
+      if (buf.hasArray) Try(Json.parse(new String(buf.array(), 0, buf.readableBytes(), cs)).as[A])
+      else Try(Json.parse(buf.toString(cs)).as[A])
+    })
```
clhodapp: It looks like you might be parsing twice here by accident?

rpless: No accident. Its based on this [trick](https://github.com/finagle/finch/blob/master/core/src/main/scala/io/finch/Endpoints.scala#L287-L292) that @vkostyukov mentioned in earlier. There may be a better to express this (i.e. only having on call to `Json.parse` once and having the `if`/ `else` only return a string).

rpless: @clhodapp I played around with it a little. Do you think this version makes the intent more clear?
```scala
val buf = ChannelBufferBuf.Owned.extract(b)
val bufAsString = Try(
  if (buf.hasArray) new String(buf.array(), 0, buf.readableBytes(), cs)
  else buf.toString(cs)
)
bufAsString.map(Json.parse(_).as[A])
```

vkostyukov:
I think Chris is right. We call Json.parse two times and drop the first
result on the floor.

playjson/src/main/scala/io/finch/playjson/package.scala
```diff
@@ -11,10 +13,11 @@ package object playjson {
    * @tparam A the type of the data to decode into
    */
   implicit def decodePlayJson[A](implicit reads: Reads[A]): Decode.Json[A] =
-    // TODO: Eliminate toString conversion
-    // See https://github.com/finagle/finch/issues/511
-    // PlayJson can parse from byte[] automatically detecting the charset.
-    Decode.json((b, cs) => Try(Json.parse(BufText.extract(b, cs)).as[A]))
+    Decode.json({ (b, cs) =>
+      val buf = ChannelBufferBuf.Owned.extract(b)
+      if (buf.hasArray) Try(Json.parse(new String(buf.array(), 0, buf.readableBytes(), cs)).as[A])
```
vkostyukov: I think, `Json.parse` can parse `byte[]` directly. No need for `new String`.

sprayjson/src/main/scala/io/finch/sprayjson/package.scala
```diff
@@ -1,20 +1,27 @@
...
   implicit def decodeSpray[A](implicit format: JsonFormat[A]): Decode.Json[A] =
-    // TODO: Eliminate toString conversion
-    // See https://github.com/finagle/finch/issues/511
-    // SprayJson can parse from byte[] if it represents a UTF-8 string.
-    Decode.json((b, cs) => Try(BufText.extract(b, cs).parseJson.convertTo[A]))
+    Decode.json({ (b, cs) =>
+      cs match {
+        case StandardCharsets.UTF_8 =>
+          val buf = ChannelBufferBuf.Owned.extract(b)
+          if (buf.hasArray) Try(JsonParser(new String(buf.array(), 0, buf.readableBytes(), cs)).convertTo[A])
```
vkostyukov: No need for `new String`.

circe/src/main/scala/io/finch/circe/Decoders.scala
```diff
@@ -1,26 +1,25 @@
...
-    // TODO: Eliminate toString conversion
-    // See https://github.com/finagle/finch/issues/511
-    // Jawn can parse from ByteBuffer's if they represent UTF-8 strings.
-    // We can check the charset here and do parsing w/o extra to-string conversion.
-    decode[A](BufText.extract(b, cs)).fold[Try[A]](
-      error => Throw[A](Error(error.show)),
-      value => Return(value)
-    )
-  )
+  implicit def decodeCirce[A: Decoder]: Decode.Json[A] = Decode.json({ (b, cs) =>
```
vkostyukov: Let's remove wrapping `()`. Just `Decode.json { ... }` is fine.

jackson/src/main/scala/io/finch/jackson/package.scala
```diff
@@ -1,21 +1,20 @@
...
 
   implicit def decodeJackson[A](implicit
     mapper: ObjectMapper, ct: ClassTag[A]
-  ): Decode.Json[A] = Decode.json((b, cs) =>
-    // TODO: Eliminate toString conversion
-    // See https://github.com/finagle/finch/issues/511
-    // Jackson can parse from byte[] automatically detecting the encoding.
-    Try(mapper.readValue(BufText.extract(b, cs), ct.runtimeClass.asInstanceOf[Class[A]]))
-  )
+  ): Decode.Json[A] = Decode.json({ (b, cs) =>
```
vkostyukov: Please, remove `()`: `Decode.json { ... }`.

playjson/src/main/scala/io/finch/playjson/package.scala
```diff
@@ -11,10 +13,11 @@ package object playjson {
    * @tparam A the type of the data to decode into
    */
   implicit def decodePlayJson[A](implicit reads: Reads[A]): Decode.Json[A] =
-    // TODO: Eliminate toString conversion
-    // See https://github.com/finagle/finch/issues/511
-    // PlayJson can parse from byte[] automatically detecting the charset.
-    Decode.json((b, cs) => Try(Json.parse(BufText.extract(b, cs)).as[A]))
+    Decode.json({ (b, cs) =>
```
vkostyukov: Please, remove wrapping `()`: `Decode.json { ... }`.

sprayjson/src/main/scala/io/finch/sprayjson/package.scala
```diff
@@ -1,20 +1,27 @@
...
   /**
     * @param format spray-json support for convert JSON val to specific type object
     * @tparam A the type of the data to decode into
     */
   implicit def decodeSpray[A](implicit format: JsonFormat[A]): Decode.Json[A] =
-    // TODO: Eliminate toString conversion
-    // See https://github.com/finagle/finch/issues/511
-    // SprayJson can parse from byte[] if it represents a UTF-8 string.
-    Decode.json((b, cs) => Try(BufText.extract(b, cs).parseJson.convertTo[A]))
+    Decode.json({ (b, cs) =>
```
vkostyukov: Please, remove wrapping `()`: `Decode.json { ... }`.

playjson/src/main/scala/io/finch/playjson/package.scala
rpless: `Json.Parse` can take a `byte[]`, but when I convert that line to `Try(Json.parse(buf.array()).as[A])` it gives me the following in the 
```
Throw(spray.json.JsonParser$ParsingException: Unexpected character '\u0000'
```
I think I'm missing something about how the ChannelBufferBuf holds onto the underlying array.

sprayjson/src/main/scala/io/finch/sprayjson/package.scala
rpless: A similar issue exists here. The tests raise this error.
```
Throw(com.fasterxml.jackson.core.JsonParseException: Illegal character ((CTRL-CHAR, code 0)): only regular white space (\r, \n, \t) is allowed between tokens
```

rpless:
Ok, it looks like using `slice` like this works: `buf.array().slice(0, buf.readableBytes())`. Does that make sense?

vkostyukov:
Slice works but it allocates a new array. You should be able to bypass
readableBytes to the parse function.

rpless:
Yeah, I'm not a fan of the new array allocation, but it seems like neither Spray Json nor Play Json expose a parsing function that takes the number of bytes to read:
- https://www.playframework.com/documentation/2.5.x/api/scala/index.html#play.api.libs.json.Json$
- https://github.com/spray/spray-json/blob/master/src/main/scala/spray/json/JsonParser.scala#L28
  (and also https://github.com/spray/spray-json/blob/master/src/main/scala/spray/json/JsonParser.scala#L237)

vkostyukov:
Play takes InputStream though. Can we wrap our byte array with
ByteArrayInputStream and specify the length there?

For Spray, let's do System.arraycopy it's still better than 'new String'.

rpless:
Can do for Spray, but it looks like I gave you the wrong docs for Play Json. We're currently on [2.3.x](https://www.playframework.com/documentation/2.3.x/api/scala/index.html#play.api.libs.json.Json$) which does not provide a parse with InputStream. So we can either bump the Play version or System.arraycopy.

vkostyukov:
Arraycopy works for me. Let's update it later.

playjson/src/main/scala/io/finch/playjson/package.scala
```diff
@@ -11,10 +14,16 @@ package object playjson {
    * @tparam A the type of the data to decode into
    */
   implicit def decodePlayJson[A](implicit reads: Reads[A]): Decode.Json[A] =
-    // TODO: Eliminate toString conversion
-    // See https://github.com/finagle/finch/issues/511
-    // PlayJson can parse from byte[] automatically detecting the charset.
-    Decode.json((b, cs) => Try(Json.parse(BufText.extract(b, cs)).as[A]))
+    Decode.json { (b, cs) =>
+      val buf = ChannelBufferBuf.Owned.extract(b)
+      if (buf.hasArray) {
```
vkostyukov: Sorry to be a burden, but can we extract that into a function so we can reuse it?

Also, let's check that if `array.length` equals `readableBytes` we can skip copying.

rpless: Its not a burden. Any particular place we should extract it to? I was thinking `io.finch.Decode`.

rpless: Or maybe the `internal` package object?

vkostyukov:
Thanks again @rpless! Merging this.

PR #674, opened 2016-11-20
Add an instance of DecodePath[String]

clhodapp:
I know there's `string`, but this seems like it should exist.


vkostyukov:
Thanks @clhodapp! I think it makes sense.

Can you also add some test?

vkostyukov:
Hey @clhodapp! This is already in place as was required to support the `path[Stirng]` machinery. So I'm closing this.

clhodapp:
Great, thanks!

PR #675, opened 2016-11-21
Escape JSON strings while encoding exceptions

akozhemiakin:
Current implementation of exceptions to JSON encoder doesn't take into account that exception message may contain special characters that must be escaped (double quotes for example). I borrowed a little piece of code from Circe Printer to fix it.

### DIFF
--- a/core/src/main/scala/io/finch/Encode.scala
+++ b/core/src/main/scala/io/finch/Encode.scala
@@ -2,7 +2,7 @@ package io.finch
 
 import cats.Show
 import com.twitter.io.Buf
-import io.finch.internal.BufText
+import io.finch.internal.{BufText, JsonUtils}
 import java.nio.charset.Charset
 
 /**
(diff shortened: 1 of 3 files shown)


vkostyukov:
This is very interesting @akozhemiakin! I already had a second thought about providing those `Encode[Exception]` instances for JSON and plain text in the core package. Maybe it's time to revisit it. Let me think about that a bit and I will get back to you.

My current thinking is that it's `io.finch.circe` that has to provide this instance using Circe APIs, while finch-core should do a very basic thing (return 500?). This also relates to `finch-sse`, where it's somewhat not clear how to define a `Encode[Exception]`.

akozhemiakin:
Yeah, that sounds perfectly reasonable. Actually my first guess was that it is a Circe issue, because I was sure that Circe is responsible for encoding. But than I was quite surprised to find out that there is that simple implementation in core responsible for encoding exceptions. So I tried to kind of fix it in place) So for me it sounds really good to move all but Application.Text encoders from core to corresponding sub-projects (if I do not miss something, for now that "all" encoders are actually limited to to this json one)

vkostyukov:
Yeah, I think `Encode.encodeExceptionAsJson` should not be defined finch-core. Instead, each supported JSON library should supply that instance. 

@akozhemiakin Any chance you want to volunteer and open such PR?

akozhemiakin:
Sure, I will be glad to take a shot.
So I close this PR for now and will create a new one soon.

PR #680, opened 2016-11-22
Refactor the way how JSON Encoders for Exceptions are provided

akozhemiakin:
This PR is a successor of #675.

The main goal of this commit is to delegate the process of transforming Exceptions into JSON to the JSON libraries instead of handling it in the core package. Thus JSON Encoder for Exceptions is removed from the core package, but appropriate instances are provided in each of subprojects responsible for integration with the corresponding JSON libraries.

The motivation is that transforming even very simple data structures into the valid JSON is not so trivial and at least involves some logic of escaping strings before inserting them into the JSON document. Also, it seems obvious, that in the situation when exception need to be encoded into JSON, other data should be JSON encoded as well. and user will end up using some JSON library anyway. So, delegating this process (of encoding Exceptions) to JSON libraries does not force users to use one of them, because they already do. On the other hand it corresponds to the SoC principle and make the code better.

### DIFF
--- a/circe/src/main/scala/io/finch/circe/Encoders.scala
+++ b/circe/src/main/scala/io/finch/circe/Encoders.scala
@@ -13,4 +13,8 @@ trait Encoders {
    */
   implicit def encodeCirce[A](implicit e: Encoder[A]): Encode.Json[A] =
     Encode.json((a, cs) => BufText(print(e(a)), cs))
+
+  implicit def encodeExceptionCirce[A <: Exception]: Encoder[A] = new Encoder[A] {
+    override def apply(e: A): Json = Json.obj(("message", Json.fromString(e.getMessage)))
+  }
 }
--- a/playjson/src/main/scala/io/finch/playjson/package.scala
+++ b/playjson/src/main/scala/io/finch/playjson/package.scala
@@ -24,4 +24,8 @@ package object playjson {
    */
   implicit def encodePlayJson[A](implicit writes: Writes[A]): Encode.Json[A] =
     Encode.json((a, cs) => BufText(Json.stringify(Json.toJson(a)), cs))
+
+  implicit val encodeExceptionPlayJson: Writes[Exception] = new Writes[Exception] {
+    override def writes(e: Exception): JsValue = Json.obj("message" -> e.getMessage)
+  }
 }
--- a/sprayjson/src/main/scala/io/finch/sprayjson/package.scala
+++ b/sprayjson/src/main/scala/io/finch/sprayjson/package.scala
@@ -12,7 +12,7 @@ package object sprayjson {
     * @param format spray-json support for convert JSON val to specific type object
     * @tparam A the type of the data to decode into
     */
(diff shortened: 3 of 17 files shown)


circe/src/main/scala/io/finch/circe/Encoders.scala
```diff
@@ -13,4 +13,9 @@ trait Encoders {
    */
   implicit def encodeCirce[A](implicit e: Encoder[A]): Encode.Json[A] =
     Encode.json((a, cs) => BufText(print(e(a)), cs))
+
+  implicit val encodeExceptionCirce: Encode.Json[Exception] =
```
vkostyukov: Fun fact: we don't have to provide Finch's `Encode.Json[Exception]` here. Instead, we can define Circe's `Encoder[Exception]` and it will be implicitly converted (the conversion is defined in this very file) into Finch's version when needed.

playjson/src/main/scala/io/finch/playjson/package.scala
```diff
@@ -24,4 +24,9 @@ package object playjson {
    */
   implicit def encodePlayJson[A](implicit writes: Writes[A]): Encode.Json[A] =
     Encode.json((a, cs) => BufText(Json.stringify(Json.toJson(a)), cs))
+
+  implicit val encodeExceptionPlayJson: Encode.Json[Exception] =
```
vkostyukov: Can we define `Writes[Exception]` instead?

sprayjson/src/main/scala/io/finch/sprayjson/package.scala
```diff
@@ -29,4 +29,9 @@ package object sprayjson {
     */
   implicit def encodeSpray[A](implicit format: JsonFormat[A]): Encode.Json[A] =
     Encode.json((a, cs) => BufText(a.toJson.prettyPrint, cs))
+
+  implicit val encodeExceptionSpray: Encode.Json[Exception] =
```
vkostyukov: Let's define `JsonFormat[Exception]` instead.

akozhemiakin: Maybe both in the encodeSpray and the encodeExceptionSpray it is more reasonable to use JsonWriter instead of JsonFormat? Because JsonFormat is JsonWriter + JsonReader. And actually we do not need to read here. I start thinking about this replacement because I do not want to define JsonFormat[Exception] with read method throwing something like UnsupportedOperationException.

akozhemiakin: And to say more, in the decodeSpray I think we can safely replace JsonFormat with JsonReader.

akozhemiakin:
Yeah, my bad, completely forgot about Argonaut.
Ok, I believe that I fixed all that you mentioned. Also take a look at my comment on sprayjson file: in the last commit I did it.

vkostyukov:
I like it! Thanks @akozhemiakin!

PR #681, opened 2016-11-22
Update circe and other versions

travisbrown:


vkostyukov:
Thanks @travisbrown! This looks good to me.

Issue #683, opened 2016-11-27
Polymorphic Encode[Exception]


vkostyukov:
As per #680, Finch now defines `Encode.Json[Exception]` instances in the corresponding JSON packages. This means any Finch app w/o explicitly defined encode instance for `Exception` will compile if the `io.finch.$jsonLibrary` import present. Although, it's still not so friendly [for users doing anything but JSON](https://gitter.im/finagle/finch?at=58386d7eb9016e42149a7d3a).

I propose we define a polymorphic `Encode.Aux[Exception, CT]` instance that converts exceptions into 500 responses with no payload.

This way we define a new default behavior for endpoints (which I think makes sense):

1. `Endpoint[A]` could be compiled if at least `Encode.Aux[A, CT]` is defined.
2. Unhandled exceptions are converted into 500 responses with no payload.
3. Handled exceptions, for which `Encode.Aux[Exception, CT]` is not defined are converted into 500 responses with no payload (previously resulted in compile error).
3. Handled exceptions are encoded according to their `Encode.Aux[Exception, CT]` instances.

vkostyukov:
Done in #684.

PR #684, opened 2016-11-28
Polymorphic Encode[Exception]

vkostyukov:
See details on #683.


core/src/main/scala/io/finch/Encode.scala
```diff
@@ -21,46 +21,57 @@ trait LowPriorityEncodeInstances {
...
 }
+
+object Encode extends HighPriorityEncodeInstances {
+
+  /**
+   * Returns a [[Encode]] instance for a given type (with required content type).
+   */
+  @inline final def apply[A, CT <: String](implicit e: Aux[A, CT]): Aux[A, CT] = e
+
+  implicit val encodeExceptionAsTextPlain: Text[Exception] =
```
vkostyukov: Eventually, I'd like us to get rid of these as well.

PR #685, opened 2016-11-28
add content types

crispywalrus:
this adds the the official apache promoted content types to allow for clients to distinguish between compact and binary protocols. This is of use to me and hopefully of use more generally.


muuki88:
@crispywalrus this looks interesting. In which way do you use these?

crispywalrus:
for some services inside a finagle environment where the thrift is transported over http rather than the normal socket connection. The other side of this conversation uses the content type to understand how to deserialize when compact vs binary are both supported.

PR #690, opened 2016-12-05
Add instances of Eq[Error] and Show[Error] in Error companion object.

haribageski:


core/src/main/scala/io/finch/Error.scala.orig
```diff
@@ -0,0 +1,121 @@
+package io.finch
```
vkostyukov: Can you remove it?

haribageski: Hi, that was not added by me.
And when I remove it I get a warning:
"Package names doesn't correspond to directories structure, this may cause problems with resolve to classes from this file.
Inspection for files with package statement which does not correspond to package structure"

vkostyukov: Just rebase your branch against the most recent master and you should be fine.

See [here](http://stackoverflow.com/questions/7929369/how-to-rebase-local-branch-with-remote-master) on how to do that.

haribageski: It's done.

vkostyukov:
Hey @haribageski! This still has some commits from master. Could you rebase that and override you branch? Presumably, an easiest way to do that would be just drop the branch and start a new one given that the diff is just a couple of lines of code.

haribageski:
Let's try one more time.
Where shell I add the corresponding test? There is no more ErrorSpec.

vkostyukov:
Let's omit the test for now. It's fine.

haribageski:
Done.

vkostyukov:
Thanks @haribageski! I just merged that in. Sorry it took me so long.

haribageski:
Thanks @vkostyukov! 
If you want me to write tests for the above let me know.

PR #695, opened 2016-12-12
Add new body endpoints

### LINKED ISSUE #666: Wire in the new API for Decode.Aux
Right now `body.as[User]` is pinned to `Application.Json`. We'd like to allow users to override that and pick the appropriate decoder at compile time.

vkostyukov:
This fixes #666.

This is a baby-step improvement that introduces a new way of decoding JSON/XML (anything with `Decode.Aux` instances) in Finch. Instead of using the `.as` machinery which became quite ambiguous already, this RP introduces a new API for building endpoints that read HTTP payloads.

```scala
def body[A, CT <: ContentType]: Endpoint[A]
def bodyOption[A, CT <: ContentType]: Endpoint[Option[A]]
```
The previous version `body.as[A]` is fixed to JSON and is still supported.

I was thinking back and forth on what kind of API would make sense here and decided to go with a very simple thing: just function with two type parameters. Here are some benefits of this approach vs. the previous one (`body.as[A]`).

1. It's less powerful.
2. It reduces the complexity/ambiguity of the `.as` API that might mean several things (decoding with `DecodeEntity` or converting from an `HList`).
3. It simpler so it involves fewer indirections (nested endpoints). Previously, `body` produces an endpoint (actually two nested endpoints) and then `.as` wraps it with one extra layer. Now `body[A, CT]` produces a flat endpoint. This means fewer allocations and better performance (see Performance section).

## Performance

1. 15% less running time and 20% fewer allocations for `body`.
2. 7% less running time and 15% fewer allocations for `bodyOption`.

`json(Option)` is before, `json(Option)2` is after.

```
[info] BodyBenchmark.json                       avgt    6  4824.580 ± 1205.444   ns/op
[info] BodyBenchmark.json:·gc.alloc.rate.norm   avgt    6  5896.004 ±  147.449    B/op
[info] BodyBenchmark.json2                      avgt    6  4179.209 ±  673.098   ns/op
[info] BodyBenchmark.json2:·gc.alloc.rate.norm  avgt    6  4936.004 ±   73.723    B/op

[info] BodyBenchmark.jsonOption                       avgt    6  4335.755 ±  150.928   ns/op
[info] BodyBenchmark.jsonOption:·gc.alloc.rate.norm   avgt    6  5712.004 ±    0.001    B/op
[info] BodyBenchmark.jsonOption2                      avgt    6  4050.681 ±  685.263   ns/op
[info] BodyBenchmark.jsonOption2:·gc.alloc.rate.norm  avgt    6  4940.004 ±   36.862    
```

## New Structure

This PR also starts a good tradition (I hope) and defines a new `Body` endpoint as an actual class in the `io.finch.endpoint` package. `Endpoints` already became quite big and it promoted a lot of design/implementation decisions that coupled lots of endpoints together by trade offing performance with code site.

At this point, I prefer we start extracting endpoints in their separate classes (similarly how [scodec does it](https://github.com/scodec/scodec/tree/series/1.10.x/shared/src/main/scala/scodec/codecs)) so they are not coupled to each other and could be evolved separately with performance/efficiency in mind.

### DIFF
--- a/core/src/main/scala/io/finch/Endpoints.scala
+++ b/core/src/main/scala/io/finch/Endpoints.scala
@@ -8,18 +8,16 @@ import com.twitter.finagle.netty3.ChannelBufferBuf
 import com.twitter.io.Buf
 import com.twitter.util.{Base64StringEncoder, Future, Try}
 import io.catbird.util.Rerunnable
+import io.finch.endpoint._
 import io.finch.internal._
 import java.util.UUID
+import scala.reflect.ClassTag
 import shapeless._
 
 /**
  * A collection of [[Endpoint]] combinators.
  */
 trait Endpoints {
-
-  private[this] val hnilFutureOutput: Rerunnable[Output[HNil]] = new Rerunnable[Output[HNil]] {
-    override val run = Future.value(Output.payload(HNil))
-  }
 
   type Endpoint0 = Endpoint[HNil]
   type Endpoint2[A, B] = Endpoint[A :: B :: HNil]
@@ -31,11 +29,11 @@ trait Endpoints {
   private[finch] class Matcher(s: String) extends Endpoint[HNil] {
     def apply(input: Input): Endpoint.Result[HNil] =
       input.headOption.flatMap {
-        case `s` => Some(input.drop(1) -> hnilFutureOutput)
+        case `s` => Some(input.drop(1) -> Rs.OutputHNil)
         case _ => None
       }
 
-    override def toString = s
+    override def toString: String = s
   }
 
   implicit def stringToMatcher(s: String): Endpoint0 = new Matcher(s)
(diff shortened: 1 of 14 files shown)


core/src/main/scala/io/finch/Endpoints.scala
```diff
@@ -469,9 +452,47 @@ trait Endpoints {
...
+  def jsonBodyOption[A: Decode.Json : ClassTag]: Endpoint[Option[A]] =
+    bodyOption[A, Application.Json]
+
+  /**
+   * Alias for `body[A, Text.Plain]`
+   */
+  def textBody[A: Decode.Text : ClassTag]: Endpoint[A] = body[A, Text.Plain]
+
+  /**
+   * Alias for `body[A, Text.Plain]`
```
DieBauer: s/body/bodyOption

vkostyukov: Ugh. Nice catch!

PR #700, opened 2016-12-14
handle io.finch.Errors

crispywalrus:
after asService...  io.finch.Errors and all io.finch.Error should be handled by default and translated into BadRequest. This adds that behavior for Errors.


vkostyukov:
Thanks @crispywalrus!

PR #702, opened 2016-12-15
Update Jackson to 2.8.5

vkostyukov:
Let's depend on circe-jackson28 and update our own Jackson dependency to be aligned with Finagle 6.40.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -11,12 +11,13 @@ lazy val utilVersion = "6.39.0"
 lazy val twitterServerVersion = "1.25.0"
 lazy val finagleOAuth2Version = "0.2.0"
 lazy val circeVersion = "0.6.1"
+lazy val circeJacksonVersion = "0.6.2"
 lazy val catbirdVersion = "0.9.0"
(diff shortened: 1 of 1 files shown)


travisbrown:
👍, but is there a reason to choose 2.8.4 over 2.8.5? I'm assuming they're compatible, but circe-jackson28 depends on the newer one, which will evict the older unless users specifically choose it.

vkostyukov:
Finagle depends on 2.8.4 so I assumed we want to use the same version here. I guess they should be compatible so let's make it 2.8.5.

vkostyukov:
So I updated the version but had to change the dependency itself. Turns out, `jackson-module-scala` isn't published for 2.8.5 and we used to depend on that. I don't see a strong reason for doing that since it's clearly user's concern which Jackson module they want to depend on.

Hope that sounds reasonable.

spockz:
On the other hand, it sounds also reasonable to me that a Scala framework would provide support for serialising scala case classes ootb.

PR #704, opened 2016-12-18
Fix reference to incorrect decode object

crisson:
The function signature and context implies that `DecodeEntity` was intended instead of `Decode`.


vkostyukov:
Thanks @crisson! Nice catch!

PR #706, opened 2016-12-20
Add D.A.Consortium as an adopter

yanana:
Add D.A.Consortium as an adopter. We are using finch in production environment for over a year :)


vkostyukov:
I'm glad to hear that! Thanks @yanana!

Issue #692, opened 2016-12-07
Add Endpoint.productWith


vkostyukov:
It will save some allocations and make it easier to use.

The idea is to also pass a function that collects `A, B` into `C` such that `product` could be defined in terms of `productWith(other)(Tuple2.apply)`.

vkostyukov:
Done in #709.

PR #709, opened 2016-12-26
[#692] Add productWith and replace `product` calls with a new one

ImLiar:
To optimize the number of allocations and save ecology with creating less instances. It means consuming less power when computing the result. 
Taking in mind how many adopters finch would have in a very near future this PR is clearly aimed to save the whole planet.

image


vkostyukov:
Thanks @ImLiar! This is awesome!

Issue #676, opened 2016-11-21
String-less encoding


vkostyukov:
#511 was a great success and improved throughput dramatically. Although, decoding large/medium size JSON objects isn't that common in HTTP APIs. The more common path is encoding. Right now we're using JSON pretty printers that print into `String` and then convert that into a `Buf` (`BufText`). This chain involves at least 2 copies (as [per implementation](https://github.com/finagle/finch/blob/master/core/src/main/scala/io/finch/internal/package.scala#L115-L117)). 

It would be an outstanding win in throughput if we could eliminate those extra copies and print directly into `ByteBuffer` or `Array[Byte]`.

This ticket (as well as #511) gives me a lot of confidence in the approach we picked in the very beginning: abstract over HTTP encoding and decoding. Given we take care about that part, there is so much we can do to handle that in a very optimal way.

ImLiar:
So if I understand it correctly, all supplied JSON libraries should support direct print into byte array to make it happen?

vkostyukov:
Yup! Being able to print directly into `Array[Byte]` (or similar) would be very nice. I think, I tried to look into that at some point, and I'm not sure I was able to find the proper API for that (yet).

I'm not sure how open @travisbrown would be for adding a binary printer implementation into Circe, but I somewhat believe that would give us a huge boost in the end-to-end benchmark.

vkostyukov:
Actually, I'm somewhat positive that it's possible to add a variant of [`jacksonPrint`](https://github.com/circe/circe-jackson/blob/master/shared/src/main/scala/io/circe/jackson/package.scala#L17-L26) into Circe (when Jackson serializer is used) that yields byte array (byte buffer) instead of a string. This should allow us to avoid some copies.

Similar approach could be taken for finch-jackson, but I wouldn't be so worried about it for now. Starting with Circe seems like a great idea.

travisbrown:
I'd definitely be happy to see this in circe, for both the default printer and the Jackson one. I'd be inclined to return `Array[Byte]` for the sake of simplicity unless there's a good argument for `ByteBuffer` instead.

travisbrown:
I went ahead and [added something](https://github.com/circe/circe/pull/537) like this for circe-core, and can do the same for circe-jackson (unless someone else wants to).

vkostyukov:
Some work is already in progress:

- circe-jackson: https://github.com/circe/circe-jackson/pull/11
- circe-core: https://github.com/circe/circe/pull/537

[...]

PR #717, opened 2017-01-12
Encode to ByteBuffer in Circe

vkostyukov:
As discussed in #676 (and implemented in [Circe 0.7.0-M2](https://github.com/circe/circe/releases/tag/v0.7.0-M2)), encoding should be performed in terms of bytes to avoid extra to-string conversion hence reduce allocations.

New benchmark confirms that printing to bytes directly cuts allocations in half and improves the throughput by ~15-20%.

```
circe-core (string)

[info] Benchmark                                               Mode  Cnt        Score        Error   Units
[info] ToServiceBenchmark.foos                                thrpt   20      463.981 ±     24.913   ops/s
[info] ToServiceBenchmark.foos:·gc.alloc.rate.norm            thrpt   20  5002135.084 ±   3252.704    B/op
[info] ToServiceBenchmark.ints                                thrpt   20     2118.919 ±     49.846   ops/s
[info] ToServiceBenchmark.ints:·gc.alloc.rate.norm            thrpt   20   873272.466 ±    106.930    B/op

circe-core (bytes)

[info] Benchmark                                               Mode  Cnt        Score        Error   Units
[info] ToServiceBenchmark.foos                                thrpt   20      523.730 ±     21.871   ops/s
[info] ToServiceBenchmark.foos:·gc.alloc.rate.norm            thrpt   20  2602833.269 ±     76.806    B/op
[info] ToServiceBenchmark.ints                                thrpt   20     2234.631 ±    111.156   ops/s
[info] ToServiceBenchmark.ints:·gc.alloc.rate.norm            thrpt   20   545837.968 ±  43806.910    B/op

circe-jackson (string)

[info] Benchmark                                               Mode  Cnt        Score        Error   Units
[info] ToServiceBenchmark.foos                                thrpt   20      420.095 ±     15.935   ops/s
[info] ToServiceBenchmark.foos:·gc.alloc.rate.norm            thrpt   20  5158946.227 ±   8930.739    B/op
[info] ToServiceBenchmark.ints                                thrpt   20     3398.087 ±    244.404   ops/s
[info] ToServiceBenchmark.ints:·gc.alloc.rate.norm            thrpt   20   691384.267 ±    114.019    B/op

circe-jackson (bytes)

[info] Benchmark                                               Mode  Cnt        Score        Error   Units
[info] ToServiceBenchmark.foos                                thrpt   20      475.034 ±     16.369   ops/s
[info] ToServiceBenchmark.foos:·gc.alloc.rate.norm            thrpt   20  3520947.809 ±  36299.420    B/op
[info] ToServiceBenchmark.ints                                thrpt   20     4046.914 ±    185.765   ops/s
[info] ToServiceBenchmark.ints:·gc.alloc.rate.norm            thrpt   20   342040.224 ±      0.011    B/op
```

### DIFF
--- a/benchmarks/src/main/scala/io/finch/benchmarks.scala
+++ b/benchmarks/src/main/scala/io/finch/benchmarks.scala
@@ -1,7 +1,9 @@
 package io.finch
 
+import com.twitter.finagle.Service
+import com.twitter.finagle.http.{Request, Response}
 import com.twitter.io.Buf
-import com.twitter.util.Future
+import com.twitter.util.{Await, Future}
 import io.finch.data.Foo
 import java.util.concurrent.TimeUnit
 import org.openjdk.jmh.annotations._
@@ -9,8 +11,8 @@ import shapeless._
 
 @BenchmarkMode(Array(Mode.AverageTime))
 @OutputTimeUnit(TimeUnit.NANOSECONDS)
-@Warmup(iterations = 3, time = 1, timeUnit = TimeUnit.SECONDS)
-@Measurement(iterations = 3, time = 1, timeUnit = TimeUnit.SECONDS)
+@Warmup(iterations = 5, time = 1, timeUnit = TimeUnit.SECONDS)
+@Measurement(iterations = 5, time = 1, timeUnit = TimeUnit.SECONDS)
 @Fork(2)
 abstract class FinchBenchmark {
   val postPayload: Input = Input.post("/").withBody[Text.Plain](Buf.Utf8("x" * 1024))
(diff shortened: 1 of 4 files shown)


vkostyukov:
@travisbrown Out of curiosity, any idea why printing int arrays/lists could be slower when targeting byte-buffers in circe-core? I'm wondering if there is anything we could do on a Circe side to improve it. I'm happy to file a ticket if you think it deserves an attention.

travisbrown:
@vkostyukov No, not of the top of my head. If you don't mind filing a ticket it'd be good to have.

travisbrown:
@vkostyukov I just put together [some simple benchmarks](https://github.com/circe/circe/pull/544) to try to track this down, but they're not very helpful. I don't really understand how you could see so much of a difference in that one case specifically, since they're both using the same `Appendable` interface.

vkostyukov:
@travisbrown I think the benchmark results are just volatile enough (at least on my machine) to not capture the difference precisely. Let me try to make it more stable and see if we can get those numbers aligned. I think be reducing the number of data set (reducing allocations) we should get fewer GC pauses and more predictable results.

PR #719, opened 2017-01-15
Custom getClass.getSimpleName logic for error handling

wennergr:
The JVM crashes with an InternalError in case a parsing
error is detected in a class that is neated inside two
objects

### DIFF
--- a/core/src/main/scala/io/finch/Error.scala
+++ b/core/src/main/scala/io/finch/Error.scala
@@ -75,9 +75,14 @@ object Error {
   final case class NotParsed(item: RequestItem, targetType: ClassTag[_], cause: Throwable)
     extends Error {
 
-    override def getMessage: String =
-      s"${item.description} cannot be converted to ${targetType.runtimeClass.getSimpleName}: " +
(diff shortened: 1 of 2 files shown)


core/src/main/scala/io/finch/Error.scala
```diff
@@ -75,9 +75,14 @@ object Error {
   final case class NotParsed(item: RequestItem, targetType: ClassTag[_], cause: Throwable)
     extends Error {
 
-    override def getMessage: String =
-      s"${item.description} cannot be converted to ${targetType.runtimeClass.getSimpleName}: " +
+    override def getMessage: String = {
+      // Note: https://issues.scala-lang.org/browse/SI-2034
+      val className = targetType.runtimeClass.getName
+      val simpleName = className.substring(className.lastIndexOf(".")+1)
```
wennergr: This logic is simpler then what exists inside getClass.getSimpleName. 

I can, if needed, make the output identical to that.

vkostyukov:
Thanks @wennergr! This looks good to me.

PR #726, opened 2017-01-25
fix Server-sent_events link

xuwei-k:


vkostyukov:
Thank you @xuwei-k! Finch is [a real Scala project now](https://twitter.com/rossabaker/status/799727645634297856)!

PR #729, opened 2017-01-27
Fix a broken link in the doc

n4to4:
Both auth.md and json.md were removed, so cookbook.md should be next, I guess.

https://github.com/finagle/finch/commit/bcb84a58dbebf38ce9435a321a04da8a05997d3b
https://github.com/finagle/finch/commit/dea22d72509c9a739b42aa2e900e95c89916eacb


vkostyukov:
Thank you @n4to4! Nice catch!

PR #731, opened 2017-01-27
Add tests to increase Output coverage.

DieBauer:
This merge request aims to increase the coverage of the `Output` class.

### DIFF
--- a/core/src/test/scala/io/finch/OutputSpec.scala
+++ b/core/src/test/scala/io/finch/OutputSpec.scala
@@ -1,8 +1,11 @@
 package io.finch
 
+import java.nio.charset.{Charset, StandardCharsets}
+import scala.util.{Failure, Try}
+
 import com.twitter.finagle.http.Status
 import com.twitter.io.Buf
-import java.nio.charset.{Charset, StandardCharsets}
+import com.twitter.util.{Await, Future}
 
 class OutputSpec extends FinchSpec {
 
@@ -10,6 +13,12 @@ class OutputSpec extends FinchSpec {
 
   it should "propagate status to response" in {
     check { o: Output[String] => o.toResponse[Text.Plain].status == o.status }
+  }
+
+  it should "propagate overridden status to response" in {
+    check { (o: Output[String], s: Status) =>
+      o.withStatus(s).toResponse[Text.Plain].status === s
+    }
   }
 
   it should "propagate charset to response" in {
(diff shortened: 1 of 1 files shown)


core/src/test/scala/io/finch/OutputSpec.scala
```diff
@@ -12,6 +15,13 @@ class OutputSpec extends FinchSpec {
     check { o: Output[String] => o.toResponse[Text.Plain].status == o.status }
   }
 
+  it should "propagate overridden status to response" in {
+    check { (o: Output[String], s: Status) =>
+      val rep = o.withStatus(s).toResponse[Text.Plain]
+      rep.status === s
```
vkostyukov: Let's inline these into `o.withStatus(s).toResponse[Text.Plain].status === s`

core/src/test/scala/io/finch/OutputSpec.scala
```diff
@@ -83,4 +93,56 @@ class OutputSpec extends FinchSpec {
         Encode[String, Text.Plain].apply(op.value, op.charset.getOrElse(StandardCharsets.UTF_8))
     }
   }
+
+  it should "create an empty endpoint with given status when calling unit" in {
+    check { s: Status =>
+      val rep = Output.unit(s).toResponse[Text.Plain]
```
vkostyukov: Please, inline `Output.unit(s).toResponse[Text.Plain].status === s`.

core/src/test/scala/io/finch/OutputSpec.scala
```diff
@@ -83,4 +93,56 @@ class OutputSpec extends FinchSpec {
...
+      rep.status === s
+    }
+  }
+
+  it should "throw an exception on calling value on an Empty output" in {
+    import org.scalatest.prop.PropertyChecks
+    PropertyChecks.forAll { e: Output.Empty =>
+      the[IllegalStateException] thrownBy e.value should have message "empty output"
```
vkostyukov: Why using `forAll` directly here? Is that b/c of `the[Exception]` syntax? If we could do just wrap `e.value` with `Try` and test the equality instead.

core/src/test/scala/io/finch/OutputSpec.scala
```diff
@@ -83,4 +93,56 @@ class OutputSpec extends FinchSpec {
...
+    }
+  }
+
+  it should "throw an exception on calling value on a Failure output" in {
+    import org.scalatest.prop.PropertyChecks
+    PropertyChecks.forAll { f: Output.Failure =>
```
vkostyukov: Same question about `forAll`.

core/src/test/scala/io/finch/OutputSpec.scala
```diff
@@ -83,4 +93,56 @@ class OutputSpec extends FinchSpec {
...
+      the[Exception] thrownBy f.value should have message f.cause.getMessage
+    }
+  }
+
+  it should "check equivalences of arbitrary string outputs" in {
+    check {
+      forAll(genOutput[String], genOutput[String]) {
```
vkostyukov: Why we need both `check` and `forAll` here?

core/src/test/scala/io/finch/OutputSpec.scala
```diff
@@ -83,4 +93,56 @@ class OutputSpec extends FinchSpec {
...
+  }
+
+  it should "check equivalences of failure outputs" in {
+    check { (fa: Output.Failure, fb: Output.Failure) =>
+      Output.outputEq[String].eqv(fa, fb) === (fa == fb)
+    }
+  }
+
+  it should "traverse arbitrary outputs" in {
+    check {
```
vkostyukov: Same question about `check` and `forAll`.

core/src/test/scala/io/finch/OutputSpec.scala
DieBauer: Yes, will do.

core/src/test/scala/io/finch/OutputSpec.scala
DieBauer: Good question, seems that check suffices then, will drop the scalatest syntactic sugar! thanks :)

DieBauer:
@vkostyukov thanks for the review, I've applied changes as requested!

vkostyukov:
Thank you again @DieBauer! This looks great and I especially like the +2% coverage increase!

Issue #659, opened 2016-10-22
Use tut for to documentation


ilya-murzinov:
We should use [tut](https://github.com/tpolecat/tut) to make sure that documentation is up-to-date and compiles.

erikwj:
I'm new to both Finch and tut so wanted to work on this issue to learn both. 

As a starter I converted the `scala` cli examples in docs/cookbook.md to `tut` examples. But lots of examples start a server on the same port. For example 
```scala
import com.twitter.finagle.Http
import com.twitter.finagle.http.Response
import con.twitter.io.Buf
import io.circe.generic.auto._
import io.finch._
import io.finch.circe._

case class Message(message: String)

val json: Endpoint[Message] = get("json") {
  Ok(Message("Hello, World!"))
}

val text: Endpoint[Response] = get("text") {
  val rep = Response()
  rep.content = Buf.Utf8("Hello, World!")
  rep.contentType = "text/plain"

  rep
}

Http.server.serve(":8081", (json :+: text).toServiceAs[Application.Json])
```

is converted into 

```tut:silent
import com.twitter.finagle.Http
import com.twitter.finagle.http.Response
import com.twitter.io.Buf
import io.circe.generic.auto._
import io.finch._
import io.finch.circe._
import com.twitter.util.Await

case class Message(message: String)

val json: Endpoint[Message] = get("json") {
  Ok(Message("Hello, World!"))
}

val text: Endpoint[Response] = get("text") {
  val rep = Response()
  rep.content = Buf.Utf8("Hello, World!")
  rep.contentType = "text/plain"

  rep
}

val server = Http.server.serve(":8081", (json :+: text).toServiceAs[Application.Json])

//Running ...

//And stopping so `tut` can do it's job
server.close()
```

It's a working solution, but are there better solutions? Let me know

vkostyukov:
Thanks @erikwj!

I think we can easily omit the `Http.server.serve` part, here and in the rest of the example. Having `toServerAs` call at the end of each (most?) example seems quite reasonable to me. What do you think?

erikwj:
Yes seems like a good idea. Have 1 full example should be enough

muuki88:
WDYT of using [sbt-microsites](https://github.com/47deg/sbt-microsites) to build the documentation? I could submit a pull request if you are interested.

erikwj:
Have been busy, but should have time to work on this coming weeks. I thinks it's a great idea, but I will first compile all docs with tut

[...]

PR #733, opened 2017-01-29
Fix doc

n4to4:
- Fix some markdown errors and example code errors
- Update circe links
- Add a link to httpie site

It would be nice if we use `tut` for documents under `docs` to prevent compile errors, probably?


vkostyukov:
Thanks @n4to4! Using tut has been on our radar for quite a while: #659.

PR #735, opened 2017-01-31
Fix example code

n4to4:
Some example programs are not runnable because of the wrong `main` signature.


n4to4:
hmm. why codecov complains...?

vkostyukov:
LGTM! Thanks again, @n4to4!

PR #736, opened 2017-01-31
Update library versions

n4to4:


vkostyukov:
Thanks @n4to4! This looks good to me.

PR #742, opened 2017-02-11
Bump dependencies

vkostyukov:


od0:
Thanks! Great work as always with Finch :)

PR #749, opened 2017-02-22
Add .codecov.yml and set threshold

n4to4:
Same as https://github.com/circe/circe/issues/572

See also [.codecov.yml spec](https://github.com/codecov/support/wiki/Codecov-Yaml#full-yaml)

I have experienced and confused by this problem (https://github.com/finagle/finch/pull/735#issuecomment-276304723), so this could be useful?

### DIFF
--- a/.codecov.yml
+++ b/.codecov.yml
@@ -1,0 +1,5 @@
+coverage:
+  status:
+    project:
+      default:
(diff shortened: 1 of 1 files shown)


vkostyukov:
Thanks @n4to4! Looks good to me!

PR #752, opened 2017-03-02
Fix Date Header to conform to RFCs 7231 and 5322

rpless:
Today we noticed one of our HTTP clients logging a warning about our Finch services responding with an invalid Date header. The Date in question looked like `Thu, 2 Mar 2017 01:47:19 GMT`. As it turns out this is invalid according to [RFC 7231](https://tools.ietf.org/html/rfc7231#section-7.1.1.1) (and also RFC 2616, but this was made obsolete by 7231). The problem is that the day of the month must be 2 digits. So the date should actually be `Thu, 02 Mar 2017 01:47:19 GMT`.

This change uses an explicit `DateTimeFormatter` pattern and uses a time zone of GMT to adhere to the spec. I could not find a way to make `DateTimeFormatter.RFC_1123_DATE_TIME` to pad the day of the month field, but if anyone knows of way I think that would better. Let me know what you think.

For reference:
The preferred date time is defined [here](https://tools.ietf.org/html/rfc7231#section-7.1.1.1).
It also references this section of [RFC 5322](https://tools.ietf.org/html/rfc5322#section-3.3)

### DIFF
--- a/core/src/main/scala/io/finch/internal/currentTime.scala
+++ b/core/src/main/scala/io/finch/internal/currentTime.scala
@@ -1,12 +1,15 @@
 package io.finch.internal
 
+import java.time.{Instant, ZoneId}
 import java.time.format.DateTimeFormatter
-import java.time.Instant
-import java.time.ZoneOffset
+import java.util.Locale
+
 
 object currentTime {
   private[this] val formatter: DateTimeFormatter =
-    DateTimeFormatter.RFC_1123_DATE_TIME.withZone(ZoneOffset.UTC)
+    DateTimeFormatter.ofPattern("EEE, dd MMM yyyy HH:mm:ss zzz")
+      .withLocale(Locale.ENGLISH)
+      .withZone(ZoneId.of("GMT"))
 
   @volatile private[this] var last: (Long, String) = (0, "")
 


DieBauer:
Nice catch, I've a question regarding the specific "GMT" timezone. I see in the RFC that it's a former version of UTC. So why don't we use "UTC" here?

From the referenced RFC7231 section 7.1.1.1: 

> An HTTP-date value represents time as an instance of Coordinated
>    Universal Time (UTC).  The first two formats indicate UTC by the
>    three-letter abbreviation for Greenwich Mean Time, "GMT", a
>    predecessor of the UTC name

from the javadocs https://docs.oracle.com/javase/8/docs/api/java/time/ZoneId.html#of-java.lang.String- : 

> If the zone ID equals 'GMT', 'UTC' or 'UT' then the result is a ZoneId with the same ID and rules equivalent to ZoneOffset.UTC.

rpless:
Hmm, yeah I only went with GMT because that was what all the examples in the spec said. Sticking with UTC is probably better.

core/src/main/scala/io/finch/internal/currentTime.scala
```diff
@@ -1,12 +1,12 @@
...
+import java.time.{Instant, ZoneId}
 import java.time.format.DateTimeFormatter
-import java.time.Instant
-import java.time.ZoneOffset
+
 
 object currentTime {
   private[this] val formatter: DateTimeFormatter =
-    DateTimeFormatter.RFC_1123_DATE_TIME.withZone(ZoneOffset.UTC)
+    DateTimeFormatter.ofPattern("EEE, dd MMM yyyy HH:mm:ss zzz").withZone(ZoneId.of("GMT"))
```
vkostyukov: Should we also set a `Locale` here? (Looks like [http4s does it](https://github.com/http4s/blaze/blob/7701fe61d826cf2bc842e34e9f6d17c9694f1b74/http/src/main/scala/org/http4s/blaze/http/util/HeaderTools.scala#L13)).

rpless: I think we may want to set the `Locale`, however I don't think we want to use `Locale.US`. Maybe `Locale.getDefault()`?

rpless: Also based on what http4s does, we may actually want to keep the `TimeZone` as GMT instead of UTC.

vkostyukov: +1 for using GMT for now.

I'm not enterlly sure about the locale though. Do you think we can look up a couple more places (like Play) to see what they are doing instead? Looks like [Finatra is using Locale.ENGLISH](https://github.com/twitter/finatra/blob/d46935d0da39e4fed0a528db078edf09e02e44d9/http/src/main/scala/com/twitter/finatra/http/filters/HttpResponseFilter.scala#L25). I think I'd be perfectly happy with `Locale.US`.

rpless: I think [Play](https://github.com/playframework/playframework/blob/38abd1ca6d17237950c82b1483057c5c39929cb4/framework/src/play/src/main/scala/play/api/http/package.scala#L20) does the same thing as Finatra. [Akka Http](https://github.com/akka/akka-http/blob/5932237a86a432d623fafb1e84eeeff56d7485fe/akka-http-core/src/main/java/akka/http/javadsl/model/DateTime.java#L90) uses its own mechanism, but the comment there seems to indicate its always "GMT" and nothing needs a Locale. I'm kinda tending toward `Locale.English`

vkostyukov: Works for me. Let's go with`Local.ENGLISH` then.

rpless: Although honestly I'm not even sure what part of the pattern would take the `Locale` into consideration. Maybe the `GMT` part, in which case I think both `Locale.ENGLISH` and `Locale.US` would do the same thing.

vkostyukov:
Thanks a lot @rpless! Outstanding work as usual!

vkostyukov:
BTW, how annoying it is? Want me to do a patch release with that fix?

rpless:
Thanks @vkostyukov. The patch release is probably not necessary if my use case is the only one it helps. We're actually also going to stop seeing the warning next week because today we switched over most of our Http Clients from Akka Http (which generates the warning) to Finagle which does not and for the rest of the services we found a way shut that validation off in Akka.

PR #760, opened 2017-03-21
Fix document

n4to4:


docs/src/main/tut/user-guide.md
```diff
@@ -441,7 +441,7 @@ For all `String`-based endpoints, Finch provides an `as[A]` method to perform ty
 is available for any `Endpoint[String]`, `Endpoint[Option[String]]` or `Endpoint[Seq[String]]` as
 long as a matching implicit `io.finch.DecodeEntity[A]` type-class is in the scope.
 
-This facility is designed to be intuitive, meaning that you do not have to provide a
+This facility is designed to be intuitive, meaning that you do not have to provide an
```
rpless: I'm not sure about this one. It is correct because the next word does begin with a vowel, but in my mind the word that you pronounce next is `DecodeEntity` (which would make `a` the correct form). I'm kinda wondering with we actually want to drop the `io.finch` part, I'm not sure it really adds anything. Its not really a big deal and I'm fine with this change as is. Would be curious if anyone else had thoughts.

docs/src/main/tut/user-guide.md
```diff
@@ -508,7 +508,7 @@ There are two API entry point into decoding JSON payloads: `jsonBody[A]` and `js
 These require a `Decode.Json[A]` instance to be available in the scope whenever they called.
 
 Finch comes with support for a number of [JSON libraries](#json). All these integration modules do
-is make the library-specific JSON decoders available for use as a `io.finch.Decode.Json[A]`. To take
+is make the library-specific JSON decoders available for use as an `io.finch.Decode.Json[A]`. To take
```
rpless: Same as above.

docs/src/main/tut/user-guide.md
BenFradet: Personally i'd be in favour of dropping the `io.finch` in these kinds of explanations since it breaks the "reading flow", at least for me.

n4to4: @rpless Thanks for your review.
I thought there's one occurrence [here](https://finagle.github.io/finch/cookbook.html#fixing-the-toservice-compile-error) in the doc. But right now I noticed that it's an `Encoder` that follows the `an`.
I'm fine dropping `io.finch` or something else we agree.

vkostyukov: +1 for dropping the `io.finch` prefix here.

n4to4: So should I remove `io.finch` in the entire doc? in this PR?
Though there seem to be some exceptions like [`io.finch.circe.dropNullKeys._` here](https://finagle.github.io/finch/best-practices.html#picking-a-json-library), and maybe some other places that could be ambiguous?

rpless: If you would like to go through and change places that are using full paths to describe specific paths, I think that would be good. We probably want to be consistent in the docs. In that case you linked, it is specifically talking about what package to import so we can't change that. If you just want to do the ones in this PR that's fine too, I'm happy to go through and change the rest at some point .

n4to4: Okay, I reverted these changes, and left `io.finch` as is for consistency. I'd like to let others to do the dropping `io.finch` parts (since I'm not so sure which one to drop and which one to keep).

rpless:
Thanks again @n4to4. Would you mind squashing this down to a single commit?

n4to4:
@rpless Done

PR #762, opened 2017-03-29
New service templates

tomjadams:
Added new Redbubble & other service templates.


vkostyukov:
Thanks @tomjadams! Looks good to me!

PR #766, opened 2017-04-22
Make Input actually immutable

vkostyukov:
While `Input` was always a case class it wasn't fully immutable given all the `withX` methods mutated the underlying (also mutable) Finagle's request. This PR makes it actually immutable by copying the request each time it's mutated.

In addition to that, I also decided to rebrand `Input.path` to `Input.route` to clearly distinguish it from `Request.path` and state that it's used with the matching/routing endpoints.

### DIFF
--- a/core/src/main/scala/io/finch/Input.scala
+++ b/core/src/main/scala/io/finch/Input.scala
@@ -10,16 +10,21 @@ import org.jboss.netty.handler.codec.http.multipart.{DefaultHttpDataFactory, Htt
 import shapeless.Witness
 
 /**
- * An input for [[Endpoint]].
+ * An input for [[Endpoint]] that glues two individual pieces together:
+ *
+ * - Finagle's [[Request]] needed for evaluating (e.g., `body`, `param`)
+ * - Finch's route (represented as `Seq[String]`) needed for matching (e.g., `path`)
  */
-final case class Input(request: Request, path: Seq[String]) {
-  def headOption: Option[String] = path.headOption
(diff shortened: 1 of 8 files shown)


core/src/main/scala/io/finch/Input.scala
```diff
@@ -10,16 +10,21 @@ import org.jboss.netty.handler.codec.http.multipart.{DefaultHttpDataFactory, Htt
...
 
 /**
- * An input for [[Endpoint]].
+ * An input for [[Endpoint]] that glues two individual pieces together:
+ *
+ * - Finagle's [[Request]] needed for evaluating (e.g., `body`, `param`)
+ * - Finch's route (represented as `Seq[String]`) needed for matching (e.g., `path`)
  */
-final case class Input(request: Request, path: Seq[String]) {
-  def headOption: Option[String] = path.headOption
```
rpless: Is the idea here that we can just call these directly on `route` and duplicating methods from `Seq` was unnecessary?

vkostyukov: Right. I think those methods were rather confusing than helpful. I like how explicit `input.route.headOption` is.

vkostyukov:
Thanks @rpless! I was on a fence whether or not I want to do a deprecation cycle for this but decided to now worry too much given how unpopular those methods should be around Finch users. We probably only used them internally.

rpless:
Yeah that's good point, I'd be a little surprised if more than a few people use them.

Issue #765, opened 2017-04-21
Failed outputs are rendered as responses with no body


aroberts:
I'm trying to upgrade to 0.14 (from 0.11), and I'm encountering an issue with custom error responses. I am using finch-circe, and I have a custom `Encoder[Exception]` which is imported into the implicit scope where I'm calling `toServiceAs`, but I don't see my custom output. Also, logging that I injected into my encoder never fires, which further indicates that my code is not getting run.

I see responses that look like:
```
HTTP/1.1 400 Bad Request
Content-Length: 0
Date: Fri, 21 Apr 2017 18:05:39 GMT
Vary: Origin
```

@vkostyukov notes that the date header means Finch is processing the response, because finagle by itself doesn't have that. However, even if there was no message, the default `Encoder[Exception]` instance would seem to return `"{message:null}"` in the body, in the event of a missing message. It's not clear why the bodies are missing entirely.

aroberts:
It appears that omitting 
```scala
    .configured(Http.Netty4Impl)
```

was the cause of this issue in my particular case. I'm going to close this issue, as I think it's too tied up in my personal dependency situation to have a clear path forward.

rlazoti:
I was having the same issue you reported (Bad Request responses with no body content).
If I understood correctly, it's caused because there're 2 encoders for Exception type. 

So there're two ways to "fix" it temporally:

1 - Removing your encoder (Encoder[Exception])
2 - Removing the default Encoder.
I ended up doing this with the following import:

```scala
import io.finch.circe.{
  jacksonSerializer,
  decodeCirce,
  encodeCirce
}
```
Instead of 

```scala
import io.finch.circe._
```

aroberts:
I created a minimal reproduction of the issue: https://github.com/aroberts/finch-14-errors

@rlazoti I hear what you're saying, but why would the compiler not resolve the implicit resolution according to precedence rules? Also, version 0.11 has an `Encoder[Exception]` instance, but does not exhibit this issue. I appreciate the workaround but I still think there's some underlying weirdness here that should be fixed.

rpless:
My initial thought was that we should have been seeing an ambiguous implicit values error from the compiler. If you turn on the `-Xlog-implicits` compiler option you will see something like this:
```encodeCirce is not a valid implicit value for io.finch.Encode.Aux[Exception,io.finch.Application.Json] because:
[info] hasMatchingSymbol reported error: ambiguous implicit values:
[...]

PR #775, opened 2017-05-15
Remove encoding for exceptions from all of the json projects.

rpless:
Removed the instances for encoding Exceptions as json as per https://github.com/finagle/finch/issues/765.


vkostyukov:
Thanks @rpless! This looks good to me!

What do you think as a follow-up (or even part of this PR) we stop testing `JsonLaws` for `Exception` (see [here](https://github.com/finagle/finch/blob/master/json-test/src/main/scala/io/finch/test/AbstractJsonSpec.scala#L31-L42))? Maybe this will be the right thing to do give we no longer provide any guarantees about encoding exceptions.

rpless:
I think that's probably the right thing to do. At this point its basically just testing that the encodes for exception that we define in tests work, its not really claiming anything about finch. I'll update this PR tonight.

vkostyukov:
Thanks again!

PR #788, opened 2017-06-10
Update catbird and other versions

travisbrown:
I just published a new catbird for Util 6.45.0, but it looks like Finch was behind a release anyway.


vkostyukov:
Thanks a lot, @travisbrown!

Issue #777, opened 2017-05-17
Move finch-oauth2 into its own project


vkostyukov:
This would simplify releases. See https://github.com/finch/finch-oauth2

PR #804, opened 2017-06-15
Move oauth2 project

jguitana:
Moving oauth2 to [finch-oauth2](https://github.com/finch/finch-oauth2) according to [#777](https://github.com/finagle/finch/issues/777).

This is the PR for the [finch-oauth2 project](https://github.com/finch/finch-oauth2/pull/1).

This should follow [moving examples](https://github.com/finagle/finch/pull/803) as that has a library dependecy on this.


vkostyukov:
Hey @jguitana! Do you want to rebase it (sorry it took us a while to review this)?

I think we should keep most of the samples in place for now but only remove OAuth2.

vkostyukov:
This is suppressed by #824.

PR #807, opened 2017-06-16
Optimize Endpoint and Output

vkostyukov:
Applying some technics I learned from @fwbrasil.

The biggest thing is to avoid closure allocations by embedding a `Funciton1` or `Function2` into an enclosing scope/object. This played really nicely in the `Endpoint` (see performance results bellow).

In addition to that optimization I also restructured the `Output` type a bit:

- Got rid of `Meta` and merged its members into each object
- Added `traverseFlatten` (a specialised and cheap version of `traverse().flatten`)
- Added `withHeaders` and `withCookies` that allow copying all things in one shot (no need to allocate an intermediate `Output` for each header).

In a nutshell
- `map*` operations are now up to **40% faster** (400 bytes fewer allocations on each call)
- `::` is now up to **5% faster** (120 bytes fewer allocations on each call)

Running time:

```
NEW:

MapBenchmark.map                             avgt   10   633.837 ±  14.613   ns/op
MapBenchmark.mapAsync                        avgt   10   624.249 ±  22.426   ns/op
MapBenchmark.mapOutput                       avgt   10   734.677 ±  30.215   ns/op
MapBenchmark.mapOutputAsync                  avgt   10   737.170 ±  29.808   ns/op
ProductBenchmark.bothMatched                 avgt   10  1175.716 ±  44.236   ns/op
ProductBenchmark.leftMatched                 avgt   10    26.510 ±   2.335   ns/op
ProductBenchmark.rightMatched                avgt   10     5.081 ±   0.112   ns/op

OLD:

MapBenchmark.map                             avgt   10   624.174 ±  28.621   ns/op
MapBenchmark.mapAsync                        avgt   10   647.369 ±  30.775   ns/op
MapBenchmark.mapOutput                       avgt   10  1221.053 ±  39.319   ns/op
MapBenchmark.mapOutputAsync                  avgt   10  1202.541 ±  44.432   ns/op
ProductBenchmark.bothMatched                 avgt   10  1224.278 ±  49.114   ns/op
ProductBenchmark.leftMatched                 avgt   10    29.856 ±   0.709   ns/op
ProductBenchmark.rightMatched                avgt   10     5.209 ±   0.112   ns/op
```

Allocations:

```
NEW: 

MapBenchmark.map:·gc.alloc.rate.norm                 avgt   10   664.000 ±   0.001    B/op
MapBenchmark.mapAsync:·gc.alloc.rate.norm            avgt   10   664.000 ±   0.001    B/op
MapBenchmark.mapOutput:·gc.alloc.rate.norm           avgt   10   832.000 ±   0.001    B/op
MapBenchmark.mapOutputAsync:·gc.alloc.rate.norm      avgt   10   832.000 ±   0.001    B/op
ProductBenchmark.bothMatched:·gc.alloc.rate.norm     avgt   10  1400.001 ±   0.001    B/op
ProductBenchmark.leftMatched:·gc.alloc.rate.norm     avgt   10    80.000 ±   0.001    B/op
ProductBenchmark.rightMatched:·gc.alloc.rate.norm    avgt   10    ≈ 10⁻⁶              B/op

OLD:

MapBenchmark.map:·gc.alloc.rate.norm                 avgt   10   696.000 ±   0.001    B/op
MapBenchmark.mapAsync:·gc.alloc.rate.norm            avgt   10   696.000 ±   0.001    B/op
MapBenchmark.mapOutput:·gc.alloc.rate.norm           avgt   10  1232.001 ±   0.001    B/op
MapBenchmark.mapOutputAsync:·gc.alloc.rate.norm      avgt   10  1232.001 ±   0.001    B/op
ProductBenchmark.bothMatched:·gc.alloc.rate.norm     avgt   10  1520.001 ±   0.001    B/op
ProductBenchmark.leftMatched:·gc.alloc.rate.norm     avgt   10   104.000 ±   0.001    B/op
ProductBenchmark.rightMatched:·gc.alloc.rate.norm    avgt   10    ≈ 10⁻⁶              B/op
```

/cc @travisbrown @rpless

### DIFF
--- a/core/src/main/scala/io/finch/Endpoint.scala
+++ b/core/src/main/scala/io/finch/Endpoint.scala
@@ -78,15 +78,19 @@ trait Endpoint[A] { self =>
   /**
    * Maps this endpoint to the given function `A => Future[B]`.
    */
-  final def mapAsync[B](fn: A => Future[B]): Endpoint[B] = new Endpoint[B] {
-    final def apply(input: Input): Endpoint.Result[B] = self(input) match {
-      case EndpointResult.Matched(rem, out) =>
-        EndpointResult.Matched(rem, out.flatMapF(oa => oa.traverse(a => fn(a))))
-      case _ => EndpointResult.Skipped
-    }
+  final def mapAsync[B](fn: A => Future[B]): Endpoint[B] =
+    new Endpoint[B] with (Output[A] => Future[Output[B]]) {
 
-    override def item = self.item
-    final override def toString: String = self.toString
+      final def apply(oa: Output[A]): Future[Output[B]] = oa.traverse(fn)
+
+      final def apply(input: Input): Endpoint.Result[B] = self(input) match {
+        case EndpointResult.Matched(rem, out) =>
+          EndpointResult.Matched(rem, out.flatMapF(this))
+        case _ => EndpointResult.Skipped
+      }
+
+      final override def item = self.item
+      final override def toString: String = self.toString
   }
 
   /**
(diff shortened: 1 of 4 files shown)


core/src/main/scala/io/finch/Endpoint.scala
```diff
@@ -78,15 +78,19 @@ trait Endpoint[A] { self =>
...
    * Maps this endpoint to the given function `A => Future[B]`.
    */
-  final def mapAsync[B](fn: A => Future[B]): Endpoint[B] = new Endpoint[B] {
-    final def apply(input: Input): Endpoint.Result[B] = self(input) match {
-      case EndpointResult.Matched(rem, out) =>
-        EndpointResult.Matched(rem, out.flatMapF(oa => oa.traverse(a => fn(a))))
-      case _ => EndpointResult.Skipped
-    }
+  final def mapAsync[B](fn: A => Future[B]): Endpoint[B] =
+    new Endpoint[B] with (Output[A] => Future[Output[B]]) {
```
rpless: Out of curiosity, what does `with (Output[A] => Future[Output[B]])` (and the similar mixins in the other map* methods) get us over the existing implementation?

core/src/main/scala/io/finch/Output.scala
```diff
@@ -10,124 +10,119 @@ import java.nio.charset.{Charset, StandardCharsets}
...
-      val ob = fn(v)
-      ob.withMeta(ob.meta.copy(
-        headers = ob.headers ++ p.headers,
-        cookies = ob.cookies ++ p.cookies
-      ))
-
-    case f @ Output.Failure(_, _) => f
-    case e @ Output.Empty(_) => e
+  final def flatMap[B](fn: A => Output[B]): Output[B] = this match {
+    case p: Output.Payload[A] => fn(p.value).withCookies(p.cookies).withHeaders(p.headers)
```
rpless: I like this. It looks a lot cleaner.

core/src/main/scala/io/finch/Endpoint.scala
vkostyukov: So we can pass the same instance (`this` in that case) as a parameter to the flatMap/map functions. We don't have to allocate a closure each time. Apparently, mixing a new interface (`Function1` in that case) and then defining a new method is super cheap.

vkostyukov:
Thanks, @rpless! Yeah, these numbers are average time and allocations: less is better.

core/src/main/scala/io/finch/Endpoint.scala
rpless: Oh that's really interesting. Makes sense now that I think about it, just not immediately intuitive. Thanks for explaining!

vkostyukov: Yeah, it's pretty mind-blowing that you can do that. I think Flavio calls this technic "merge interfaces". We did a whole bunch of optimizations for [Twitter Futures](https://github.com/twitter/util/commits/develop/util-core/src/main/scala/com/twitter/util/Future.scala) based on that (and other ideas). Results were outstanding.

Issue #513, opened 2016-01-29
Replace ValidationRule with Cat's Validated


vkostyukov:
I don't have a particular solution in mind, but I think it's worth to have an opened ticket for this before we forget. I eagerly want to replace Finch's data types with the ones community share and understand. `ValidationRule` is a good candidate to be reaplced (in some way) with Cat's [`Validated`](http://non.github.io/cats//tut/validated.html).

ilya-murzinov:
@vkostyukov can you please clarify how exactly do you see this possible?
I don't get it, because finch's ```ValidationRule``` is just a some kind of convenient DSL for combining boolean predicates, but cats' ```Validated``` is a data type which encodes validation result.

vkostyukov:
I think what we should aim for is to have default endpoints (i.e., `param`, `header`) returning `Endpoint[ValidatedNel[Error, A]]` instead of `Endpoint[A]`. A very similar (yet simpler) approach is described in [this gist](https://gist.github.com/vkostyukov/6192a5280e485661d3d5f88dd5899f54).

ImLiar:
What about the composition of endpoints?
Let's say I have:
```scala
get(param("foo") :: header("bar")) { (foo: A?, bar: B?) => {
    ...
  }
}
```

What is a type of `A?` and `B?` here?

sergeykolbasov:
@prayagupd Yes, status should be a part of an error.

In our project services always return `F[Either[Error, A]]` where `Error <: Throwable` is a super type that guarantees `status`, `description` etc. And then we just fold it into `Output` using `Output.payload` or `Output.failure`

joroKr21:
Validation removed in #1471

PR #808, opened 2017-06-17
Experiment with cats Validated

rpless:
I was dealing with some heavy use of cats Validated recently and was reminded of #513. This PR adds a `validate` method to `Endpoint` that uses Cats' `Validated` to validate the data of an endpoint. I did not remove `ValidationRule`, `should`, and `shouldNot` because I think we should let users decide which they like better. We should let users experiment and provide feedback on it.

In addition, I add an `Errors.of` method that mimics `NonEmptyList.of` to make constructing `Errors` involve slightly less typing. 

I also implemented `should` and `shouldNot` in terms of `validate`. After @vkostyukov's last PR I'm suddenly very aware of allocations and am willing to undo this if we think the extra allocation of a `Validated` isn't ok.

### DIFF
--- a/core/src/main/scala/io/finch/Endpoint.scala
+++ b/core/src/main/scala/io/finch/Endpoint.scala
@@ -4,7 +4,7 @@ import java.nio.charset.Charset
 import scala.reflect.ClassTag
 
 import cats.Alternative
-import cats.data.NonEmptyList
+import cats.data.{NonEmptyList, Validated}
 import com.twitter.finagle.Service
 import com.twitter.finagle.http.{Cookie, Request, Response}
 import com.twitter.util.{Future, Return, Throw, Try}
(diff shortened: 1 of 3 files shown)


core/src/main/scala/io/finch/Endpoint.scala
```diff
@@ -330,6 +332,17 @@ trait Endpoint[A] { self =>
    */
   final def shouldNot(rule: ValidationRule[A]): Endpoint[A] = shouldNot(rule.description)(rule.apply)
 
+  /** Validates the result of this enpoint using a function that returns a [[cats.data.Validated]].
```
hderms: typo in `enpoint` => `endpoint`

core/src/main/scala/io/finch/Endpoint.scala
```diff
@@ -289,10 +289,12 @@ trait Endpoint[A] { self =>
...
    *         Otherwise the future fails with an [[Error.NotValid]] error.
    */
-  final def should(rule: String)(predicate: A => Boolean): Endpoint[A] = mapAsync(a =>
-    if (predicate(a)) Future.value(a)
-    else Future.exception(Error.NotValid(self.item, "should " + rule))
-  )
+  final def should(rule: String)(predicate: A => Boolean): Endpoint[A] = {
+    validate { (a: A, item) =>
+      if (predicate(a)) Validated.Valid(a)
+      else Validated.Invalid(Errors.of(Error.NotValid(item, "should " + rule)))
```
joroKr21, line 295: Besides the allocation that's also a change of semantics that I don't think we should do. Other than that I'd be happy to add the `validate` method. But I think to make it useful we also need to add a `Semigroup[Errors]` instance. It's unfortunate that it uses `NonEmptyList` though because lists don't support efficient concatenation.

rpless: Ah right it would require a Semigroup for errors, but thats fairly straightforward to implement. This PR is fairly old (it was made prior to cats effect being introduced). I think it might make sense to close this and start fresh. In terms of performance ValidatedNec was not available when I wrote this, perhaps moving to that would address the performance concerns.

PR #809, opened 2017-06-18
Update wrk benchmark and numbers

vkostyukov:
Upgrading the benchmark so it's now pretty much simulates TE (no inbound parsing). Also, re-run everything on my computer and updated the numbers. It's crazy, but due to recent performance optimizations in Circe and Finch, we do much better now: **95% of Finagle throughput**.

/cc @spockz, @travisbrown, @rpless


travisbrown:
Wow! And that's even without all the most recent stuff in circe.

rpless:
Those numbers for Finch look great. I'm a little surprised about the Finagle ones dropping so much, though. Maybe related to the performance issue with Finagle Http/Netty 4 that you mentioned in the 0.15.0 release notes?

vkostyukov:
@rpless Oh yeah. That's the overhead from Netty4 HTTP. It should go away in a month or two.

PR #811, opened 2017-07-12
Multiple file uploads

jguitana:
Adds an endpoint fileUploads(paramName). As mentioned in gitter, this is useful for us.

A suggestion would be to maintain only APIs for multiple file uploads, essentially replacing fileUpload and fileUploadOption (one FileUpload) with fileUploads and fileUploadsOption (Seq[FileUpload]) as the latter also represent the former. If you want I can do this.

### DIFF
--- a/core/src/main/scala/io/finch/endpoint/multipart.scala
+++ b/core/src/main/scala/io/finch/endpoint/multipart.scala
@@ -1,5 +1,6 @@
 package io.finch.endpoint
 
+import cats.data.NonEmptyList
 import com.twitter.finagle.http.exp.{Multipart => FinagleMultipart}
 import io.catbird.util.Rerunnable
 import io.finch._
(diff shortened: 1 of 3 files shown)


vkostyukov:
Thanks, @jguitana! What do you think about the following structure (basically mirroring what we have for regular params):

- `fileUpload: Endpoint[FileUpload]` required file upload (singular)
- `fileUploads: Endpint[Seq[FileUploads]]` multiple file uploads (could be empty)
- `fileUploadOption: Endpoint[Option[FileUpload]]` optional file uplad (singular)
- `fileUploadsNel: Endpoint[NonEmptyList[FileUploads]]` multiple file uploads (can not be empty)

jguitana:
Looks good! There is a consideration for `fileUploads: Endpoint[Seq[FileUploads]]`.

- `fileUpload: Endpoint[FileUpload]` must have param and upload
- `fileUploads: Endpoint[Seq[FileUploads]]` _no param / param and no uploads means empty seq?_
- `fileUploadOption: Endpoint[Option[FileUpload]]` optional param
- `fileUploadsNel: Endpoint[NonEmptyList[FileUploads]]` must have param and >1 uploads

vkostyukov:
Hey @jguitana! I appreciate your patience with this review and I'm sorry it's taking me so long to work through.

I want to make sure to include this in the next release, but it has a merge conflict now (as we changed the structure around multipart in #865). I'd be happy to take your branch, squash/rebase it and go from there, unless you want to do that yourself. Let me know!

jguitana:
Hi @vkostyukov ! Just updated with latest changes.

Tried adding better tests but multiple `.add(FileElement(name, value, Some("image/gif"), Some("dealwithit.gif")))` do not seem to work for the same name.

So basically this is working for multipart files with the same `name`. I can try to add functionality for multiple names sometime, if you want. That may require more complex changes than what was done here.

core/src/main/scala/io/finch/endpoint/multipart.scala
```diff
@@ -38,12 +39,22 @@ private abstract class Multipart[A, B](name: String) extends Endpoint[B] {
 private object Multipart {
   trait Required[A] { _: Multipart[A, A] =>
     protected def missing(name: String): Rerunnable[Output[A]] = Rs.paramNotPresent(name)
-    protected def present(a: A): Rerunnable[Output[A]] = Rs.payload(a)
+    protected def present(a: NonEmptyList[A]): Rerunnable[Output[A]] = Rs.payload(a.head)
```
vkostyukov: This may fail at runtime with `Seq.empty.head`. I think we should somehow distinguish empty/non-empty cases earlier.

vkostyukov: Oh please ignore me - that should work just fine given `a` is non-empty!

Issue #667, opened 2016-11-08
Streaming JSON Requests


vkostyukov:
Right now (as per 0.11) it's possible to stream JSON objects over the HTTP response (chunked) via `AsyncStream[CaseClass]`. Although, there is near zero support for the inbound streaming (streaming requests). This issue is related to #336.

I assume, for each JSON library we support, we can provide meanings for converting `AsyncStream[Buf]` into `AsyncStream[CaseClass]`. This might require a new type class for decoding since the `Decode` abstraction we have goes from `Buf` to `Try[A]`. Technically, we could utilize that and build something close to Netty's replying decoder (return `Throw` when we need more data), but in that case, it would be impossible to distinct streaming and non-streaming decoders (in case if we need too).

Anyways, there are lots of open questions here, but it seems totally possible to me (with zero API changes).

NOTE: @travisbrown has kindly agreed to find some time and work on iteratees support for Circe. This would mean that for Circe there will be two streaming APIs via AsyncStream and via iteratees (see #557).

vkostyukov:
I think the path forward here is to push this kind of functionality out of the finch-core and use something like fs2 and iteratee instead. There is a PR that introduces iteratee support in Finch (see #557) - this is what we need to be focusing on.

I still want to keep the very basic `AsyncStream` support in finch-core for the sake of simplicity. Although I'm not 100% sure that `Endpoint[AsyncStream[User]]` (as a streaming JSON response) is the right pattern here.

ImLiar:
I've picked it up and made some good progress. My approach is following:
- Get `AsyncStream[Buf]` from request
- Build `Enumerator[Future, Buf]` from it
- Use circe-streaming and transform `Buf` into `Json` and then `Json` into some `A`
- Use `Iteratee.fold` to build `AsyncStream[A]`

I guess it'll simplify life of potential users by hiding Iteratee/Enumeratee stuff under the hood.

vkostyukov:
Thanks @ImLiar! This is really cool!

One thought: I think we shouldn't try sticking to `AsyncStream[_]` API so hard. In fact, I'd prefer us using Circe/Iteratee abstractions instead (I honestly don't think `AsyncStream[_]` is a great tool for this job). I know we already support serving `AsyncStream[Foo]` (as a streaming JSON), but I really think we should reconsider that (and probably cut the support). Instead, we should probably support serving `Iteratee` (or something).

[...]

PR #812, opened 2017-07-13
Request streaming support through circe-streaming

ImLiar:
Related to https://github.com/finagle/finch/issues/667

- Added a project `finch-iteratee` that depends on `io.iteratee`
- Added default decoder atop of `Enumeratee[Future, Buf, A]` to `finch-circe`

With this implementation user would need to add following imports:
```scala
import io.finch.iteratee._
```
to get an access to `enumeratorBody[A, CT]` and `enumeratorJsonBody[A]` endpoints (beside adding a new dependency `finch-iteratee`).


streaming/src/main/scala/io/finch/streaming/package.scala
```diff
@@ -0,0 +1,49 @@
...
+import com.twitter.util.Future
+import io.catbird.util._
+import io.finch.Endpoint.Result
+import io.finch.items.RequestItem
+import io.iteratee.{Enumeratee, Enumerator}
+
+/**
+  * Streaming module
+  */
+package object streaming {
```
vkostyukov: Should we extend `Module[Fturue]` (as Travis did here: https://github.com/finagle/finch/pull/557)?

streaming/src/main/scala/io/finch/streaming/package.scala
```diff
@@ -0,0 +1,49 @@
...
+      case None => Enumerator.empty[Future, Buf]
+      case Some(buf) => Enumerator.enumOne[Future, Buf](buf).append(enumeratorFromReader(reader))
+    })
+  }
+
+  /**
+    * An evaluating [[Endpoint]] that reads a required chunked streaming binary body, interpreted as
+    * an `Enumerator[Future, Buf]`. The returned [[Endpoint]] only matches chunked (streamed) requests.
+    */
+  def asyncBufBody: Endpoint[Enumerator[Future, Buf]] = new Endpoint[Enumerator[Future, Buf]] {
```
vkostyukov: I think it's fine not to expose this endpoint. Let's just expose one that deals with `A` (not `Buf`).

streaming/src/main/scala/io/finch/streaming/package.scala
```diff
@@ -0,0 +1,49 @@
...
+
+    final override def item: RequestItem = items.BodyItem
+    final override def toString: String = "asyncBufBody"
+  }
+
+  /**
+    * An evaluating [[Endpoint]] that reads a required chunked streaming binary body, interpreted as
+    * an `Enumerator[Future, A]`. The returned [[Endpoint]] only matches chunked (streamed) requests.
+    */
+  def asyncJsonBody[A](implicit ee: Enumeratee[Future, Buf, A]): Endpoint[Enumerator[Future, A]] = {
```
vkostyukov: This isn't really coupled to JSON as far as I can tell and can work with anything that's know how to go from `Buf` to `A`. How do you feel about renaming it to `enumeratorBody`?

streaming/src/main/scala/io/finch/streaming/package.scala
ImLiar: I guess there is no need for it

streaming/src/main/scala/io/finch/streaming/package.scala
ImLiar: True, agree on that

streaming/src/main/scala/io/finch/streaming/package.scala
vkostyukov: What about [syntax](https://github.com/travisbrown/iteratee/blob/master/core/src/main/scala/io/iteratee/Module.scala#L17-L21)?

ImLiar: Well, I guess it would save few symbols when lifting into a `Enumerator` but also add some overhead here and there, such as defining a monad, instead of picking it from imports.
Not sure if it worth it.

ImLiar:
@vkostyukov 
I think I'll add it under the same PR here

streaming/src/main/scala/io/finch/streaming/package.scala
ImLiar: Yep, instead I'll add an implicit identity `Enumeratee[Future, Buf, Buf]` in scope.
Just in case if someone would like to use this endpoint for uploading binary data

streaming/src/main/scala/io/finch/streaming/package.scala
vkostyukov: We can just point monad instance to one deinfed din catbird. I'm ok not doing this now (extending module), but I think it's a right thing to do at the end of the day.

streaming/src/main/scala/io/finch/streaming/package.scala
vkostyukov: I like it. Good idea!

streaming/src/main/scala/io/finch/streaming/package.scala
vkostyukov: Looks like we can just extend this: https://github.com/travisbrown/iteratee/blob/master/twitter/src/main/scala/io/iteratee/twitter/FutureModule.scala#L9

iteratee/src/main/scala/io/finch/iteratee/AsyncEncode.scala
```diff
@@ -0,0 +1,37 @@
...
+import io.iteratee.Enumerator
+
+trait AsyncEncode[A] {
+
+  type ContentType <: String
+
+  def apply(enumerator: Enumerator[Future, A], cs: Charset): Enumerator[Future, Buf]
+}
+
+object AsyncEncode extends EncodeInstances {
```
vkostyukov: Thinking about this more. Maybe it's an overkill?

Looks like we can go from `Enumerator[Future, A]` to `Enumerator[Future, Buf]` just by using `Encode.Aux[CT, A]` (it takes charset). Maybe we don't really need this type class? I understand it makes things symmetric, but I think if we remove it and rename `AsyncDecode` to something like `Enumerate` it should all line up pretty nicely.

iteratee/src/main/scala/io/finch/iteratee/package.scala
```diff
@@ -0,0 +1,76 @@
...
+    * an `Enumerator[Future, A]`. The returned [[Endpoint]] only matches chunked (streamed) requests.
+    */
+  def enumeratorJsonBody[A](implicit ad: Enumerate.Aux[A, Application.Json]): Endpoint[Enumerator[Future, A]] =
+    enumeratorBody[A, Application.Json]
+
+}
+
+trait IterateeInstances extends FutureModule  {
+
+  implicit def enumeratorToResponse[A, CT <: String](implicit
```
vkostyukov: Why this has to go in its own trait (do we need to de-prioritize it?)?

iteratee/src/main/scala/io/finch/iteratee/package.scala
```diff
@@ -0,0 +1,76 @@
...
+import shapeless.Witness
+
+/**
+  * Iteratee module
+  */
+package object iteratee extends IterateeInstances {
+
+  import syntax._
+
+  private[finch] def enumeratorFromReader(reader: Reader): Enumerator[Future, Buf] = {
```
vkostyukov: Can we embed this under the defined `enumeratorBody` endpoint? I don't think it's used anywhere else.

iteratee/src/main/scala/io/finch/iteratee/package.scala
```diff
@@ -0,0 +1,76 @@
...
+    reader.read(Int.MaxValue).intoEnumerator.flatMap({
```
vkostyukov: nit: No need for wrapping `()`. Just `.flatMap {}` should work.

iteratee/src/main/scala/io/finch/iteratee/package.scala
```diff
@@ -0,0 +1,76 @@
...
+      final override def item: RequestItem = items.BodyItem
+      final override def toString: String = "asyncJsonBody"
+    }
+  }
+
+  /**
+    * An evaluating [[Endpoint]] that reads a required chunked streaming JSON body, interpreted as
+    * an `Enumerator[Future, A]`. The returned [[Endpoint]] only matches chunked (streamed) requests.
+    */
+  def enumeratorJsonBody[A](implicit ad: Enumerate.Aux[A, Application.Json]): Endpoint[Enumerator[Future, A]] =
```
vkostyukov: Love it!

iteratee/src/main/scala/io/finch/iteratee/package.scala
```diff
@@ -0,0 +1,76 @@
...
+  implicit def enumeratorToResponse[A, CT <: String](implicit
+                                           e: Encode.Aux[A, CT],
+                                           w: Witness.Aux[CT]
+                                          ): ToResponse.Aux[Enumerator[Future, A], CT] = {
+    ToResponse.instance[Enumerator[Future, A], CT]((enum, cs) => {
+      val response = Response()
+      response.setChunked(true)
+      response.contentType = w.value
+      val writer = response.writer
+      val iteratee = foreachM((buf: Buf) => writer.write(buf))
```
vkostyukov: Should we add new-line delimiters between each streamed object? I think we do that for `AsyncStream`, but I'm not sure what the right answer here.

iteratee/src/main/scala/io/finch/iteratee/Enumerate.scala
```diff
@@ -0,0 +1,38 @@
...
+
+  type Aux[A, CT <: String] = Enumerate[A] {type ContentType = CT}
+
+  type Json[A] = Aux[A, Application.Json]
+}
+
+trait EnumerateInstances {
+  def instance[A, CT <: String]
+  (f: (Enumerator[Future, Buf], Charset) => Enumerator[Future, A]): Enumerate.Aux[A, CT] = new Enumerate[A] {
+    override type ContentType = CT
```
vkostyukov: nit: I don't think it overrides anything.

iteratee/src/main/scala/io/finch/iteratee/Enumerate.scala
```diff
@@ -0,0 +1,38 @@
...
+
+    override def apply(enumerator: Enumerator[Future, Buf], cs: Charset): Enumerator[Future, A] = f(enumerator, cs)
```
vkostyukov: Same here: no need for `override`.

iteratee/src/main/scala/io/finch/iteratee/package.scala
ImLiar: @vkostyukov for the sake of API cleanness and readability only, to separate "internal" stuff from user API

iteratee/src/main/scala/io/finch/iteratee/package.scala
ImLiar: Dunno, what if it's a buf binary stream?

ImLiar:
@vkostyukov I'd like to add tests under this PR. It'll take day or something. Is there any rush?

iteratee/src/main/scala/io/finch/iteratee/package.scala
ImLiar: Used for testing

iteratee/src/main/scala/io/finch/iteratee/package.scala
ImLiar: Also `circe-streaming` supports decoding of only arrays or objects, so they should be valid JSON structures. I assume that response structure and behaviour should be symmetrical

vkostyukov: Yeah, I think I'm fine with that - let's just make sure to document it.

ImLiar: Thinking about it more, it's gonna be invalid construction currently, just concatenated JSON objects. May be `\n` is not a bad idea in the end.

vkostyukov:
@ImLiar No rush at all! Just let me know when you think it's ready!

iteratee/src/main/scala/io/finch/iteratee/package.scala
vkostyukov: Yeah. That's what we do for `AsyncStream`-powered streaming: we distinguish JSON vs. non-JSON (it's quite possible with `Application.Json` content-type) streaming and insert newlines between each JSON object.

ImLiar: @vkostyukov another approach would be to wrap it inside of array, prepending and appending brackets and commas. Not sure how correct it would be though

vkostyukov: Yeah, but I don't think this considered idiomatic. What we decided to do with AsyncStream was Line-delimited JSON Streaming (see https://en.wikipedia.org/wiki/JSON_Streaming) and it looks like several other libraries adopt the same approach (examples: https://github.com/uhop/stream-json, http://doc.akka.io/docs/akka-http/10.0.7/scala/http/routing-dsl/source-streaming-support.html#json-streaming)

vkostyukov: FWIW, Twitter's streaming API is [new-line delimited as well](https://dev.twitter.com/streaming/overview/processing).

ImLiar: Yea, I know it, but confusing part for me is that `circe-streaming` works in a different manner.
So when we're talking about JSON streaming, request and response formats are going to be different.

vkostyukov: I think they don't have to. I'm pretty sure, circe-streaming (along with Jawn) can parse line-delimited JSON objects just fine.

vkostyukov: Actually, I take that back - I'm not so sure. Let me play with it a little bit. If it doesn't work, I think, it's our responsibility to add/fix it in Circe.

ImLiar: @vkostyukov 

https://github.com/circe/circe/blob/master/modules/streaming/src/main/scala/io/circe/streaming/ParsingEnumeratee.scala#L17

https://github.com/non/jawn/blob/master/README.md#parsing

My guess, we should have an ability to pass `ValueStream` mode (and other modes) as some configuration.

vkostyukov:
@ImLiar I have a quick question / topic for discussion. One of the pain points of `AsyncStream` was a difficulty of signaling mid-stream that we're no longer interested in sending/receiving data. Do you think there are ways for us to do that with iteratee? I'd imagine we need to use `ensure` (to close a reader) for an inbound `Enumerator` such that when we throw/raise (I guess, enumerate an error element, i.e., `Future.exception`) on it, it will close the reader. Do you think we can do the same for the outbound stream?

ImLiar:
@vkostyukov 

Could you clarify more, please.
Are we talking about "backpressure" in sense how it could be implemented in HTTP 1.1? You want to close a connection when there is an exception during `reader.read()` process?

ImLiar:
@rpless `io.iteratee` is already in use by `circe-streaming` anyway.

ImLiar:
Anyway, I think we have external dependency for now

iteratee/src/main/scala/io/finch/iteratee/package.scala
```diff
@@ -0,0 +1,93 @@
...
+
+/**
+  * Iteratee module
+  */
+package object iteratee extends IterateeInstances {
+
+  import syntax._
+
+  private[finch] def enumeratorFromReader(reader: Reader): Enumerator[Future, Buf] = {
+    reader.read(Int.MaxValue).intoEnumerator.flatMap {
```
vkostyukov: I mention that in the main thread, but I think we should do `ensure(reader.discard())` here.

vkostyukov:
@ImLiar I wouldn't think of that as a back-pressure it's more like an API for users to control the stream. And you're right - there is not much we can do in HTTP/1.1 besides just cutting the connection (and it's totally fine).

I think we should be fine on the outbound path and essentially failing an `Enumerator` should result in a closed connection. An inbound path is slightly more complicated, this is why I decided to start this discussion. Assuming we have `ensure(reader.discard())` in place, I was wondering if we can trigger that mid-stream somehow.

ImLiar:
@vkostyukov I'm afraid there is not much space for inbound stream.
Enumerator is immutable, enumeratee & iteratee have no backpressure, so there is no way to send exception back to enumerator. May be @travisbrown could give some advice on that.

vkostyukov:
I see. I think we still need `ensure(reader.discard())` to make sure we cut the connection if one of the `reader.read()` failed (I guess it should get it triggered).

vkostyukov:
@ImLiar Do you want to wait for https://github.com/circe/circe/pull/712 to make sure we match inbound/outbound streaming approach or you think it's fine to ship this now as is?

iteratee/src/main/scala/io/finch/iteratee/Enumerate.scala
```diff
@@ -0,0 +1,38 @@
...
+trait Enumerate[A] {
+
+  type ContentType <: String
+
+  def apply(enumerator: Enumerator[Future, Buf], cs: Charset): Enumerator[Future, A]
+}
+
+object Enumerate extends EnumerateInstances {
+
+  type Aux[A, CT <: String] = Enumerate[A] {type ContentType = CT}
```
vkostyukov: Can we add `implicitNoFound` here to produce more friendly compile error message?

iteratee/src/main/scala/io/finch/iteratee/package.scala
```diff
@@ -0,0 +1,93 @@
...
+          val req = input.request
+          EndpointResult.Matched(
+            input,
+            Rerunnable(Output.payload(decode(enumeratorFromReader(req.reader), req.charsetOrUtf8)))
+          )
+        }
+      }
+
+      final override def item: RequestItem = items.BodyItem
+      final override def toString: String = "asyncJsonBody"
```
vkostyukov: nit: I think it's named differently now.

iteratee/src/main/scala/io/finch/iteratee/package.scala
```diff
@@ -0,0 +1,93 @@
...
+  def enumeratorJsonBody[A](implicit ad: Enumerate.Aux[A, Application.Json]): Endpoint[Enumerator[Future, A]] =
+    enumeratorBody[A, Application.Json]
+
+}
+
+trait IterateeInstances extends LowPriorityInstances {
+
+  implicit def enumeratorToJsonResponse[A](implicit
+                                           e: Encode.Aux[A, Application.Json],
+                                           w: Witness.Aux[Application.Json]
```
vkostyukov: This indentation looks a bit weird to my taste.

ImLiar:
@vkostyukov I think it won't hurt nobody if we wait for circe first. Meanwhile could think about it closing of stream

iteratee/src/main/scala/io/finch/iteratee/Enumerate.scala
ImLiar: @vkostyukov Not sure that it would help because this implicit is hidden behind `ToResponse` resolution, so it's gonna be another `implicitNoFound` (for `ToResponse`)

json-test/src/main/scala/io/finch/test/JsonLaws.scala
```diff
@@ -62,4 +101,9 @@ object JsonLaws {
     new DecodeJsonLaws[A] {
       val decode: Decode.Json[A] = implicitly[Decode.Json[A]]
     }
+
+  def enumeratorDecoding[A : Enumerate.Json]: EnumerateJsonLaws[A] =
```
vkostyukov: `enumerating`?

iteratee/src/main/scala/io/finch/iteratee/package.scala
ImLiar: Not here, but after function call. There is no sense to add `.ensure` for every chunk

vkostyukov: Oh yeah, that makes sense.

iteratee/src/main/scala/io/finch/iteratee/package.scala
```diff
@@ -0,0 +1,91 @@
...
+    * an `Enumerator[Future, A]`. The returned [[Endpoint]] only matches chunked (streamed) requests.
+    */
+  def enumeratorBody[A, CT <: String](implicit decode: Enumerate.Aux[A, CT]): Endpoint[Enumerator[Future, A]] = {
+    new Endpoint[Enumerator[Future, A]] {
+      final def apply(input: Input): Endpoint.Result[Enumerator[Future, A]] = {
+        if (!input.request.isChunked) {
+          EndpointResult.Skipped
+        } else {
+          val req = input.request
+          def enum = enumeratorFromReader(req.reader).ensureEval(Eval.later(Future.value(req.reader.discard())))
```
vkostyukov: I'd probably prefer if `enumeratorFromReader` returned an enumerator that already ensures to discard the reader.

ImLiar:
@vkostyukov I've played a little bit more with Iteratee, looks like `.ensureEval` would be executed for `Enumerator` if inside of i.e. `Iteratee.foreachM` you return `Future.exception`. So yeah, it's totally possible to signal from iteratee to enumerator to close reader/writer.

ImLiar:
Also I think it would be nice to write some docs/cooking/examples before release. I think, I could handle it in this PR before circe release.

vkostyukov:
Looks like Circe [0.9-M1](https://github.com/circe/circe/releases/tag/v0.9.0-M1) is out. @ImLiar do you want to rebase this?

ImLiar:
@vkostyukov sure thing, gonna finish it today/tmrw

ImLiar:
@vkostyukov there is a binary incompatibility between latest circe and argonaut (different jawn version dependencies). It hurts! And also breaks tests

travisbrown:
@ImLiar I've got a [PR open](https://github.com/argonaut-io/argonaut/pull/280) for that on the Argonaut side.

Is there evidence that using argonaut-jawn is significantly better in the context of Finch?

travisbrown:
Also there is a jawn-argonaut artifact published from jawn that should be more or less a drop-in replacement for argonaut-jawn and has the most recent versions (also Erik just said he may be removing this in the longer term).

vkostyukov:
I don't see a reason to stick with Jawn in the finch-argonaut context. If it helps, let's just drop this dependency now.

ImLiar:
Removed argonaut-jawn. Tests passed. Yay.

vkostyukov:
Thanks a lot for all the hard work here, @ImLiar! I can "squash-and-merge" unless you're willing to update your branch (so it's one commit).

@rpless Mind taking a look as well?

rpless:
@vkostyukov +1 from me. Great work @ImLiar!

ImLiar:
Happy to help 😄

PR #814, opened 2017-07-20
Fix withToString recursion

ImLiar:
Current implementation ends up with infinite recursion. Fix argument name


vkostyukov:
Nice catch!

PR #815, opened 2017-07-27
Automatically mapping an Endpoint to A => ScalaFuture[B]

gabfssilva:
Basically making Mapper accept also Scala futures instead of force the developer convert it to Twitter futures.


gabfssilva:
I'm not really sure about how I could add more tests, neither if I added some unused implicit method, but, I'm not beeing able how to test every method that I added in Mapper. Does anyone know what test spec that I could add to solve the coverage issue? Thanks. ;)

ImLiar:
Hi @gabfssilva 
Please, check out following conversation: https://github.com/finagle/finch/issues/422

One of the ideas was to have this kind of mapper in a different package, so it'll be explicitly defined by the import block if user would like to use scala or twitter futures. 

I think it's not the best idea to mix em up together in a single package, because this implicit conversion between different futures will be totally hidden from user

vkostyukov:
Another idea (something I've been thinking of recently) is to define our syntax in terms of abstract `F[_]` for which there is a `cats.Monad` instance defined. This way we could easily distinguish `io.finch.syntax._` (default Twitter Futures modules) from `in.finch.syntax.scala._` (or even Monix). Similar to how iteratee, for examples, handles its modules.

ImLiar:
True, just thought about it earlier.

But the thing about iteratee that in the end you have your monad `F[_]` back. Here we always should have twitter future

vkostyukov:
It's fine. We're still going to be abstracted over the input type.

gabfssilva:
Well, I think the whole point of implicits is to abstract these conversions from the user. Personally, I'd not care to know what kind of Future I'm working with, I just want the library to accept it. Anyway, I'm not against separating it using another package either.

gabfssilva:
Since I'm not familiar with cats library, I just followed @ImLiar suggestion and separated it to another package. There's just a little annoying issue: when I use the import from io.finch.syntax.scala._, IntelliJ is mapping it as an used import, and I have no idea why this is happening. Anyway, it is working as expected. =P 
If you guys really want to, I could spend sometime to learn cats properly and do what @vkostyukov suggested.

PR #817, opened 2017-08-07
fix a wrong attribute name in user-guide.md

jinuk17:
modify bodyString to stringBody

I was very confused because of bodyString.
I could find stringBody instead of bodyString in io.finch.endpoint.Bodies.


vkostyukov:
Thanks @jinuk17! Nice catch!

Issue #806, opened 2017-06-16
Update docs as of 0.16


vkostyukov:

vkostyukov:
Also see #829, #820, #822.

PR #819, opened 2017-08-22
Update documentation about defining Exception Encoders for Json.

rpless:
I've updated the documentation for Encoding Json Exceptions as part of #806. Let me know if there are any changes that would make these docs more clear.


vkostyukov:
Nice! Thanks @rpless!

rpless:
@vkostyukov Is there anything I need to do to publish the new docs, or is that on your side?

vkostyukov:
@rpless I think you should be able to do that. Try `sbt docs/makeMicrosite`.

PR #828, opened 2017-08-29
Fix docs build outputting 'unused import' for every import

rpless:
Under Scala 2.12 `-Xlint` brings in some features of `-Xwarn-unused-imports` and this causes Tut to output these warnings in the docs. This removes `-Xlint` from the docs build only and fixes the #827. We may want to try to re-enable the other features of `-Xlint` at some point. 

Also fixed a couple compilation issues in the Enumerator docs.

Also @vkostyukov do you just run `sbt publishMicrosite` to push the docs to github pages or do you have another work flow?

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -141,7 +141,7 @@ lazy val docSettings = allSettings ++ Seq(
     "-doc-root-content", (resourceDirectory.in(Compile).value / "rootdoc.txt").getAbsolutePath
   ),
   scalacOptions ~= {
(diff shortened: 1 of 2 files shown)


vkostyukov:
This is awesome! Thanks for fixing it, @rpless!

Yeah, I just do `sbt docs/publishMicrosite`.

rpless:
@vkostyukov awesome. I've published the doc changes and did a pass over to make sure everything looks good.

PR #831, opened 2017-08-30
Fix Cookie#toString displaying 'param' and instead show 'cookie.

rpless:
I noticed that `Cookie`'s `toString` prints `param` to the string instead of `cookie` while working on some on Endpoint ADTs. Figured it better to fix separately.


vkostyukov:
Oh nice! Thanks @rpless!

PR #816, opened 2017-08-04
Rework Mapper as type class, introduce ToTwitterFuture type class

ImLiar:
Mapper now is not a magnet pattern, but a type class. Signature of `EndpointMapper.apply` has changed a little, but it's backward compatible. Strictly speaking, from API point of view nothing was changed, except that IDE should stop whining about "wrong" arguments: https://github.com/finagle/finch/issues/631

At the same time, added `ToTwitterFuture` type class to support different async mechanisms. There is an identity for twitter future and scala futures support in different package. Inspired by https://github.com/finagle/finch/pull/815

Potentially, this PR could open a way to support monix and scalaz tasks as parts of different subprojects: https://github.com/finagle/finch/pull/815#issuecomment-318662893

It's a working code, but without specs. If it's valid from "ideological" PoV I could write proper tests for Mapper.

PR #834, opened 2017-09-13
Add ToTwitterFuture type class

ImLiar:
Type class to support different async mechanisms. There is an identity for twitter future and scala futures support in different package.

Separate PR from https://github.com/finagle/finch/pull/816


ImLiar:
Summon @vkostyukov @rpless

core/src/main/scala/io/finch/syntax/ToTwitterFuture.scala
```diff
@@ -0,0 +1,26 @@
+package io.finch.syntax
+
+import cats.Monad
+import com.twitter.util.Future
+import io.catbird.util._
+
+/**
+  * Type class for conversion of some HKT (i.e. scala.concurrent.Future) to twitter future
```
vkostyukov: Wrap `s.c.Future` with ticks please.

vkostyukov: Also s/twitter future/Twitter Future.

core/src/main/scala/io/finch/syntax/ToTwitterFuture.scala
```diff
@@ -0,0 +1,26 @@
...
+  */
+trait ToTwitterFuture[F[_]] {
+
+  def M: Monad[F]
```
vkostyukov: Why is this necessary?

core/src/test/scala/io/finch/syntax/MapperSyntaxBehaviour.scala
```diff
@@ -0,0 +1,104 @@
+package io.finch.syntax
+
+import cats.syntax.functor._
+import com.twitter.finagle.http.Response
+import io.finch._
+import io.finch.{FinchSpec, Text}
+import org.scalacheck.Arbitrary
+import org.scalatest.prop.GeneratorDrivenPropertyChecks
+
+trait MapperSyntaxBehaviour extends FinchSpec with GeneratorDrivenPropertyChecks {
```
vkostyukov: I think we name these `XSpec`?

core/src/main/scala/io/finch/syntax/ToTwitterFuture.scala
ImLiar: To make generic testing via behaviour possible

core/src/test/scala/io/finch/syntax/MapperSyntaxBehaviour.scala
ImLiar: It's not a `Spec` strictly speaking, it's a trait with behaviour method

rpless:
LGTM. Great work @ImLiar!

core/src/main/scala/io/finch/syntax/ToTwitterFuture.scala
```diff
@@ -0,0 +1,26 @@
...
+import cats.Monad
+import com.twitter.util.Future
+import io.catbird.util._
+
+/**
+  * Type class for conversion of some HKT (i.e. `scala.concurrent.Future`) to `com.twitter.util.Future`
+  */
+trait ToTwitterFuture[F[_]] {
+
+  def M: Monad[F]
```
vkostyukov: Can we derive it in tests (if it's the only place it's used) instead?

ImLiar:
@vkostyukov done

core/src/test/scala/io/finch/syntax/MapperSyntaxBehaviour.scala
vkostyukov: Sure. But we have zero tests named `Behaviour` and something like 30 tests named `Spec`. Maybe we can rename it for the sake of consistency?

ImLiar:
@vkostyukov done

vkostyukov:
Thanks again, @ImLiar! Really nice work!

PR #837, opened 2017-09-27
Update docs to use value classes in Future conversion example.

amazzeo:
Updated example to use value classes similar to the example found at https://twitter.github.io/util/guide/util-cookbook/futures.html.


vkostyukov:
LGTM! Thanks @amazzeo!

PR #840, opened 2017-10-02
Replace string with int as test input for MapperSyntaxSpec

ImLiar:
Eventually, scalacheck passed `/` (right when it was merged to master) as an argument to `Input.get`.
Replaced `string` with `int` to prevent such situation.


vkostyukov:
Thanks @ImLiar!

Issue #833, opened 2017-09-04
Deprecate implicit syntax


vkostyukov:
It should be explicitly imported.

PR #843, opened 2017-10-06
Enable all syntax in io.finch.syntax.scala._

vkostyukov:
This allows to enable all syntax (including methods `get`, `post`, etc) by just importing `io.finch.syntax.scala._` (no need for `io.finch.syntax._` import).

@ImLiar Please, take a look.

### DIFF
--- a/core/src/main/scala/io/finch/syntax/ToTwitterFuture.scala
+++ b/core/src/main/scala/io/finch/syntax/ToTwitterFuture.scala
@@ -3,18 +3,14 @@ package io.finch.syntax
 import com.twitter.util.Future
 
 /**
-  * Type class for conversion of some HKT (i.e. `scala.concurrent.Future`) to `com.twitter.util.Future`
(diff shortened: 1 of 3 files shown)


ImLiar:
👍 
LGTM @vkostyukov

shado23:
This does however conflict with importing both io.finch._ and io.finch.syntax.scala._ as the former includes get etc. as well, making them ambiguous, which forces you to either just import io.finch.syntax.scala.scalaToTwitterFuture or pick out everything else individually from io.finch._.
That was surprising to me and I'm not so sure that such great usability.

vkostyukov:
Hey @shado23! That's a bummer for sure. We do, however, plan to deprecate implicit syntax (see #833) in 0.16 so it should be better after that.

Do you want to file a ticket about this particular problem?

shado23:
As long as it's on the radar that's fine by me.

PR #844, opened 2017-10-07
Deprecate old-style path matchers

### LINKED ISSUE #832: Deprecate old-style path matchers
We should stick w/ a single style and deprecate `int`, `string`, etc matchers.

vkostyukov:
Fixes #832.

### DIFF
--- a/core/src/main/scala/io/finch/endpoint/path.scala
+++ b/core/src/main/scala/io/finch/endpoint/path.scala
@@ -66,75 +66,90 @@ private[finch] trait Paths {
   /**
    * A matching [[Endpoint]] that reads an integer value from the current path segment.
    */
+  @deprecated("Use path[Int] instead", "0.16")
   val int: Endpoint[Int] = path[Int]
 
   /**
    * A matching [[Endpoint]] that reads a long value from the current path segment.
    */
+  @deprecated("Use path[Long] instead", "0.16")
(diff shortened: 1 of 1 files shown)


ImLiar:
Can you please elaborate abous the reason? Because it's quite noisy IMHO

vkostyukov:
@ImLiar here is my reasoning. Back in the days, when we had both `RequestReader` (i.e., `param`, `header`, etc) and `Endpoint` (`string`, `int`, etc) it was pretty obvious what `string`, for example, means given the `?` combinator used to separate "routing" and "request reading". Here is an example (that doesn't compile as of 0.11):

```scala
val a = get(int :: int ? param[Int])
```

It's clear that `int` is about routing as it goes before `?`.

Today, as we don't have `RequestReader`s so `int`, `string`, etc may go in any order in the composition. The following works just fine as you'd expect:

```scala
val a = get(int :: param[Int] :: int)
```
Given there is no clear indication where "routing" ends (i.e., there is no `?`), it's not so easy to reason about what `int` or `string` means as they all just endpoints. I believe it's not that obvious for newcomers if `int` is an integer param or an integer header or an integer path. By making it explicit (i.e., `path[Int]`) we should make it easier to _discover_ and reason about.

ImLiar:
@vkostyukov I see. Good point

Issue #781, opened 2017-05-30
Metrics for endpoints


sergeykolbasov:
Hey hey!

Could we use finagle-stats for reporting metrics of endpoints out-of-the-box?

**WHY**
To get http stats of individual endpoints for monitoring how performant they are

**HOW**
Use StatsReceiver provided by finagle-stats, since this project is dependant on finagle already

Of course, this could be achieved manually, adding some boilerplate, but I wonder if it's possible to make some endpoint that could write metrics for underlying endpoints

spockz:
I did something for this in our application. We add a metrics filter in finagle that puts a MetricContext case class on a local Context. Each endpoint definition is wrapped in a withEndpointMetrics method that used the toString to get the endpoint definition/path and put it in the metriccontext class. On the return in the metrics filter we used the metric context to lookup the name and used it for the metrics in the stats receiver.

Would you be interested if we made a PR out of this?

- Alessandro

ImLiar:
Well, I believe `Endpoint.transform` is a more natural way to measure things inside endpoints, but it's impossible to get response size with this method because I have to operate with `Output[A]`, not a `Response`..

BenWhitehead:
I created a filter to address this exact issue, and automatically add stats for any request that was handled by the finagle server in my finch-server project.

The finagle filter can be seen here: [StatsFilter](https://github.com/BenWhitehead/finch-server/blob/master/src/main/scala/io/github/benwhitehead/finch/filters/StatsFilter.scala) and can be woven into the server similar to the following:
```
Http.server.serve(":8080", StatsFilter andThen api.toServiceAs[Text.Plain])
```

I can open a PR to add this stats filter into finch itself if it is something people would find useful.

rpless:
One challenge I see with adding metrics at the endpoint level is that it is kinda difficult to choose a place in `Endpoint` or a single piece of boilerplate that makes sense given all of the different ways people use Finch (whether that's using finchrich, a controller style abstraction, or just sticking with functions). I don't know that there is a one size fits all solution in `Endpoint` or as boilerplate around endpoint. There are a number of ways I can thing of to implement metrics, although a Finagle `Filter` currently makes the most sense to me (I use a filter based on the one @BenWhitehead linked to).

[...]

PR #845, opened 2017-10-08
Metrics interface and its calls in ToService.apply

sergeykolbasov:
Here I'm again with #781 

After some good points in review of PR #835, I decided to try it in another way.

- Metrics interface and its default `Null` implementation
- Users could prove their own implementation through `Bootstrap`
- Default finagle-stats based implementation could be used by the user with `finch-metrics` module (different PR)

@vkostyukov @rpless @spockz take a look please

I still in doubts about how to properly count failures since `Output.failure` is encoded as successful `Future[Response]`.
Also `path: String` will be replaced when `Endpoint.Meta` introduced, giving even more flexibility.

### DIFF
--- a/core/src/main/scala/io/finch/ToService.scala
+++ b/core/src/main/scala/io/finch/ToService.scala
@@ -117,7 +117,7 @@ object ToService {
             val now = Stopwatch.timeMillis().toFloat
             metrics.countSuccess(path, response.statusCode)
             metrics.successTime(path, response.statusCode, now - start)
-            metrics.size(path, response.content.length.toFloat)
+            metrics.size(path, response.statusCode, response.content.length.toFloat)
           }).onFailure(e => {
             val now = Stopwatch.timeMillis().toFloat
             metrics.countFailure(path, e)
--- a/core/src/main/scala/io/finch/ToService.scala
+++ b/core/src/main/scala/io/finch/ToService.scala
@@ -1,10 +1,12 @@
 package io.finch
 
+import scala.annotation.implicitNotFound
+
 import com.twitter.finagle.Service
 import com.twitter.finagle.http.{Request, Response, Status, Version}
-import com.twitter.util.Future
+import com.twitter.util.{Future, Stopwatch}
 import io.finch.internal.currentTime
-import scala.annotation.implicitNotFound
+import io.finch.metrics.Metrics
 import shapeless._
 
 /**
@@ -35,7 +37,8 @@ trait ToService[ES <: HList, CTS <: HList] {
   def apply(
     endpoints: ES,
     includeDateHeader: Boolean,
-    includeServerHeader: Boolean
+    includeServerHeader: Boolean,
+    metrics: Metrics
   ): Service[Request, Response]
 }
 
@@ -78,7 +81,8 @@ object ToService {
     def apply(
         endpoints: HNil,
         includeDateHeader: Boolean,
-        includeServerHeader: Boolean): Service[Request, Response] = new Service[Request, Response] {
+        includeServerHeader: Boolean,
+        metrics: Metrics): Service[Request, Response] = new Service[Request, Response] {
 
       def apply(req: Request): Future[Response] = Future.value(
         conformHttp(Response(Status.NotFound), req.version, includeDateHeader, includeServerHeader)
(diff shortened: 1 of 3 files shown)


core/src/main/scala/io/finch/metrics/Metrics.scala
```diff
@@ -0,0 +1,53 @@
+package io.finch.metrics
+
+trait Metrics {
+
+  /**
+    * Count a call to endpoint by path and method
+    */
+  def countCall(path: String): Unit
```
spockz: How do I provide the HTTP Method to this method? Also, is this the materialized path or the matched path template/route?

core/src/main/scala/io/finch/metrics/Metrics.scala
```diff
@@ -0,0 +1,53 @@
...
+
+  /**
+    * Count
+    */
+  def successTime(path: String, code: Int, duration: Float): Unit
+
+  /**
+    * Count size of response
+    */
+  def size(path: String, s: Float): Unit
```
spockz: Nit: Why just `s` and not `size`? Also, why count the size of messages seperately from the responses?

core/src/main/scala/io/finch/ToService.scala
```diff
@@ -94,19 +98,33 @@ object ToService {
...
+          val path = underlying.toString
+          val start = Stopwatch.timeMillis().toFloat
+          metrics.countCall(path)
+          out.map(oa => conformHttp(
+            oa.toResponse(trA, trE),
+            req.version,
+            includeDateHeader,
+            includeServerHeader
+          )).run.onSuccess(response => {
+            val now = Stopwatch.timeMillis().toFloat
```
spockz: Wouldn't using nano's be more performant? (My takeaway from the akka-http code, may not apply here.)

core/src/main/scala/io/finch/ToService.scala
```diff
@@ -94,19 +98,33 @@ object ToService {
...
+            metrics.countSuccess(path, response.statusCode)
+            metrics.successTime(path, response.statusCode, now - start)
+            metrics.size(path, response.content.length.toFloat)
```
spockz: How does this size hold up when using streaming? Wouldn't this either force the stream, or be inaccurate?

core/src/main/scala/io/finch/metrics/Metrics.scala
ImLiar: Good question about a method. Currently, there is no way to get endpoint's method. It's somewhere inside of the string. As I said, `Endpoint.Meta` would make it better.

It's a template path.

core/src/main/scala/io/finch/metrics/Metrics.scala
ImLiar: Sorry, I don't get the question about size of message

core/src/main/scala/io/finch/ToService.scala
ImLiar: AFAIK for streams it's gonna be `0`

core/src/main/scala/io/finch/metrics/Metrics.scala
```diff
@@ -0,0 +1,53 @@
...
+    */
+  def failureTime(path: String, e: Throwable, duration: Float): Unit
+
+  /**
+    * Count success if request was processed without exceptions
+    */
+  def countSuccess(path: String, code: Int): Unit
+
+  /**
+    * Count
```
spockz: Really small nit, incomplete doc.

core/src/main/scala/io/finch/metrics/Metrics.scala
spockz: I was wondering why the size was counted through a seperate method instead of combining it like with `successTime` or `successCount`. Now there is now way to correlate the size of responses with the status code or response time.

core/src/main/scala/io/finch/metrics/Metrics.scala
spockz: Okay, I missed the `Endpoint.Meta` part in the PR description. :)

core/src/main/scala/io/finch/metrics/Metrics.scala
ImLiar: Ah, I see.

It makes sense

core/src/main/scala/io/finch/ToService.scala
ImLiar: Not exactly. It totally depends on how it's implemented in concrete JVM/OS. Usually, nano is slower

rpless:
I've thought a bit about this one a bit. My main concerns with this are:
- It doesn't let Finch users choose the metrics they want to export or what format they are in. and if they do want to, then they have to implement an instance of `ToService`
- as @spockz pointed, it will not metric other filters, which is desirable in certain use cases
- we have to maintain our own interface for metrics, which I'm not sure is necessarily ideal

However, your changes in `ToService` gave me an idea. What if we changed `ToService` to include a [Response#ctx](https://twitter.github.io/finagle/docs/com/twitter/finagle/http/Response.html#ctx:com.twitter.finagle.http.Response.Schema.Record) that was set to the result of `Endpoint#toString`? This would offer access to the formatted form of the path in filters. With this, users can can define a metric filter (or really use it for anything they see fit) using the endpoint's toString without being constrained by an interface we control or a metric format we define. We can also provide an example in the cookbook that outputs metrics the way you have them structured here and link out to gists for other metric libraries. I believe this also might also allow one to determine if a 404 is coming from an endpoint as opposed to the input not matching any of the defined endpoints.

@ImLiar what do you think?

ImLiar:
@rpless seems legit.

I'd still vote for providing some optional module to get metrics out of the box, but your proposal sounds good.

spockz:
@rpless Yes, [I agree](https://github.com/finagle/finch/issues/781#issuecomment-304957583).

vkostyukov:
Suppressed by #957.

Issue #631, opened 2016-08-25
Intellij syntax error: "Application does not take parameters"


chrisbenincasa:
Hello, I'm wondering if there's a way to get IntelliJ (I'm using v15.0.6 with Scala plugin 2.2.0) to be able to recognize the implicit conversion to a `Mapper` when applying an `Endpoint`.

Screenshot:
![screen shot 2016-08-25 at 11 43 32 am](https://cloud.githubusercontent.com/assets/1640671/17975822/2fc87c16-6ab9-11e6-9d00-cc0c7ff45a07.png)

The reported error is "Application does not take parameters". Of course, this error is spurious (the code compiles). IntelliJ just seems to have some trouble with the conversion of my functions to `Mapper`s (and probably also to `FnToProduct`). Not sure if this is a common issue for people using Finch in IntelliJ. Any help would be greatly appreciated!

vkostyukov:
Thanks @chrisbenincasa!

I'm (and I think everyone else is) aware of this problem. One workaround I use is to disable type-checker in IntelliJ (hit that green `[T]` in the right bottom corner). Although, this basically disables the everything in the IDE besides autocompletion. I think the best we can do here is to file a ticket against the Scala plugin (and vote it!) and mention this problem as false-alarm. They usually respond quite promptly.

I personally nearly hate this "magnet pattern" thing we use as `Mapper` to allow this DSL-like syntax. I have some ideas how to make it more explicit (extract the syntax layer into its own namespace) and perhaps simplify. Although, this doesn't mean IntelliJ shouldn't support this pattern. Lots of Scala libraries use it and I think, the Scala plugin is making a quite a decent progress in better supporting it.

od0:
Just a note - this appears to not be an issue in 2016.2, but resurfaces in 2016.3 (along with other unrelated Scala bugs)

Here is a link to download the latest good release, 2016.2.5: https://confluence.jetbrains.com/display/IntelliJIDEA/Previous+IntelliJ+IDEA+Releases

chrisbenincasa:
@od0 interesting. I'm on 2016.3.1 (Scala plugin 2016.3.5) and the error has changed slightly:

<img width="1187" alt="screen shot 2016-12-16 at 2 09 55 pm" src="https://cloud.githubusercontent.com/assets/1640671/21275128/57d9d518-c399-11e6-8cd9-b34fa9fa179d.png">

Basically that, but for every endpoint.

[...]

PR #849, opened 2017-10-10
Mapper type class

ImLiar:
Reincarnation of https://github.com/finagle/finch/pull/816

Replaced Mapper implicit conversion with implicit type parameter for some type `A`. It partially solve IDEA issue https://github.com/finagle/finch/issues/631 though for a case of a function with more than 1 argument and Future as a result, the output would be `Endpoint[mapper.Out]`:

```scala
//IntelliJ IDEA complains that right side is equal to Endpoint[mapper.Out]
val e: Endpoint[Int] = get(path[Int] :: path[Int]) { (x: Int, y: Int) => 
  Future.value(Output(x+y)) 
}
```


ImLiar:
ping @vkostyukov

vkostyukov:
@ImLiar should this be abandoned as #858 has been merged?

ImLiar:
@vkostyukov yes

closing it

PR #850, opened 2017-10-14
Call /hello endpoint in cookbook example 5

DieBauer:
Example 5 in the Finch cookbook (https://finagle.github.io/finch/cookbook.html) shows how to extract headers from a request, but the Input doesn't call the set-up endpoint, which results in
```
res31: Option[String] = None
```
The desired behaviour is to show that the headers are indeed extracted.
```scala
res1: Option[String] = Some(Headers: Map(foo -> bar))
```


vkostyukov:
Thanks @DieBauer!

PR #854, opened 2017-10-19
Update .toService compile error link

magicknot:


rpless:
Great catch! Thanks @magicknot! I'll re-publish the docs in a little bit.

Issue #784, opened 2017-06-09
Support 405


vkostyukov:
The idea is if the path is matched but the method is not we need to return 405 (we return 404 right now).

Just an implementation idea:

1. Extend `Input` to have a third member - `method`.
2. When the request comes in, set `Input.method` to request's method.
3. Change the implementation of `post`, `get`, etc combinators so they always match and update the `Input.method` member with their method value. This way, the `EndpointResult.remainder` contains the "remaining" input that carries path and the method.
4. In the `ToService` class, if the endpoint is matched, check `EndpointResult.remainder.method` against `request.method` and they are equal - run the endpoint; if they are different return 405.

vkostyukov:
I need to think about this more. Maybe things like `POST /users` and `GET /users` won't be working any more.

ImLiar:
@vkostyukov 
I have something in my mind. Going to take care of it

vkostyukov:
Fixed in #927.

PR #855, opened 2017-10-20
Export Endpoint.toString in a Response.Schema.Record

rpless:
A possible way to handle #781 that takes a similar approach to what @spockz described [here](https://github.com/finagle/finch/issues/781#issuecomment-304957583). The only difference is that instead of explicitly calling a method to include the path, we always have it available if the endpoint matches.

I'm suspicious of a couple things with this initial sketch, but wanted to solicit feedback on both the general approach and the following:
1. I don't think we want to re-evaluate `toString` on every endpoint call, so maybe pre-computing it is in order
2. I'm unsure if I'm using `StatsReceiver` correctly in the docs, as I use Dropwizard metrics at work and am unfamiliar with the gotchas of `StatsReceiver`.
3. Calling `apply` on the Record can yield a `IllegalStateException` if the endpoint didn't match will raise an exception

I would be interested in hearing what others think about this approach.

### DIFF
--- a/core/src/main/scala/io/finch/ToService.scala
+++ b/core/src/main/scala/io/finch/ToService.scala
@@ -2,8 +2,9 @@ package io.finch
 
 import com.twitter.finagle.Service
 import com.twitter.finagle.http.{Request, Response, Status, Version}
-import com.twitter.util.Future
+import com.twitter.util.{Future, Return, Throw}
 import io.finch.internal.currentTime
+
 import scala.annotation.implicitNotFound
 import shapeless._
 
@@ -74,8 +75,9 @@ object ToService {
     rep
   }
 
-  private def withPathContext(e: Endpoint[_])(rep: Response): Response = {
-    rep.ctx.updateAndLock(FinchContext.PathField, e.toString)
+  private def withPathContext(e: Endpoint[_], req: Request)(rep: Response): Response = {
+    req.ctx.updateAndLock(FinchContext.RequestPathField, e.toString)
+    rep.ctx.updateAndLock(FinchContext.ResponsePathField, e.toString)
     rep
   }
 
@@ -109,7 +111,10 @@ object ToService {
           req.version,
           includeDateHeader,
           includeServerHeader
-        )).run.onSuccess(withPathContext(underlying))
+        )).run.respond({
+          case Return(rep) => withPathContext(underlying, req)(rep)
+          case Throw(_) => req.ctx.updateAndLock(FinchContext.RequestPathField, underlying.toString)
+        })
 
         case _ => tsT(endpoints.tail, includeDateHeader, includeServerHeader)(req)
       }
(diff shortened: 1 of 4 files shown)


docs/src/main/tut/best-practices.md
```diff
@@ -207,6 +207,31 @@ object Main extends TwitterServer {
...
+import com.twitter.finagle.stats.StatsReceiver
+import com.twitter.util.Future
+import io.finch.FinchContext
+
+class MetricFilter(statsReceiver: StatsReceiver) extends SimpleFilter[Request, Response] {
+
+  def apply(request: Request, service: Service[Request, Response]): Future[Response] = {
+    service(request).onSuccess({ resp =>
+      val path: String = resp.ctx[String](FinchContext.PathField)
+      statsReceiver.counter(path.replace("/","_"), resp.statusCode.toString).incr()
```
ImLiar: Counter initialization on each request is relatively expensive operation. Ideally, they should be cached

core/src/main/scala/io/finch/ToService.scala
```diff
@@ -104,7 +109,7 @@ object ToService {
           req.version,
           includeDateHeader,
           includeServerHeader
-        )).run
+        )).run.onSuccess(withPathContext(underlying))
```
ImLiar: What about failure (exception) case?

rpless: My initial thought was to just leave it empty to signal that it was never set, but that will raise an exception when you attempt to get the field. Maybe this should be the field should be an `Option[String]` or some ADT.

docs/src/main/tut/best-practices.md
rpless: Yeah figured it be something like that, I can add something for that. Thanks!

core/src/main/scala/io/finch/ToService.scala
ImLiar: Maybe provide it on failure also?

rpless:
While trying to implement @ImLiar's suggestion of I've come across something. 

The first thing is that in all the metric PR's so far we've used `underlying` to generate the path, however this is not the endpoint that was matched, it is the entire Endpoint that is passed to toService. This means that every call will use the same metric name no matter what. I've added a failing test to show this.

The second is that its not possible to add the path to the Response's ctx because it doesn't exist. However, we can add it to the initial request's ctx. This feels weird to me, and I'm not sure we want to do this, assuming we can solve the first issue.

I need to think more on how to do this, because we're not accumulating the path, only traversing the remaining pieces of it until there is nothing left.

vkostyukov:
@rpless I had some ideas on how to populate remainder (i.e., `Input.Matched(rem)`) with an endpoint being matched (see #784). Maybe we can incorporate something like that here?

rpless:
@vkostyukov I think that approach might work, but we'd need a way to identify what parts of the path are dynamic and what parts are static. This might be another use case for the meta ADT. Basically if could we accumulated the parts of the ADT that match, we'd probably be in a better place for implementing one of these metrics PRs.

vkostyukov:
Suppressed by #957.

PR #858, opened 2017-11-03
Fix Intellij IDEA syntax error

ImLiar:
@vkostyukov 
Seems that this fix will finally solves https://github.com/finagle/finch/issues/631


vkostyukov:
Oh interesting. Does this mean we don't need #849 any more?

ImLiar:
@vkostyukov yep

vkostyukov:
Pretty cool @ImLiar! Thanks!

PR #859, opened 2017-11-04
Optimize Scala Future -> Twitter Future conversion

shado23:
By using a custom CallingThreadExecutionContext we can save a trip through the thread pool and two context switches.

This is an optimization that Twitter Future does by default, but for Scala Future this needs to be done manually (where appropriate).

I've also made ScalaToTwitterFuture a lazy val instead of a def, for a decrease in garbage creation and increase in performance.
(I've noticed that there seem to be quite a few other places in Finch where this might make sense.)

This leads to a significant increase in performance in a simple benchmark of an Endpoint returning `scala.concurrent.Future.successful { Ok() }` :
```
scala.concurrent.ExecutionContext.Implicits.global:
Running 30s test @ http://127.0.0.1:8080/health
  6 threads and 6 connections
  Thread Stats   Avg      Stdev     Max   +/- Stdev
    Latency   276.31us    1.29ms  67.06ms   97.69%
    Req/Sec     7.83k     2.53k   14.01k    68.91%
  1403466 requests in 30.10s, 120.46MB read
Requests/sec:  46627.27
Transfer/sec:      4.00MB

CallingThreadExecutionContext (def):
Running 30s test @ http://127.0.0.1:8080/health
  6 threads and 6 connections
  Thread Stats   Avg      Stdev     Max   +/- Stdev
    Latency   223.08us    0.96ms  28.54ms   97.22%
    Req/Sec    10.53k     3.10k   17.74k    65.24%
  1886399 requests in 30.10s, 161.91MB read
Requests/sec:  62671.91
Transfer/sec:      5.38MB

CallingThreadExecutionContext (lazy val):
Running 30s test @ http://127.0.0.1:8080/health
  6 threads and 6 connections
  Thread Stats   Avg      Stdev     Max   +/- Stdev
    Latency   208.32us    0.91ms  22.86ms   97.38%
    Req/Sec    11.01k     3.17k   17.78k    67.68%
  1977729 requests in 30.10s, 169.75MB read
Requests/sec:  65705.25
Transfer/sec:      5.64MB
```

### DIFF
--- a/core/src/main/scala/io/finch/syntax/scala/CallingThreadExecutionContext.scala
+++ b/core/src/main/scala/io/finch/syntax/scala/CallingThreadExecutionContext.scala
@@ -1,0 +1,9 @@
+package io.finch.syntax.scala
+
+import scala.concurrent.ExecutionContext
+
+private[scala] object CallingThreadExecutionContext extends ExecutionContext {
+  def execute(runnable: Runnable): Unit = runnable.run()
+
+  def reportFailure(throwable: Throwable): Unit = throwable.printStackTrace()
+}
--- a/core/src/main/scala/io/finch/syntax/scala/ScalaToTwitterFuture.scala
+++ b/core/src/main/scala/io/finch/syntax/scala/ScalaToTwitterFuture.scala
@@ -1,20 +1,22 @@
 package io.finch.syntax.scala
 
-import scala.concurrent.{ExecutionContext, Future => ScalaFuture}
+import scala.concurrent.{Future => ScalaFuture}
 import scala.util.{Failure, Success}
 
 import com.twitter.util.{Future, Promise}
 import io.finch.syntax.ToTwitterFuture
 
 trait ScalaToTwitterFuture {
-  implicit def scalaToTwitterFuture(implicit ec: ExecutionContext): ToTwitterFuture[ScalaFuture] =
+  implicit val scalaToTwitterFuture: ToTwitterFuture[ScalaFuture] =
     new ToTwitterFuture[ScalaFuture] {
       def apply[A](f: ScalaFuture[A]): Future[A] = {
-        val p = Promise[A]
+        val p = Promise[A]()
+
         f.onComplete {
           case Success(a) => p.setValue(a)
           case Failure(t) => p.setException(t)
-        }
+        }(CallingThreadExecutionContext)
+
         p
       }
     }


ImLiar:
But now user can not provide his own EC, right?
May be it would be better to mention this trick in documentation and leave it for users to decide.

shado23:
> But now user can not provide his own EC, right?
May be it would be better to mention this trick in documentation and leave it for users to decide.

The CallingThreadExecutionContext is only used for the conversion step, the Future itself will still be scheduled on whatever ExecutionContext the user provides. (In fact it would be a bad idea to force the future to run on the calling thread as it could be long running/blocking.) Essentially the conversion step will be run on whatever thread the future finishes executing on.

As far as I know there's no point in offering a choice of ExecutionContext for the conversion step, as any other choice is slower and there's no possible benefit of forcing an asynchronous boundary here. The conversion step itself is very simple (which is why the scheduling overhead of the future cause such a large performance hit).

This optimization is used for similar purposes by [futiles](https://github.com/johanandren/futiles/blob/master/src/main/scala/markatta/futiles/CallingThreadExecutionContext.scala) and [swave](https://github.com/sirthias/swave/blob/master/core/src/main/scala/swave/core/impl/CallingThreadExecutionContext.scala) and is well established.

ImLiar:
Okay, it makes sense

shado23:
I've added another commit special casing already completed futures. This yields an additional slight performance increase:
```
Running 30s test @ http://127.0.0.1:8080/health
  6 threads and 6 connections
  Thread Stats   Avg      Stdev     Max   +/- Stdev
    Latency   142.82us  470.46us  14.59ms   97.88%
    Req/Sec    11.22k     2.89k   17.18k    69.29%
  2010878 requests in 30.10s, 172.60MB read
Requests/sec:  66807.86
Transfer/sec:      5.73MB
```

core/src/main/scala/io/finch/syntax/scala/ScalaToTwitterFuture.scala
```diff
@@ -1,21 +1,30 @@
...
-import scala.concurrent.{ExecutionContext, Future => ScalaFuture}
+import scala.concurrent.{Future => ScalaFuture}
 import scala.util.{Failure, Success}
 
 import com.twitter.util.{Future, Promise}
 import io.finch.syntax.ToTwitterFuture
 
 trait ScalaToTwitterFuture {
-  implicit def scalaToTwitterFuture(implicit ec: ExecutionContext): ToTwitterFuture[ScalaFuture] =
+  implicit lazy val scalaToTwitterFuture: ToTwitterFuture[ScalaFuture] =
```
vkostyukov: Just make it a `val` (no need for `lazy`).

core/src/main/scala/io/finch/syntax/scala/ScalaToTwitterFuture.scala
```diff
@@ -1,21 +1,30 @@
...
     new ToTwitterFuture[ScalaFuture] {
       def apply[A](f: ScalaFuture[A]): Future[A] = {
-        val p = Promise[A]
-        f.onComplete {
-          case Success(a) => p.setValue(a)
-          case Failure(t) => p.setException(t)
+        if (f.isCompleted) {
```
vkostyukov: I'd prefer we kill this optimization as it's premature.

core/src/main/scala/io/finch/syntax/scala/CallingThreadExecutionContext.scala
```diff
@@ -0,0 +1,9 @@
+package io.finch.syntax.scala
+
+import scala.concurrent.ExecutionContext
+
+object CallingThreadExecutionContext extends ExecutionContext {
```
vkostyukov: This could be package-private.

core/src/main/scala/io/finch/syntax/scala/CallingThreadExecutionContext.scala
```diff
@@ -0,0 +1,9 @@
...
+  override def execute(runnable: Runnable): Unit = runnable.run()
```
vkostyukov: It doesn't override anything.

core/src/main/scala/io/finch/syntax/scala/CallingThreadExecutionContext.scala
```diff
@@ -0,0 +1,9 @@
...
+
+  override def reportFailure(throwable: Throwable): Unit = throwable.printStackTrace()
```
vkostyukov: It doesn't override anything.

core/src/main/scala/io/finch/syntax/scala/ScalaToTwitterFuture.scala
shado23: Done.

core/src/main/scala/io/finch/syntax/scala/ScalaToTwitterFuture.scala
shado23: Ok. It's unlikely to make a big difference under real-world circumstances anyway (i.e. futures that do more than return a static value).

core/src/main/scala/io/finch/syntax/scala/CallingThreadExecutionContext.scala
shado23: Done.

core/src/main/scala/io/finch/syntax/scala/CallingThreadExecutionContext.scala
shado23: You're right, this is unnecessary.

core/src/main/scala/io/finch/syntax/scala/CallingThreadExecutionContext.scala
shado23: Removed.

vkostyukov:
@rpless Mind taking a quick look?

shado23:
> Yeah, @vkostyukov this looks good to me. Its a nice bump in performance. Thanks @shado23!

Your're welcome.

PR #861, opened 2017-11-06
Update finagle to 17.10.0

ImLiar:
New version was released few days ago

Also update catbird & iteratee

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -7,7 +7,7 @@ lazy val buildSettings = Seq(
   crossScalaVersions := Seq("2.11.11", "2.12.3")
 )
 
-lazy val finagleVersion = "7.1.0"
+lazy val finagleVersion = "17.10.0"
 lazy val twitterServerVersion = "1.32.0"
 lazy val circeVersion = "0.9.0-M1"
 lazy val circeJacksonVersion = "0.9.0-M1"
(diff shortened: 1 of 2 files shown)


travisbrown:
If the only thing blocking the upgrade to Cats 1.0.0-RC1 is circe-jackson I can try to get that published today.

ImLiar:
@travisbrown it would be awesome! Thanks

vkostyukov:
I think we also need circe-streaming (which is being extracted from Circe) to do a Cats upgrade.

vkostyukov:
@ImLiar Looks like we need to depend on logging explicitly (based on https://travis-ci.org/finagle/finch/jobs/297979347). I'd just replace `log()` with `println` for now.

travisbrown:
@vkostyukov You're right, sorry—I'll try to sort out https://github.com/travisbrown/iteratee/pull/218 tonight. In the meantime circe-jackson's ready.

ImLiar:
@vkostyukov 

Yep, I'm on it with fixing a logger

vkostyukov:
Thanks @ImLiar! Merging this in.

ImLiar:
@vkostyukov there are still some warning (potential binary incompatibilities) because `circe-streaming` wasn't releassed yet. Waiting for iteratee tests fix

vkostyukov:
Yeah. Looks like we also need to bump discipline's version.

PR #865, opened 2017-11-12
Decouple multipart attributes form params

### LINKED ISSUE #860: Error in post json request containing character '='.
Finch Version - 0.15.0
Random calls failing with exception in org.jboss.netty.handler.codec.http.multipart.HttpPostRequestDecode.
These calls working when I use HTTP PUT call with the same signature.
Observation-
Call fails when JSON contains '=' character.
Incidentally, these calls are working when JSON data is in the pretty format.
Is there any way by which I can disable multipart decoding for post call.

vkostyukov:
This is aiming to fix #860 as well as make Finch's multipart support more explicit.

Previously:

- attributes are read via `param` (such that multipart decoding is triggered on nearly every request)
- file uploads via `fileUpload`

Now:

- attributes are read via `multipartAttribute`
- file uploads via `multipartFileUpload`


ImLiar:
Yep, it looks better. It was always a little confusing for me that finch treats query parameters and multipart one in the same way.

rpless:
Yeah, I agree, its much less confusing.

Issue #786, opened 2017-06-09
Deprecate .as[A] method for decoding


vkostyukov:
The `.as[A]` method should only be supported for `HList => A` conversion (through the `Generic` instance). As far as the API changes go, I see the following migration schema:

- `param("foo").as[A]` becomes `param[A]("foo")`
- Should we do the same for headers? How popular is that to read non-string values out of headers?

ImLiar:
Going to pick up this one if it's ok

spockz:
> Should we do the same for headers? How popular is that to read non-string values out of headers?

The only cases I know are authentication and.., just authentication.

rpless:
Yes, iirc, `.as` for `HList => A` should stay the same. I think we should include headers with this, because would be a more consistent API and there are use cases for numbers and (maybe UUIDs) being in headers.

vkostyukov:
Yeah, I guess the idea end state we want to be in is to keep `as[Foo]` as only a syntax extension for `HList => Foo` conversion.

I think I played with it in the past and turns out supporting a deprecation cycle would be a huge pain.

vkostyukov:
Looks like we also need to add `multipartAttribute` to the list.

vkostyukov:
I'm closing this for now as I'm not sure we have to wait for `multipartAttribute` to support a new pattern. The `multipart*` API is quite new and I think it's actually quite fine to wait a couple of more iterations before evolving it.

PR #873, opened 2017-11-27
Introduce ResolveDecodeEntity

vkostyukov:
I've been playing with this idea of supporting an API migration from `param("foo").as[Int]` to `param[Int]("foo")`. What makes it complicated is that we have to maintain a support for just `param("foo")` and have it resolved into `Endpont[String]`. See #786.

This PR introduces a new, temporary type-class `ResolveDecodeEntity` that maps a given type parameter (`Nothing` in the case of `param("foo")` onto a `DecodeEntity` instance). Here is how it works:

```scala
scala> import io.finch._
import io.finch._

scala> param("foo")
res0: io.finch.Endpoint[io.finch.internal.ResolveDecodeEntity.nothing.Out] = param(foo)

scala> param[Int]("foo")
res1: io.finch.Endpoint[Int] = param(foo)

scala> param("foo").as[Int]
res2: io.finch.Endpoint[Int] = param(foo)

scala> param("foo"): Endpoint[String]
res3: io.finch.Endpoint[String] = param(foo)
```

Note: `res0` type looks rather confusing but it's perfectly convertible into `String` (as shown in `res3`). I don't believe this should be a problem for any application's code.

I'd appreciate any feedback people may have about this.

### DIFF
--- a/core/src/main/scala/io/finch/endpoint/param.scala
+++ b/core/src/main/scala/io/finch/endpoint/param.scala
@@ -1,18 +1,26 @@
 package io.finch.endpoint
 
 import cats.data.NonEmptyList
+import com.twitter.util.{Return, Throw}
 import io.finch._
-import io.finch.internal.Rs
+import io.finch.internal.{ResolveDecodeEntity, Rs}
+import scala.reflect.ClassTag
 
-private abstract class Param[A](name: String) extends Endpoint[A] {
+private abstract class Param[A, B](
+    name: String,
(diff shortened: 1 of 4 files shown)


ImLiar:
Oh, I also did some work last night on this issue. Such a coincidence.

Anyway, I think this new type class is not required since simple prioritization of implicits should solve the problem with default type.

vkostyukov:
Abandoning this in favor of #875.

Issue #787, opened 2017-06-09
Introduce Error.Source


vkostyukov:
Basically a replacement for `RequestItem`.

This seems related to #581, but I still think we need both: `Error.Source` and `Endpoint.Meta`. We will use the first only in errors and we'll use the second for things like Swagger generation.

rpless:
I've been playing with an implementation of `Endpoint.Meta` and I see it working as a replacement for `RequestItems`, but I'm not quite sure about `Error.Source`. Was there some deeper modeling you were thinking of or was the idea to just change were `Error.scala` lives? I don't have a problem with either, just need more info to include that.

PR #875, opened 2017-11-27
Deprecate .as[A] for Endpoint[String]

ImLiar:
https://github.com/finagle/finch/issues/786

- Add type parameters for all params and headers methods
- Introduce priority in `DecodeEntity` to make `String` a default case for missing type in `param("foo")`
- Change specs to work with different types instead of strings


vkostyukov:
I think it's really cool @ImLiar! Do you have an idea why exactly this works? I'm sort of lost here. I thought if we don't provide a type explicitly it's inferred as `Nothing` yet it correctly picks `String`.

```scala
scala> param("foo") // print
   io.finch.`package`.param[String]("foo")(io.finch.DecodeEntity.decodeString, `package`.this.materializeClassTag[String]()) // : io.finch.Endpoint[String]

scala> param[Int]("foo") // print
   io.finch.`package`.param[scala.Int]("foo")(io.finch.DecodeEntity.decodeInt, `package`.this.materializeClassTag[Int]()) // : io.finch.Endpoint[Int]
```

core/src/main/scala/io/finch/DecodeEntity.scala
```diff
@@ -12,32 +12,21 @@ trait DecodeEntity[A] {
...
-   */
-  implicit def decodeFromGeneric[A](implicit
-    gen: Generic.Aux[A, String :: HNil]
-  ): DecodeEntity[A] = instance(s => Return(gen.from(s :: HNil)))
+}
 
-  implicit val decodeString: DecodeEntity[String] = instance(s => Return(s))
+trait LowPriorityDecode extends LowLowPriorityDecode {
+
+  implicit def decodeOptional[A](implicit d: DecodeEntity[A]): DecodeEntity[Option[A]] =
```
vkostyukov: Do we actually need this?

core/src/main/scala/io/finch/DecodeEntity.scala
```diff
@@ -54,3 +43,27 @@ object DecodeEntity {
...
+
+
+  /**
+    * Creates a [[Decode]] from [[shapeless.Generic]].
+    *
+    * Note: This is mostly a workaround for `Endpoint[String].as[CaseClassOfASingleString]`,
+    *       because by some reason, compiler doesn't pick `ValueEndpointOps` for
+    *       `Endpoint[String]`.
+    */
+  implicit def decodeFromGeneric[A, H <: HList, E](implicit
```
vkostyukov: What's the reason for this to be a `LowLow` priority?

core/src/main/scala/io/finch/endpoint/header.scala
```diff
@@ -1,39 +1,46 @@
...
 
+import scala.reflect.ClassTag
+
+import com.twitter.util.{Return, Throw}
 import io.finch._
 import io.finch.internal._
 import io.finch.items._
 
-private abstract class Header[A](name: String) extends Endpoint[A] {
+private abstract class Header[A](name: String)(implicit d: DecodeEntity[A], tag: ClassTag[A]) extends Endpoint[A] {
```
vkostyukov: Let's pass `DecodeEntity` and `ClassTag` as explicit parameters (as I did in #873). There is no value for them to be implicit.

core/src/main/scala/io/finch/endpoint/header.scala
```diff
@@ -1,39 +1,46 @@
...
-  protected def present(input: Input, value: String): Endpoint.Result[A]
+  protected def present(input: Input, value: A): Endpoint.Result[A]
 
   final def apply(input: Input): Endpoint.Result[A] =
     input.request.headerMap.getOrNull(name) match {
       case null => missing(input, name)
-      case value => present(input, value)
+      case value => d(value) match {
+        case Return(r) => present(input, r)
+        case Throw(e) => notParsed(self, input, e, tag)
```
vkostyukov: I'd prefer we introduce `Rs.paramNotParsed` so we can stop relying on request items (I really really want them to go away in a long run; #787 should replace them). See #873 for example.

core/src/test/scala/io/finch/BodySpec.scala
```diff
@@ -13,12 +13,12 @@ class BodySpec extends FinchSpec {
 
   def withBody(b: String): Input = Input.post("/").withBody[Text.Plain](Buf.Utf8(b))
 
-  checkAll("Body[Int]", EntityEndpointLaws[Int](stringBodyOption)(withBody).evaluating)
-  checkAll("Body[Long]", EntityEndpointLaws[Long](stringBodyOption)(withBody).evaluating)
-  checkAll("Body[Boolean]", EntityEndpointLaws[Boolean](stringBodyOption)(withBody).evaluating)
-  checkAll("Body[Float]", EntityEndpointLaws[Float](stringBodyOption)(withBody).evaluating)
-  checkAll("Body[Double]", EntityEndpointLaws[Double](stringBodyOption)(withBody).evaluating)
-  checkAll("Body[UUID]", EntityEndpointLaws[UUID](stringBodyOption)(withBody).evaluating)
+  /*checkAll("Body[Int]", EntityEndpointLaws[Int](textBodyOption)(withBody).evaluating)
```
vkostyukov: Why these are commented out?

core/src/test/scala/io/finch/EntityEndpointLaws.scala
```diff
@@ -12,13 +12,12 @@ trait EntityEndpointLaws[A] extends Laws with MissingInstances with AllInstances
 
   def decoder: DecodeEntity[A]
   def classTag: ClassTag[A]
-  def endpoint: Endpoint[Option[String]]
-  def serialize: String => Input
+  def endpoint: Endpoint[Option[A]]
```
vkostyukov: Maybe we can keep the tests unchanged for now?

generic/src/main/scala/io/finch/generic/FromParams.scala
```diff
@@ -38,7 +38,7 @@ private[generic] object Extractor extends Poly1 {
     dh: DecodeEntity[V],
     ct: ClassTag[V]
   ): Case.Aux[String, Endpoint[Option[V]]] = at[String] { key =>
-    paramOption(key).as(dh, ct)
+    paramOption(key)
```
vkostyukov: Let's pass a `V` type-param explicitly.

generic/src/main/scala/io/finch/generic/FromParams.scala
```diff
@@ -59,6 +59,6 @@ private[generic] object Extractor extends Poly1 {
     dh: DecodeEntity[V],
     ct: ClassTag[V]
   ): Case.Aux[String, Endpoint[V]] = at[String] { key =>
-    param(key).as(dh, ct)
+    param(key)
```
vkostyukov: Let's pass a type-param explicitly.

ImLiar:
@vkostyukov actually, this is a way how shapeless works most of the time actually.
The thing is that compiler infer types based on passed arguments, including implicits. So if implicit is resolved successfully, it picks up all the types on its own. The only thing here is to make a priority for resolution, that's it.

core/src/main/scala/io/finch/DecodeEntity.scala
```diff
@@ -54,3 +40,22 @@ object DecodeEntity {
     else Try(UUID.fromString(s))
   )
 }
+
+trait LowLowPriorityDecode {
```
vkostyukov: `LowPriorityDecode`?

core/src/main/scala/io/finch/DecodeEntity.scala
```diff
@@ -12,32 +12,18 @@ trait DecodeEntity[A] {
...
-   *       because by some reason, compiler doesn't pick `ValueEndpointOps` for
-   *       `Endpoint[String]`.
-   */
-  implicit def decodeFromGeneric[A](implicit
-    gen: Generic.Aux[A, String :: HNil]
-  ): DecodeEntity[A] = instance(s => Return(gen.from(s :: HNil)))
+}
 
-  implicit val decodeString: DecodeEntity[String] = instance(s => Return(s))
+trait LowPriorityDecode extends LowLowPriorityDecode {
```
vkostyukov: `HighPriorityDecode`?

core/src/test/scala/io/finch/BodySpec.scala
```diff
@@ -11,14 +12,20 @@ class BodySpec extends FinchSpec {
...
-  checkAll("Body[Boolean]", EntityEndpointLaws[Boolean](stringBodyOption)(withBody).evaluating)
-  checkAll("Body[Float]", EntityEndpointLaws[Float](stringBodyOption)(withBody).evaluating)
-  checkAll("Body[Double]", EntityEndpointLaws[Double](stringBodyOption)(withBody).evaluating)
-  checkAll("Body[UUID]", EntityEndpointLaws[UUID](stringBodyOption)(withBody).evaluating)
+  implicit def decodeBodyEntity[A](implicit de: DecodeEntity[A]): Decode.Text[A] = Decode.text[A]((b, _) => {
+    val Buf.Utf8(str) = b
+    de(str)
+  })
+
+  checkAll("Body[Int]", EntityEndpointLaws[Int](textBodyOption)(withBody).evaluating)
```
vkostyukov: So I think `textBodyOption` doesn't have to follow the laws from `EntityEndpointLaws`. Maybe `stringBodyOption` does (I'm not sure)? I'd probably vote for just removing these checks for `strinbBodyOption` for now.

core/src/main/scala/io/finch/Endpoint.scala
```diff
@@ -478,6 +478,7 @@ object Endpoint {
    * The resulting endpoint will fail when type conversion fails.
    */
   implicit class StringEndpointOps(val self: Endpoint[String]) extends AnyVal {
+    @deprecated(s"Use type parameter instead for a corresponding endpoint (param[A], header[A], ...)", "0.16")
```
vkostyukov: We should also deprecate `as` under `StringOptionEndpointOps`, `StringSeqEndpointOps`, and `StringNelEndpointOps`.

core/src/main/scala/io/finch/endpoint/param.scala
```diff
@@ -1,80 +1,94 @@
...
-  protected def missing(input: Input, name: String): Endpoint.Result[A]
-  protected def present(input: Input, value: Iterable[String]): Endpoint.Result[A]
+  protected def missing(input: Input, name: String): Endpoint.Result[F[A]]
+  protected def present(input: Input, value: Iterable[A]): Endpoint.Result[F[A]]
 
-  def apply(input: Input): Endpoint.Result[A] = input.request.params.getAll(name) match {
+  def apply(input: Input): Endpoint.Result[F[A]] = input.request.params.getAll(name) match {
     case value if value.isEmpty => missing(input, name)
-    case value => present(input, value)
+    case value => Try.collect(value.map(d.apply).toSeq) match {
```
vkostyukov: I think we should keep collecting the errors here. Right now we return just the first observed failure. See `StringNelEndpointOps` for an inspiration.

core/src/main/scala/io/finch/Endpoint.scala
ImLiar: What about `ValueEndpointOps`?

I've made a generic `DecodeEntity` for this case:
https://github.com/ImLiar/finch/blob/47d05844f323ac1bc5cfda1208ba302f6540ccfe/core/src/main/scala/io/finch/DecodeEntity.scala#L56

So it's possible to use `param[Foo]` now.

core/src/test/scala/io/finch/EntityEndpointLaws.scala
ImLiar: I think it should be ok to slightly change it toward new syntax

core/src/main/scala/io/finch/Endpoint.scala
vkostyukov: Yeah, let's keep it. I think we still want to support `param[String].as[Foo]` yet I'm not sure we have to support `param[Foo]` (unless there is `DecodeEntity[Foo]` available).

core/src/main/scala/io/finch/DecodeEntity.scala
```diff
@@ -54,3 +40,22 @@ object DecodeEntity {
...
+    * Creates an [[DecodeEntity]] instance from a given function `String => Try[A]`.
+    */
+  def instance[A](fn: String => Try[A]): DecodeEntity[A] = new DecodeEntity[A] {
+    def apply(s: String): Try[A] = fn(s)
+  }
+
+  /**
+    * Creates a [[Decode]] from [[shapeless.Generic]] for single value case classes.
+    */
+  implicit def decodeFromGeneric[A, H <: HList, E](implicit
```
vkostyukov: My guess is if we remove this, things like `param[String].as[Foo]` will stop working.

vkostyukov:
LGTM! Thanks @ImLiar!

I'd probably add the same support for `mutlipartAttribute*` though, but I'm happy to do that in a separate PR.

core/src/main/scala/io/finch/DecodeEntity.scala
ImLiar: Yes. Also `param[Foo]` won't work anymore.

core/src/main/scala/io/finch/Endpoint.scala
ImLiar: This implicit does exactly the same. It lifts parameter into a single value case class `Foo` for free.

ImLiar:
@vkostyukov no problem, I'll add multipart* support here

core/src/main/scala/io/finch/Endpoint.scala
vkostyukov: Yeah I know. There was a comment on that instances saying it's a workaround for `param[String].as[Foo]`, which otherwise doesn't work (even though there is `ValueEndpointOps` that's basically a noop right now).

core/src/main/scala/io/finch/DecodeEntity.scala
vkostyukov: I'm less worried about `param[Foo]` (that said, I'm not 100% sure we need/want to support that), but we definitely need / want to keep.

ImLiar:
upd: Seems that multipart support is not that trivial mostly because of a mix with file upload.

I'll land separate PR for it.

vkostyukov:
Sounds good to me, @ImLiar!

@rpless Want to take a quick look before I merge this?

core/src/main/scala/io/finch/endpoint/header.scala
```diff
@@ -1,46 +1,54 @@
...
+  protected def present(input: Input, value: A): Endpoint.Result[F[A]]
 
-  final def apply(input: Input): Endpoint.Result[A] =
+  final def apply(input: Input): Endpoint.Result[F[A]] =
     input.request.headerMap.getOrNull(name) match {
       case null => missing(input, name)
-      case value => present(input, value)
+      case value => d(value) match {
+        case Return(r) => present(input, r)
+        case Throw(e) => EndpointResult.Matched(input, Rs.paramNotParsed(name, tag, e))
```
jsancio: Please see https://github.com/finagle/finch/issues/929.

PR #879, opened 2017-12-08
Split files and multipart attributes and enable type parameters for attributes

ImLiar:
Related to https://github.com/finagle/finch/issues/786

API should stay the same, just a slight refactoring


core/src/main/scala/io/finch/endpoint/attribute.scala
```diff
@@ -0,0 +1,124 @@
...
+
+import cats.Id
+import cats.data.NonEmptyList
+import com.twitter.finagle.http.exp.{Multipart => FinagleMultipart}
+import com.twitter.util._
+import io.catbird.util.Rerunnable
+import io.finch._
+import io.finch.internal.Rs
+
+private abstract class Attribute[F[_], A](val name: String, val d: DecodeEntity[A], val tag: ClassTag[A])
```
vkostyukov: I'd vote we keep both `Attribute` and `FileUpload` under the same file, `multipart.scala`. We do that for params.

core/src/main/scala/io/finch/package.scala
```diff
@@ -15,6 +15,7 @@ package object finch extends Endpoints
     sealed abstract class RequestItem(val kind: String, val nameOption:Option[String] = None) {
       val description = kind + nameOption.fold("")(" '" + _ + "'")
     }
+    final case class AttributeItem(name: String) extends RequestItem("attribute", Some(name))
```
vkostyukov: Why do you think we have to introduce a new request item for attributes? We've been using `ParamItem` until now and I actually think it's perfectly reasonable.

core/src/main/scala/io/finch/endpoint/attribute.scala
```diff
@@ -0,0 +1,124 @@
...
+  extends Endpoint[F[A]] {
+
+  protected def missing(name: String): Rerunnable[Output[F[A]]]
+  protected def present(value: NonEmptyList[A]): Rerunnable[Output[F[A]]]
+  protected def notparsed(errors: NonEmptyList[Throwable]): Rerunnable[Output[F[A]]]
```
vkostyukov: `unparsed`?

core/src/main/scala/io/finch/endpoint/attribute.scala
```diff
@@ -0,0 +1,124 @@
...
+    protected def present(value: NonEmptyList[A]): Rerunnable[Output[Seq[A]]] = Rs.payload(value.toList)
+  }
+
+  trait NonEmpty[A] { _: Attribute[NonEmptyList, A] =>
+    protected def missing(name: String): Rerunnable[Output[NonEmptyList[A]]] = Rs.attributeNotPresent(name)
+    protected def present(value: NonEmptyList[A]): Rerunnable[Output[NonEmptyList[A]]] = Rs.payload(value)
+  }
+}
+
+private trait SingleItem[F[_], A] { _: Attribute[F, A] =>
```
vkostyukov: Let's move both `SingleItem` and `MultipleItems` under the `Attribute` object. Also, what do you think of renaming them to `SingleError` and `MultipleErrors` correspondingly?

project/build.properties
```diff
@@ -1 +1 @@
-sbt.version=1.0.3
+sbt.version=1.0.4
```
vkostyukov: Also you should rebase - I think we already upgraded sbt.

core/src/main/scala/io/finch/package.scala
ImLiar: Well, my idea behind it is that they are different entities, so different items. But if you have a strong opinion on that, I could remove it

ImLiar:
@vkostyukov done

core/src/main/scala/io/finch/package.scala
vkostyukov: My personal preference is to keep using `ParamItem` just b/c

- We've been using it before and
- I really want to see us removing `RequestItem` thing all-together at some point (and possibly replacing with something else)

vkostyukov:
Thanks again, @ImLiar! This looks good to me.

Issue #510, opened 2016-01-24
Negotiate content-type


vkostyukov:
As discussed [on Gitter](https://gitter.im/finagle/finch?at=56a3c29b6b6468374a09197e), I'd be useful to pick a response's content-type depending on request's content-type.

vkostyukov:
Implemented in #905.

PR #882, opened 2017-12-23
Content negotiation based on Accept header

ImLiar:
Related to https://github.com/finagle/finch/issues/510

This is a very naive and basic implementation, based on the idea of using `Bootstrap.serve` with Coproduct of content-types:
```
Bootstrap.serve[Text.Plain :+: Application.Json :+: CNil](...).toService
```

It checks if all required `ToResponse` instances are presented and picks one of them depending on `Accept:` header.

Current behavior:

* Parse `Request.accept` sequence and wrap it into `Seq[Accept]`. Only primary and subtype are preserved.
* Loop through the coproduct of selected content-types looking for a match
* If match is **found**, use corresponding `ToResponse`
* If match is **not found**, use last `ToResponse` instance from the coproduct
* If accept header is missing, empty or all entities are parsed incorrectly, use first `ToResponse` instance from the coproduct


ImLiar:
There is no tests yet, but I'd be happy to get some feedback on this idea

vkostyukov:
Hey @ImLiar! I took a quick look and I think it's a reasonable approach - I like how it's built upon what was already available and doesn't really change the current API/behavior.

I'm going to take a deeper look later today so we can start a more concrete discussion. I don't see a reason why this couldn't be shipped as part of 0.17 (given we iterate on the current version). Thanks! This is a nice work!

core/src/main/scala/io/finch/ContentTypeNegotiation.scala
```diff
@@ -0,0 +1,48 @@
+package io.finch
+
+import io.finch.internal.Accept
+import shapeless._
+
+trait ContentTypeNegotiation[ContentType] {
```
vkostyukov: Why does it have to be this order for typer parameters? Most of our components define content-type as a type-member vs a type-parameter here.

vkostyukov: How do you feel about naming this entity `NegotiateToResponse` so it takes an `Accept` and returns a `ToResponse` instance for it.

core/src/main/scala/io/finch/ToService.scala
```diff
@@ -111,3 +87,58 @@ object ToService {
     }
   }
 }
+
+trait LowPriorityToService {
+
+  implicit def hlistNegTS[A, EH <: Endpoint[A], ET <: HList, CTH <: Coproduct, CTT <: HList](implicit
```
vkostyukov: I'm thinking there is a way to unify two cases: when a single content-type or a coproduct content-type is available. Maybe we can define `ToService` in terms of `ContentTypeNegotiation` (or `NegotiateToResponse`) instead of `ToResponse` and also derive an instance of `ContentTypeNegotiation` for a case when single content-type is available. This way we can posily get rid of code duplication. What do you think about this?

core/src/main/scala/io/finch/internal/Accept.scala
```diff
@@ -0,0 +1,21 @@
+package io.finch.internal
+
+import com.twitter.util.Try
+import javax.activation.MimeType
+
+case class Accept(`type`: String, subtype: String) { self =>
```
vkostyukov: What about `primary: String` and `sub: String` instead?

core/src/main/scala/io/finch/internal/Accept.scala
```diff
@@ -0,0 +1,21 @@
...
+
+  def matches(other: Accept): Boolean = {
```
vkostyukov: Should we define `Eq[Accept]` instance instead?

core/src/main/scala/io/finch/internal/Accept.scala
```diff
@@ -0,0 +1,21 @@
...
+    (self.`type` == other.`type` || self.`type` == "*") &&
+      (self.subtype == other.subtype || self.subtype == "*")
+  }
+
+}
+
+object Accept {
+
+  def apply(s: String): Option[Accept] = Try {
+    val mt = new MimeType(s)
```
vkostyukov: Is there a way to do that w/o relying on Java APIs?

vkostyukov: Also just using try-catch would probably be more efficient as it allows us to eliminate the `Try` allocation.

core/src/main/scala/io/finch/ContentTypeNegotiation.scala
ImLiar: It's the way to fix implicit divergence. Compiler seems to fail to resolve recursion of `mkCoproduct` because `A` parameter is stable from resolution to resolution.

core/src/main/scala/io/finch/ToService.scala
ImLiar: Yep, I had this in mind, though didn't want to add an overhead penalty for those who use single type.

core/src/main/scala/io/finch/internal/Accept.scala
ImLiar: Yep, I like this idea

core/src/main/scala/io/finch/internal/Accept.scala
ImLiar: Didn't want to use a regex because it could lead to potential ReDoS.
I could parse it manually though

https://www.w3.org/Protocols/rfc2616/rfc2616-sec14.html

core/src/main/scala/io/finch/ToService.scala
vkostyukov: Why it has to be an overhead? I'd imagine `Bootstrap.serve[CT]` (single content-type) should still work.

ImLiar: overhead in terms of allocations per request

core/src/main/scala/io/finch/ContentTypeNegotiation.scala
vkostyukov: Interesting. How does this failure look like? An implicit not found, I'd guess. Yet I don't understand how is it different really. What if we change the order of params for `mkCoproduct`?

core/src/main/scala/io/finch/ToService.scala
vkostyukov: I'd say it's worth it. Also I doubt it's going to be on a per-request basis as we only need this materialize the `NegotiateToReponse` once for a given content-type/output type pair.

core/src/main/scala/io/finch/internal/Accept.scala
vkostyukov: I see. Let use this Java thing for now and evolve it later if we want to.

core/src/main/scala/io/finch/ContentTypeNegotiation.scala
ImLiar: Here is a minimal working example:

```scala
import shapeless.{:+:, CNil, Coproduct}

trait DivergenceTest[A] {

  type B

  def apply(): Unit

}

type Aux[a, b] = DivergenceTest[a] { type B = b }

implicit def mkCNil[A]: Aux[A, CNil] = new DivergenceTest[A] {
  type B = CNil

  def apply(): Unit = ()
}

implicit def mkCoproduct[A, H, T <: Coproduct]
  (implicit tail: Aux[A, T]): Aux[A, H :+: T] = {
  new DivergenceTest[A] {

    type B = H :+: T

    def apply(): Unit = tail()
  }
}

type C1 = Int :+: CNil
type C2 = String :+: Int :+: CNil
implicitly[Aux[String, C1]] //works
implicitly[Aux[String, C2]] //exception:  diverging implicit expansion for type Aux[String,C1] starting with method mkCNil
```

core/src/main/scala/io/finch/ToService.scala
ImLiar: Yeah, but it at least doubles amount of `ToResponse` instances generated per request

vkostyukov: My guess is it's just a drop in the ocean. We can always run the `ToServiceBenchmark` to get some concrete numbers.

core/src/main/scala/io/finch/ContentTypeNegotiation.scala
ImLiar: I've replaced Aux pattern with two type parameters. It seems that there is no need for Aux anyway.

ImLiar:
@vkostyukov I think it should be fine now

core/src/main/scala/io/finch/ToService.scala
```diff
@@ -99,15 +75,41 @@ object ToService {
...
+            includeDateHeader,
+            includeServerHeader
+          )).run
 
         case _ => tsT(endpoints.tail, includeDateHeader, includeServerHeader)(req)
       }
     }
   }
+
+  protected val respond400OnErrors: PartialFunction[Throwable, Output[Nothing]] = {
```
vkostyukov: Can we move this and `conformHttp` back to the top now? Also make them private?

core/src/main/scala/io/finch/internal/Accept.scala
```diff
@@ -0,0 +1,22 @@
+package io.finch.internal
+
+import cats.Eq
+import javax.activation.MimeType
+import scala.util.control.NonFatal
+
+case class Accept(primary: String, subtype: String)
```
vkostyukov: As `NegotiateToResponse` lives under a root namespace, maybe we should move this too (one level up)?

core/src/main/scala/io/finch/NegotiateToResponse.scala
```diff
@@ -0,0 +1,54 @@
...
+object NegotiateToResponse extends MediumPriorityNegotiation {
+
+  implicit def mkCoproduct[A, CTH <: String, CTT <: Coproduct](implicit
+    h: ToResponse.Aux[A, CTH],
+    t: NegotiateToResponse[A, CTT],
+    w: Witness.Aux[CTH]
+  ): NegotiateToResponse[A, CTH :+: CTT] = instance { accept =>
+    Accept(w.value) match {
+      case Some(ct) if accept.exists(_ === ct) =>
+        ToResponse.instance[A, CTH :+: CTT]((a, cs) => h(a, cs))
```
vkostyukov: Maybe a silly question, but can't we just do something like this instead?

```scala
h.asInstanceOf[ToResponse.Aux[A, CTH :+: CTT]]
```

core/src/main/scala/io/finch/NegotiateToResponse.scala
```diff
@@ -0,0 +1,54 @@
...
+  def instance[A, CT](f: Seq[Accept] => ToResponse.Aux[A, CT]): NegotiateToResponse[A, CT] = {
+    new NegotiateToResponse[A, CT] {
+      def apply(accept: Seq[Accept]): ToResponse.Aux[A, CT] = f(accept)
+    }
+  }
+
+  implicit def mkLast[A, CTH <: String](implicit
+    tr: ToResponse.Aux[A, CTH]
+  ): NegotiateToResponse[A, CTH :+: CNil] = instance { _ =>
+    ToResponse.instance[A, CTH :+: CNil]((a, cs) => tr(a, cs))
```
vkostyukov: Similar question about just calling `asInstanceOf` instead.

core/src/main/scala/io/finch/NegotiateToResponse.scala
```diff
@@ -0,0 +1,54 @@
...
+      case Some(ct) if accept.exists(_ === ct) =>
+        ToResponse.instance[A, CTH :+: CTT]((a, cs) => h(a, cs))
+      case _ =>
+        ToResponse.instance[A, CTH :+: CTT]((a, cs) => t(accept)(a, cs))
+    }
+  }
+
+}
+
+trait MediumPriorityNegotiation {
```
vkostyukov: `LowPriority`? Actually, can't we just move them back to the object now?

core/src/main/scala/io/finch/ToService.scala
```diff
@@ -99,15 +75,41 @@ object ToService {
...
 
       def apply(req: Request): Future[Response] = underlying(Input.fromRequest(req)) match {
-        case EndpointResult.Matched(rem, out) if rem.route.isEmpty => out.map(oa => conformHttp(
-          oa.toResponse(trA, trE),
-          req.version,
-          includeDateHeader,
-          includeServerHeader
-        )).run
+        case EndpointResult.Matched(rem, out) if rem.route.isEmpty =>
+          val accept = req.accept.flatMap(a => Accept(a))
```
vkostyukov: I'm wondering if we should keep it under a flag for now? Something like `Bootstrap.configure(negotiateContentType=true)` (disabled by default) so we can avoid header's query for applications where it might affect performance (think of TechEmpower benchmark that might get some performance penalty for `req.accept`).

core/src/main/scala/io/finch/NegotiateToResponse.scala
```diff
@@ -0,0 +1,48 @@
...
+    Accept(w.value) match {
+      case Some(ct) if accept.exists(_ === ct) =>
+        h.asInstanceOf[ToResponse.Aux[A, CTH :+: CTT]]
+      case _ =>
+        t(accept).asInstanceOf[ToResponse.Aux[A, CTH :+: CTT]]
+    }
+  }
+
+  implicit def mkLast[A, CTH <: String](implicit
+                                        tr: ToResponse.Aux[A, CTH]
```
vkostyukov: Format this, please.

core/src/main/scala/io/finch/NegotiateToResponse.scala
```diff
@@ -0,0 +1,48 @@
...
+                                       ): NegotiateToResponse[A, CTH :+: CNil] = instance { _ =>
+    tr.asInstanceOf[ToResponse.Aux[A, CTH :+: CNil]]
+  }
+
+  implicit def mkSingle[A, CT <: String](implicit
+                                         tr: ToResponse.Aux[A, CT]
```
vkostyukov: Format this, please

core/src/main/scala/io/finch/ToService.scala
```diff
@@ -74,40 +75,45 @@ object ToService {
     rep
   }
 
-  implicit val hnilTS: ToService[HNil, HNil] = new ToService[HNil, HNil] {
+  implicit def hnilTS: ToService[HNil, HNil] = new ToService[HNil, HNil] {
```
vkostyukov: Why do we have to change this to `def`?

ImLiar:
Added benchmarks to this branch.

Output:
```
Master:

[info] Benchmark                                               Mode  Cnt       Score       Error   Units
[info] BootstrapBenchmark.foos                                thrpt   10   18700.160 ±  7381.969   ops/s
[info] BootstrapBenchmark.foos:·gc.alloc.rate                 thrpt   10    1334.944 ±   525.299  MB/sec
[info] BootstrapBenchmark.foos:·gc.alloc.rate.norm            thrpt   10  112361.261 ±    17.848    B/op
[info] BootstrapBenchmark.foos:·gc.churn.Par_Eden_Space       thrpt   10    1326.420 ±   610.823  MB/sec
[info] BootstrapBenchmark.foos:·gc.churn.Par_Eden_Space.norm  thrpt   10  108262.771 ± 25008.267    B/op
[info] BootstrapBenchmark.foos:·gc.count                      thrpt   10      61.000              counts

This branch:

[info] Benchmark                                                      Mode  Cnt       Score       Error   Units
[info] BootstrapBenchmark.foos                                       thrpt   10   17905.063 ±  6828.829   ops/s
[info] BootstrapBenchmark.foos:·gc.alloc.rate                        thrpt   10    1259.395 ±   478.560  MB/sec
[info] BootstrapBenchmark.foos:·gc.alloc.rate.norm                   thrpt   10  110856.874 ±  2436.336    B/op
[info] BootstrapBenchmark.foos:·gc.churn.Par_Eden_Space              thrpt   10    1260.865 ±   435.057  MB/sec
[info] BootstrapBenchmark.foos:·gc.churn.Par_Eden_Space.norm         thrpt   10  113350.379 ± 22603.607    B/op
[info] BootstrapBenchmark.foos:·gc.count                             thrpt   10      58.000              counts
[info] BootstrapBenchmark.foos:·gc.time                              thrpt   10     144.000                  ms
[info] BootstrapBenchmark.negotiation                                thrpt   10   26370.146 ±  3190.968   ops/s
[info] BootstrapBenchmark.negotiation:·gc.alloc.rate                 thrpt   10    1470.497 ±   178.315  MB/sec
[info] BootstrapBenchmark.negotiation:·gc.alloc.rate.norm            thrpt   10   87734.576 ±   173.153    B/op
[info] BootstrapBenchmark.negotiation:·gc.churn.Par_Eden_Space       thrpt   10    1457.398 ±   269.884  MB/sec
[info] BootstrapBenchmark.negotiation:·gc.churn.Par_Eden_Space.norm  thrpt   10   86816.991 ±  9609.368    B/op
[info] BootstrapBenchmark.negotiation:·gc.count                      thrpt   10      67.000              counts
[info] BootstrapBenchmark.negotiation:·gc.time                       thrpt   10     154.000                  ms
```

benchmarks/src/main/scala/io/finch/benchmarks.scala
```diff
@@ -181,6 +182,39 @@ class ToServiceBenchmark extends FinchBenchmark {
   def ints: Response = Await.result(intService(Request()))
 }
 
+@State(Scope.Benchmark)
+@BenchmarkMode(Array(Mode.Throughput))
+@OutputTimeUnit(TimeUnit.SECONDS)
+class BootstrapBenchmark extends FinchBenchmark {
```
vkostyukov: I'd probably vote to unify our ToService and Boostrap benchmarks, but I'm happy to do that in a separate PR.

vkostyukov:
Just want to make sure we discuss this: looks like we're ignoring [the quantity](https://www.w3.org/Protocols/rfc2616/rfc2616-sec14.html#sec14.1) right now, which may be ok for the first version. @ImLiar have you thought about adding a support for them in the future?

ImLiar:
@vkostyukov 

Yes, I know about a quality factor, but decided to keep it simple for a first iteration.

vkostyukov:
@ImLiar I think this change is fine as it is (even w/o qualities). I was curious if you thought about future iterations that may/may not include supporting qualities.

ImLiar:
@rpless Exactly, I had the same thoughts. Some builder or configuration DTO would help I guess

vkostyukov:
I agree there is been a tendency of increasing the number of boolean flags within `Bootstrap/ToService` entities, yet I'm not too worried about those. It's pretty much a finite space of things we want to be able to configure and as long as we stay within a reasonable limit (I doubt we ever go to a two-digit number of arguments) we should be fine. It's hard to come with the most efficient and simple way of configuring something rather than a boolean flag. I'd guess having a separate object (i.e., `case class BootstrapParams`) would just increase the allocations while not necessarily improving reasoning/maintainability.

ImLiar:
I'm currently busy with implementation of quality factor and just found interesting thing in RFC:
> If an Accept header field is present, and if the server cannot send a response which is acceptable according to the combined Accept field value, then the server SHOULD send a 406 (not acceptable) response.

I think it's reasonable to return `ToResponse` with 406 inside in that case. @vkostyukov @rpless what are your thoughts on this?

rpless:
@ImLiar I think it makes sense to return `406` if we can't serve the content types specified in the `Accept` header.

spockz:
It does make sense. However, I’ve also encountered servers who would return the best/only content regardless of the Accept content type.

vkostyukov:
I think returning 406 if the client-server can't agree on a content-type is a reasonable thing to do. Yet I'm not sure we have to strike for this right now.

It's probably a good time to merge this in. @ImLiar are you happy with the stat of this PR or there is anything else you wanted to do before I merge this in (let me look one more time though)?

core/src/main/scala/io/finch/NegotiateToResponse.scala
```diff
@@ -0,0 +1,48 @@
+package io.finch
+
+import cats.syntax.eq._
+import shapeless._
+
+/**
+  * `NegotiateToResponse` enables content negotiation with client.
+  */
+trait NegotiateToResponse[A, ContentType] {
```
vkostyukov: On crazy thought. What do people think about building the negotiation into `ToResponse` itself (and avoiding introducing a new abstraction)? I'd imagine, it /would/ be possible to extend `ToResponse` with an extra param `Accept` and make it behaving as `NegotiateToResponse`:


```scala
trait ToResponse[A] {
  type ContentType

  def apply(a: A, cs: Charset, accept: Accept): Response
}
```

So we can derive instances as `ToRespone.Aux[Foo :+: Bar :+: CNil, Application.Json :+: Text.Plain :+: CNil]`, meaning that both `Foo` and `Bar` can be served as JSON and text.

core/src/main/scala/io/finch/Accept.scala
```diff
@@ -0,0 +1,22 @@
...
+
+import cats.Eq
+import javax.activation.MimeType
+import scala.util.control.NonFatal
+
+case class Accept(primary: String, subtype: String)
+
+object Accept {
+
+  def apply(s: String): Option[Accept] = try {
```
vkostyukov: Can we rename it to `fromString` (otherwise it looks like a case class constructor, which is a total function).

core/src/main/scala/io/finch/NegotiateToResponse.scala
ImLiar: I like it in general, but it would mean that anyone who wants to use custom `ToResponse` (i.e. for error rendering) should implement content negotiation on their own.

vkostyukov: Why would people implement their own `ToResponse`? I'd assume everything can be done at the `Encode` level. `ToResponse` was mostly only needed to accommodate streaming types (that don't necessarily go into `Buf`s).

That said, we might want to add a new `ToResponse` instance if we want to support some new streaming types (i.e., fs2) yet I'm not sure why we'd want to alter the negotiation logic?

ImLiar: This is one of the examples I can come up with: https://tools.ietf.org/html/rfc7807
Briefly, to return different `Content-Type` header when there is an error.

`ToResponse` gives more control over the actual response since encoder takes care only of the content/body.

vkostyukov: I don't think we support that though? `ToService` resolves two `ToResponse` instances of the same content-type (one for success, one for failure).

ImLiar: Technically, one can define a function `implicit def tre[CT]: ToResponse[Exception, CT]` and return correct problem json, though this one is resolved for any `CT`

ImLiar:
@vkostyukov I think it's good to go know

core/src/main/scala/io/finch/Output.scala
```diff
@@ -195,13 +195,13 @@ object Output {
     /**
      * Converts this [[Output]] to the HTTP response of the given `version`.
      */
-    def toResponse[CT <: String](implicit
+    def toResponse[CT](accept: Seq[Accept] = Seq.empty)(implicit
```
vkostyukov: Ugh. I totally forgot we need to propagate `accept` here. It's been on review for too long (and I appreciate your patience!) so let me just grab this branch, fix the problem I found with `Bootstrap` and revert changes to `ToResponse`.

vkostyukov: I rebased this branch and squashed everything. It's here: #905 (will merge that when it's green).

ImLiar:
Closing this one in favor of https://github.com/finagle/finch/pull/905

PR #883, opened 2017-12-27
Introduce MethodNotAllowed

sergeykolbasov:
This one should fix https://github.com/finagle/finch/issues/784

The logic is following:  
- If Endpoint was matched by path, but method is different, enable a boolean flag `methodNotAllowed` and continue resolution of endpoints.
- At `HNil` check this boolean flag, if it's true - return 405. 404 otherwise

External API again stays the same

### DIFF
--- a/core/src/main/scala/io/finch/Endpoint.scala
+++ b/core/src/main/scala/io/finch/Endpoint.scala
@@ -209,7 +209,7 @@ trait Endpoint[A] { self =>
     final def apply(input: Input): Endpoint.Result[B] = self(input) match {
       case a @ EndpointResult.Matched(_, _) => other(input) match {
         case b @ EndpointResult.Matched(_, _) =>
-          if (a.rem.route.length <= b.rem.route.length) a else b
+          if (a.rem.route.length <= b.rem.route.length && a.rem.method == input.method) a else b
         case _ => a
       }
       case _ => other(input)
--- a/core/src/main/scala/io/finch/Input.scala
+++ b/core/src/main/scala/io/finch/Input.scala
@@ -15,12 +15,17 @@ import shapeless.Witness
  * - Finagle's [[Request]] needed for evaluating (e.g., `body`, `param`)
  * - Finch's route (represented as `Seq[String]`) needed for matching (e.g., `path`)
  */
-final case class Input(request: Request, route: Seq[String]) {
+final case class Input(request: Request, route: Seq[String], method: Method) {
 
   /**
    * Returns the new `Input` wrapping a given `route`.
    */
-  def withRoute(route: Seq[String]): Input = Input(request, route)
+  def withRoute(route: Seq[String]): Input = Input(request, route, method)
+
+  /**
+    * Returns the new `Input` wrapping a given `method`.
+    */
+  def withMethod(method: Method): Input = Input(request, route, method)
 
   /**
    * Returns the new `Input` wrapping a given payload. This requires the content-type as a first
@@ -42,7 +47,7 @@ final case class Input(request: Request, route: Seq[String]) {
     val copied = Input.copyRequest(request)
     headers.foreach { case (k, v) => copied.headerMap.set(k, v) }
 
-    Input(copied, route)
+    Input(copied, route, method)
   }
 
   /**
@@ -110,7 +115,7 @@ object Input {
       copied.contentLength = content.length.toLong
       charset.foreach(cs => copied.charset = cs.displayName().toLowerCase)
 
-      Input(copied, i.route)
+      Input(copied, i.route, i.method)
     }
   }
 
@@ -119,7 +124,7 @@ object Input {
   /**
    * Creates an [[Input]] from a given [[Request]].
    */
-  def fromRequest(req: Request): Input = Input(req, req.path.split("/").toList.drop(1))
+  def fromRequest(req: Request): Input = Input(req, req.path.split("/").toList.drop(1), req.method)
 
   /**
    * Creates a `GET` input with a given query string (represented as `params`).
--- a/core/src/main/scala/io/finch/ToService.scala
+++ b/core/src/main/scala/io/finch/ToService.scala
@@ -37,7 +37,24 @@ trait ToService[ES <: HList, CTS <: HList] {
     includeDateHeader: Boolean,
     includeServerHeader: Boolean,
     negotiateContentType: Boolean,
-    methodNotAllowed: Boolean
+    enableMethodNotAllowed: Boolean
+  ): Service[Request, Response] =
+    run(
+      endpoints,
+      includeDateHeader,
+      includeServerHeader,
+      enableMethodNotAllowed,
+      negotiateContentType,
+      matchedWithoutMethod = false
+    )
+
+  protected def run(
+    endpoints: ES,
+    includeDateHeader: Boolean,
+    includeServerHeader: Boolean,
+    negotiateContentType: Boolean,
+    enableMethodNotAllowed: Boolean,
+    matchedWithoutMethod: Boolean
   ): Service[Request, Response]
 }
 
@@ -77,15 +94,16 @@ object ToService {
   }
 
   implicit val hnilTS: ToService[HNil, HNil] = new ToService[HNil, HNil] {
-    def apply(
+    def run(
         endpoints: HNil,
         includeDateHeader: Boolean,
         includeServerHeader: Boolean,
         negotiateContentType: Boolean,
+        enableMethodNotAllowed: Boolean,
         methodNotAllowed: Boolean): Service[Request, Response] = new Service[Request, Response] {
 
       def apply(req: Request): Future[Response] = {
-        val status = if (methodNotAllowed) Status.MethodNotAllowed else Status.NotFound
+        val status = if (methodNotAllowed && enableMethodNotAllowed) Status.MethodNotAllowed else Status.NotFound
         Future.value(conformHttp(Response(status), req.version, includeDateHeader, includeServerHeader))
       }
     }
@@ -96,12 +114,13 @@ object ToService {
     ntrE: ToResponse.Negotiable[Exception, CTH],
     tsT: ToService[ET, CTT]
   ): ToService[Endpoint[A] :: ET, CTH :: CTT] = new ToService[Endpoint[A] :: ET, CTH :: CTT] {
-    def apply(
+    def run(
         endpoints: Endpoint[A] :: ET,
         includeDateHeader: Boolean,
         includeServerHeader: Boolean,
         negotiateContentType: Boolean,
-        methodNotAllowed: Boolean): Service[Request, Response] = new Service[Request, Response] {
+        enableMethodNotAllowed: Boolean,
+        matchedWithoutMethod: Boolean): Service[Request, Response] = new Service[Request, Response] {
 
       private[this] val underlying = endpoints.head.handle(respond400OnErrors)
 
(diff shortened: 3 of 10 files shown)


vkostyukov:
Thanks @ImLiar! This is pretty much what I had in mind at first. Although, I decided to give this idea a little more thoughts as I think the following will now be broken:

```scala
Bootstrap
  .serve[Text.Plain](get("foo") { Ok("foo")} )
  .serve[Text.Plain](post("foo") { Ok("foo")} )
  .toService
```

I'd assume it will return 405 on `POST /foo` even though there is a route available (please, correct me if missed something).

core/src/main/scala/io/finch/ToService.scala
```diff
@@ -94,19 +97,25 @@ object ToService {
...
+        case EndpointResult.Matched(rem, out) if rem.route.isEmpty =>
+          if (rem.method == req.method) {
+            out.map(oa => conformHttp(
+              oa.toResponse(trA, trE),
+              req.version,
+              includeDateHeader,
+              includeServerHeader
+            )).run
+          } else {
+            tsT(endpoints.tail, includeDateHeader, includeServerHeader, methodNotAllowed = true)(req)
```
vkostyukov: Oh I see it now - this what makes it fall through to the next endpoint.

ImLiar: Yep. I've updated a spec to reflect this case

vkostyukov: I like the idea of notifying the next endpoint that the match was found, but I think we might want to use a different flag for this (while keeping the `methodNotAllowed` unchanged). What do you think?

ImLiar: Could you elaborate on this one, please? I don't get it

vkostyukov: Oh sorry. I think it's just me not understanding this change completely. At first, I thought `methodNotAllowed` is something that goes into `Bootstrap` so users can toggle it on/off (which I think we should do anyway). Then I figured you override it to signal the next endpoint that the match was found so I suggested using a different flag for this (turns out you already do that!). How about we wire in `Bootstrap.configured(enableMethodNotAllowed)`?

Also I'd probably vote to change `ToService.apply(methodNotAllowed)` to something like `matchIsFound` (or `endpointIsMatched`). What do you think?

core/src/main/scala/io/finch/ToService.scala
```diff
@@ -94,19 +97,25 @@ object ToService {
...
       def apply(req: Request): Future[Response] = underlying(Input.fromRequest(req)) match {
-        case EndpointResult.Matched(rem, out) if rem.route.isEmpty => out.map(oa => conformHttp(
-          oa.toResponse(trA, trE),
-          req.version,
-          includeDateHeader,
-          includeServerHeader
-        )).run
-
-        case _ => tsT(endpoints.tail, includeDateHeader, includeServerHeader)(req)
+        case EndpointResult.Matched(rem, out) if rem.route.isEmpty =>
```
vkostyukov: How about we move that method check under the if-guard? Something along the lines:

```scala
case EndpointResult.Matched(rem, out) if (rem.route.isEmpty && rem.method == req.method) =>
  // full match
case EndpointResult.Matched(rem, out) if (rem.route.isEmpty) =>
  // method wasn't matched
case _ => // fall through
```

core/src/main/scala/io/finch/ToService.scala
ImLiar: Ah, okay, got it.
But the match is actually not found because a method is incorrect.

Regarding configuration. I guess flag from the user should enable/disable the option instead of corrupting a behavior because it will turn all 404s into 405s if one will set it as `true` by default.

vkostyukov: Right. So ToService should have both: one from a user (whether or not to support 405), one from the previous endpoint (whether or not it was matched).

core/src/main/scala/io/finch/ToService.scala
ImLiar: Then it will have to unapply it twice, so I think it's not that efficient compared to current solution.

core/src/main/scala/io/finch/ToService.scala
vkostyukov: How about `matchedWithoutMethod` flag then (to indicate that very special case)?

core/src/main/scala/io/finch/ToService.scala
vkostyukov: It will be fine - unapply is supper for case classes (it just a couple of `isInstanceOf`/ `asInstanceOf` calls).

ImLiar:
I'm slightly confused by changing a behavior of verbs endpoints, though. It always matches now.
Is it ok?

vkostyukov:
Yeah, I know. I wish there was another way to do that. I guess we can consider adding a new `EndpointResult` state to capture the "matched w/o method" state, but it seems like a way bigger change to me.

I think I'm generally fine with `Input` also capturing the `method` (it feels quite natural). I like how @ImLiar moved the new boolean flag under the run method so it doesn't pollute the `apply` API.

core/src/main/scala/io/finch/ToService.scala
```diff
@@ -35,7 +35,17 @@ trait ToService[ES <: HList, CTS <: HList] {
   def apply(
     endpoints: ES,
     includeDateHeader: Boolean,
-    includeServerHeader: Boolean
+    includeServerHeader: Boolean,
+    enableMethodNotAllowed: Boolean = false
+  ): Service[Request, Response] =
+    run(endpoints, includeDateHeader, includeServerHeader, enableMethodNotAllowed, methodNotAllowed = false)
+
+  protected def run(
```
vkostyukov: What do you think about `applyWithContext` instead? Assuming `methodNotAllowed` is somewhat part of the context that needs to be propagated across the chain of `ToService` instances.

core/src/main/scala/io/finch/ToService.scala
```diff
@@ -35,7 +35,17 @@ trait ToService[ES <: HList, CTS <: HList] {
...
+    endpoints: ES,
+    includeDateHeader: Boolean,
+    includeServerHeader: Boolean,
+    enableMethodNotAllowed: Boolean,
+    methodNotAllowed: Boolean
```
vkostyukov: I'd probably still vote to rename it to `matchedWithoutMethod` (or something like this).

core/src/main/scala/io/finch/ToService.scala
vkostyukov: Oh, also. How do you feel about moving context-params into a separate param list (to visually distinguish them)?

```scala
def applyWithContext(
  includeDateHeader: Boolean,
  includeServerHeader: Boolean,
  enableMethodNotAllowed: Boolean)(matchedWithoutMethod: Boolean)
```

vkostyukov:
Thanks again @ImLiar! I'm excited to see this shipped as part of 0.17!

core/src/main/scala/io/finch/Input.scala
```diff
@@ -15,12 +15,17 @@ import shapeless.Witness
  * - Finagle's [[Request]] needed for evaluating (e.g., `body`, `param`)
  * - Finch's route (represented as `Seq[String]`) needed for matching (e.g., `path`)
  */
-final case class Input(request: Request, route: Seq[String]) {
+final case class Input(request: Request, route: Seq[String], method: Method) {
```
rpless: Can we update the doc for this class to include `method`? Maybe something like:
```
 * An input for [[Endpoint]] that glues three individual pieces together:
 *
 * - Finagle's [[Request]] needed for evaluating (e.g., `body`, `param`)
 * - Finch's route (represented as `Seq[String]`) needed for matching (e.g., `path`)
 * - Finagle's [[Method]] needed for matching the the request's HTTP Verb
```

ImLiar: Sure, thanks for pointing out

vkostyukov:
Going back to this. I wonder if following would still work:

```scala
val ab = (get("foo") { Ok("a") } :+: post("foo") { Ok("b") }).toService
```

I'd assume not, given we always match in verb-endpoints. So the first will always be picked up even though we asked for `POST /foo`. Am I missing something?

ImLiar:
It shouldn't pick up `get` endpoint because there is an if-guard in pattern matching:
https://github.com/ImLiar/finch/blob/5ec9f423a5dad6ac132e0a869d6975e2ac7dfa9c/core/src/main/scala/io/finch/ToService.scala#L116

vkostyukov:
Yeah, but it happens much earlier in the `Endpoit.:+:` implementation (assuming we pass a coproduct endpoint into `Bootstrap.serve`). I think we can run a quick experiment to confirm this.

ImLiar:
Yes, you're correct on this one.

vkostyukov:
Ugh. Looks like it's a bit more involved than we thought it is.

ImLiar:
@vkostyukov 
I think 4200440 fixes this case

ImLiar:
travis fails with no reason, some internal error

rpless:
@ImLiar I restarted the Travis build and it seems to be running ok now.

ImLiar:
Thanks @rpless

ImLiar:
ping @vkostyukov 

This PR also requires some attention 😄

vkostyukov:
Hey @ImLiar! Let's make this part of 0.18 release. Do you mind rebasing this so I can take another look?

ImLiar:
@vkostyukov done

core/src/main/scala/io/finch/Endpoint.scala
```diff
@@ -209,7 +209,7 @@ trait Endpoint[A] { self =>
     final def apply(input: Input): Endpoint.Result[B] = self(input) match {
       case a @ EndpointResult.Matched(_, _) => other(input) match {
         case b @ EndpointResult.Matched(_, _) =>
-          if (a.rem.route.length <= b.rem.route.length) a else b
+          if (a.rem.route.length <= b.rem.route.length && a.rem.method == input.method) a else b
```
vkostyukov: So I've been thinking about this a lot and I'm not sure I like this particular piece. I tried experimenting with something in a local branch and it seems we can do something better. Here is an idea: encode method-not-allowed in `EndpointResult`. The scope of the change doesn't seem to break a lot of things:

The ADT would now have three cases to model HTTP routing result. It comes down to three things we're interested: a) route found b) route found but method is not allowed c) route not found

```scala
object EndpointResult {

  final case class Matched[A](rem: Input, out: Rerunnable[Output[A]]) extends EndpointResult[A] {
    def isMatched: Boolean = true
  }

  case object NotMatched extends EndpointResult[Nothing] { // previosuly Skipped
    def isMatched: Boolean = false
  }

  case object MethodNotMatched extends EndpointResult[Nothing] {
    def isMatched: Boolean = false
  }
}
```

The verb-matching combinators are now behaving correctly (returning an appropriate result):

```scala
      if (input.request.method == m) e(input)
      else e(input) match {
        case EndpointResult.Matched(_, _) => EndpointResult.MethodNotMatched
        case skipped => skipped
      }
```

Given the new endpoint result, `Endpoint.coproduct` can now be implemented more carefully. We don't have to check the method, we just need to return the most appropriate result. Here is the sketch:

```scala
  final def coproduct[B >: A](other: Endpoint[B]): Endpoint[B] = new Endpoint[B] {
    final def apply(input: Input): Endpoint.Result[B] = self(input) match {
      case a @ EndpointResult.Matched(_, _) => other(input) match {
        case b @ EndpointResult.Matched(_, _) =>
          if (a.rem.route.length <= b.rem.route.length) a else b
        case _ => a
      }
      case a => other(input) match {
        case EndpointResult.NotMatched => a
        case b => b
      }
    }
}
```

The rest is pretty much the same code as in this PR (think of changes to `ToService`).

@sergeykolbasov What do you think about this? I like how it's not breaking our `foo(Input.get("/"))`behavior.

vkostyukov: I experimented with it here: https://github.com/finagle/finch/pull/927

vkostyukov:
We're considering an alternative solution in #927.

sergeykolbasov:
closed in favor of #927

Issue #587, opened 2016-04-09
Circe error accumulation


vkostyukov:
As Travis described [on Gitter ](https://gitter.im/finagle/finch?at=5709750fddb5a2cf3bbae825).

wennergr:
More then happy to help out here, but I would need some help ;)

Some initial thoughts and questions:

## Issue One, decode function
We would need a slightly modified version of the circe [Parser.decode](https://github.com/travisbrown/circe/blob/master/core/shared/src/main/scala/io/circe/Parser.scala)

Best case (for me) would be a decode function with the following signature:
```scala
def decode[A](input: String)(implicit d: Decoder [A]): Xor[Nel[Error], A]
```

This will however not work because `Nel` is invarient and this function code return both `ParsingFailure` and `DecodingFailure` (and these relates to error Error through sub-classing).

We could instead use **list** instead of **nel**:
```scala
def decode[A](input: String)(implicit d: Decoder [A]): Xor[List[Error], A]
```

## Issue Two, RequestErrors
The Endpoint logic `StringEndpointOps.as` will create a new Error.NotParsed with our RequestErrors as the cause. E.g we will end up with our multiple errors wrapped inside the NotParser error (cause property). (Not what we want)

There are multiple ways to fix this but they are all non additive changes. Any thought on this would be appreciated.

travisbrown:
@wennergr I've just opened [a quick PR](https://github.com/travisbrown/circe/pull/260) that adds a `decodeAccumulating` method to `Parser` in circe (we'll eventually want this kind of thing for multiple web service frameworks, so we might as well do it there).

Just because `NonEmptyList` is invariant, that doesn't mean upcasting the element type isn't safe—you just have to do it "manually".

bergmark:
This would be super nice to have!

Did anyone start working on this? Is what @wennergr suggested the right way to go? I'd be interested in helping out if we can decide how to do this.

PR #884, opened 2017-12-29
Add accumulating option to io.finch.circe

ImLiar:
This one addresses https://github.com/finagle/finch/issues/587
To enable accumulating decoding one should use `import io.finch.circe.accumulating._` instead of `import io.finch.circe._`.

🎄🎁

### DIFF
--- a/circe/src/main/scala/io/finch/circe/AccumulatingDecoders.scala
+++ b/circe/src/main/scala/io/finch/circe/AccumulatingDecoders.scala
@@ -1,0 +1,54 @@
+package io.finch.circe
+
+import java.nio.charset.StandardCharsets
+
+import cats.MonadError
+import cats.data.Validated
+import com.twitter.util.{Decoder => _, _}
+import io.catbird.util._
+import io.circe._
+import io.circe.iteratee._
+import io.circe.jawn.{decodeAccumulating, decodeByteBufferAccumulating}
+import io.finch.{Application, Decode}
+import io.finch.internal.HttpContent
+import io.finch.iteratee.Enumerate
+import io.iteratee.{Enumeratee, Enumerator}
+
+trait AccumulatingDecoders {
+
+  /**
+    * Maps a Circe's [[Decoder]] to Finch's [[Decode]].
+    */
+  implicit def decodeCirce[A: Decoder]: Decode.Json[A] = Decode.json { (b, cs) =>
+    val attemptJson = cs match {
+      case StandardCharsets.UTF_8 =>
+        decodeByteBufferAccumulating[A](b.asByteBuffer)
+      case _ => decodeAccumulating[A](b.asString(cs))
+    }
+
+    attemptJson.fold[Try[A]](nel => Throw(Errors(nel)), Return.apply)
+  }
+
+  implicit def enumerateCirce[A : Decoder]: Enumerate.Json[A] = {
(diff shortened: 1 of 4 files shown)


circe/src/main/scala/io/finch/circe/Printers.scala
```diff
@@ -0,0 +1,42 @@
+package io.finch.circe
+
+import java.nio.charset.{Charset, StandardCharsets}
+
+import com.twitter.io.Buf
+import io.circe.{Json, Printer}
+
+trait Printers {
+
+  protected def print(json: Json, cs: Charset): Buf =
```
vkostyukov: I think this method is unused?

circe/src/main/scala/io/finch/circe/Printers.scala
```diff
@@ -0,0 +1,42 @@
+package io.finch.circe
+
+import java.nio.charset.{Charset, StandardCharsets}
+
+import com.twitter.io.Buf
+import io.circe.{Json, Printer}
+
+trait Printers {
```
vkostyukov: I don't really understand why we need to extract this? is that because we want to support `io.finch.circe.accumulating.dropNullValues._` (in the meaning of accumulative decoding and a special printing)?

Maybe it will actually be fine to just start with `io.circe.accumulating._` (as an embedded object into a circe package object) as a first iteration and then extend it from there (if people will start asking about overriding both decoding and encoding). Here is why I think it's not really a problem: most people define endpoints in one places and bootstram them somewhere else so they can use accumulating decoders to define and special-printing to bootstrap.

ImLiar: Ok, it makes sense

circe/src/main/scala/io/finch/circe/AccumulatingDecoders.scala
```diff
@@ -0,0 +1,54 @@
...
+          enum.map(_.asByteArray).through(byteStreamParser[Future])
+        case _ =>
+          enum.map(_.asString(cs)).through(stringStreamParser[Future])
+      }
+      parsed.through(decoderAccumulating[Future, A])
+    })
+  }
+
+  private def decoderAccumulating[F[_], A](implicit
+                                           F: MonadError[F, Throwable],
```
vkostyukov: Fix formatting, please.

circe/src/main/scala/io/finch/circe/package.scala
```diff
@@ -4,7 +4,7 @@ import com.twitter.io.Buf
 import io.circe.{Json, Printer}
 import java.nio.charset.{Charset, StandardCharsets}
 
-package object circe extends Encoders with Decoders {
+package object circe extends Encoders with Decoders { self =>
   protected def print(json: Json, cs: Charset): Buf =
     Buf.ByteBuffer.Owned(Printer.noSpaces.prettyByteBuffer(json, cs))
```
vkostyukov: What if we move this basic implementation under `Encoders` itself and then add overrides when necessary (in each of the configured instances). We wouldn't need the `self.print` redirect in this case.

vkostyukov:
I think this can be shipped in 0.16 (no need to wait for 0.17).

ImLiar:
@vkostyukov done

vkostyukov:
Thanks a lot @ImLiar! I love it!

PR #886, opened 2018-01-01
Updates Cats to 1.0

vkostyukov:
We still need:
 - [ ] iteratee-twitter
 - [ ] iteratee-circe
 - [ ] circe-jackson


core/src/main/scala/io/finch/ToService.scala
```diff
@@ -35,7 +35,7 @@ trait ToService[ES <: HList, CTS <: HList] {
   def apply(
     endpoints: ES,
     includeDateHeader: Boolean,
-    includeServerHeader: Boolean
```
ImLiar: scalastyle fails here

travisbrown:
Thanks for the PRs! All three releases are ready, with the two circe ones versioned 0.9.0 and iteratee-twitter as 0.18.0.

vkostyukov:
Thanks a lot, @travisbrown!

PR #888, opened 2018-01-02
Remove '-Ywarn-unused-import' from docs project's scalacOptions 

rpless:
Right now the code fences in the docs are filled with `warning: Unused import`. This due to how tut uses the REPL to evaluate code. For now we can filter out `-Ywarn-unused-import` in the docs project. One downside is that we won't get these errors now for the docs, but I think we'll just have to live with it. We'll also need to republish the docs after this.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -150,7 +150,7 @@ lazy val docSettings = allSettings ++ Seq(
     "-doc-root-content", (resourceDirectory.in(Compile).value / "rootdoc.txt").getAbsolutePath
   ),
   scalacOptions ~= {
-    _.filterNot(Set("-Yno-predef", "-Xlint"))
+    _.filterNot(Set("-Yno-predef", "-Xlint", "-Ywarn-unused-import"))
   },
   git.remoteRepo := "git@github.com:finagle/finch.git",
   unidocProjectFilter in (ScalaUnidoc, unidoc) := inAnyProject -- inProjects(benchmarks, jsonTest),


build.sbt
```diff
@@ -150,7 +150,7 @@ lazy val docSettings = allSettings ++ Seq(
     "-doc-root-content", (resourceDirectory.in(Compile).value / "rootdoc.txt").getAbsolutePath
   ),
   scalacOptions ~= {
-    _.filterNot(Set("-Yno-predef", "-Xlint"))
```
travisbrown: It would also work to do this just for `scalacOptions in Tut`, right?

rpless: I don't think so based on https://github.com/tpolecat/tut/issues/105. But I suppose we could `tutScalacOptions`. Its also odd cause it seems like tut should already be filtering this out based on https://github.com/tpolecat/tut/pull/108.

travisbrown: That issue is quite old, though. I actually needed to do this [this morning for circe](https://github.com/circe/circe/commit/6ada85a04ec84125a812c3650c527e46854337c7), and (surprisingly) the `in Tut` fixed the issue.

rpless: Good to know it works. We could do that instead. I only added there because that's how we'd added `-Xlint` and `-Yno-predef` previously. Happy to put it in the base settings though.

vkostyukov:
LGTM! Thanks @rpless!

PR #893, opened 2018-01-03
Fix link in Scaladoc

typeness:


vkostyukov:
Nice catch! Thanks @typeness!

PR #894, opened 2018-01-04
Welcome Sergey!

vkostyukov:
Adding @ImLiar as a Finch maintainer.


ImLiar:
@spockz Have you seen my card? It says "When this card is played, increase your test coverage automatically".

Seriously speaking, it's because of scalacheck and property testing. Sometimes it generate an input that hits some lines, sometimes it doesn't

Issue #895, opened 2018-01-05
import io.finch.syntax._ shadows _root_.scala._


mtomko:
Changes from https://github.com/finagle/finch/pull/876 to support explicit syntax selection result in potential and undesirable package conflicts between the packages `io.finch.syntax.scala` and `_root_.scala`.

With syntax selection, we explicitly import a syntax by importing either `io.finch.syntax._` or `io.finch.syntax.future._`. 

Since `io.finch.syntax` contains a package called `scala`, `import io.finch.syntax.scala` shadows `_root_.scala`, which causes potentially confusing compiler errors for anyone using something from `_root_.scala`:

```scala
[info] Compiling 20 Scala sources to /Users/mtomko/Documents/Idea/aguila/server/target/scala-2.12/classes...
[error] /Users/mtomko/Documents/Idea/aguila/server/src/main/scala/org/broadinstitute/gpp/aguila/server/ManagementEndpoints.scala:23: object collection is not a member of package io.finch.syntax.scala
[error] import scala.collection.JavaConverters._
[error]              ^
```

Opening issue as suggested by @vkostyukov on the gitter channel.

ImLiar:
Thanks for report @mtomko 

Yeah, it's a real bummer.
I think the most reasonable solution would be to rename `scala` package into something different.

I'll land a PR this weekend

mtomko:
made a PR: https://github.com/finagle/finch/pull/897

PR #897, opened 2018-01-05
Fix for https://github.com/finagle/finch/issues/895

mtomko:
rename  `io.finch.syntax.scala` to `io.finch.syntax.scalafutures` (per @vkostyukov)


core/src/main/scala/io/finch/syntax/scalafutures.scala
```diff
@@ -1,12 +1,17 @@
...
+package io.finch.syntax
 
-import scala.concurrent.{Future => ScalaFuture}
+import com.twitter.util.{Future, Promise}
+import scala.concurrent.{ExecutionContext, Future => ScalaFuture}
 import scala.util.{Failure, Success}
 
-import com.twitter.util.{Future, Promise}
-import io.finch.syntax.ToTwitterFuture
+object scalafutures extends EndpointMappers {
```
vkostyukov: As it's no longer a package object, can we rename it to `scalaFutures` instead?

mtomko: Will do.

vkostyukov:
This looks good to me. Thanks, @mtomko!

PR #898, opened 2018-01-06
Fix example for the match verbs section of the documentation

slabko:
Reading the documentation, I stumbled upon a strange example in the Match Verb section. To be precise, It didn't give any examples of matching verbs.

This PR reflects the way I understand what should be there, but I'm not sure if this is exactly what author wanted to show.

### DIFF
--- a/docs/src/main/tut/user-guide.md
+++ b/docs/src/main/tut/user-guide.md
@@ -177,13 +177,16 @@ For every HTTP verb, there is a function `Endpoint[A] => Endpoint[A]` that takes
 an arbitrary type and enriches it with an additional check/match of the HTTP method/verb.
 
 ```tut
-import io.finch._, com.twitter.finagle.http.{Request, Method}
+
+import io.finch._
 
 val e = path("foo")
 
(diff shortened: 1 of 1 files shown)


docs/src/main/tut/user-guide.md
```diff
@@ -177,13 +177,16 @@ For every HTTP verb, there is a function `Endpoint[A] => Endpoint[A]` that takes
...
 
 ```tut
-import io.finch._, com.twitter.finagle.http.{Request, Method}
+
+import io.finch._
 
 val e = path("foo")
 
-e(Input.get("/foo")).isMatched
+val a = get(e)
```
vkostyukov: How about we inline `e` here? `val a = get("foo")` would do it.

vkostyukov:
Hey @slabko! You're totally right, the documentation was misleading and your changes fix it. Thanks! Just a quick nit from my side before we can land it.

docs/src/main/tut/user-guide.md
vkostyukov: I will merge that in and fix that piece later.

PR #900, opened 2018-01-09
Revisit how Scala syntax is enabled

vkostyukov:
This PR pretty much reverts the Scala syntax support to its original version, when only `ToTwitterFuture[ScalaFuture]` is being imported from a `io.finch.syntax.scalaFutures`.

I think I made a mistake trying to improve the ergonomics by enabling everything with just one import. It effectively makes it impossible to mix Scala and Twitter Futures in the scope/file. Let's fix it until it's too late.


mtomko:
This retains the key property from before (the naming). Is the idea that you'd locally import `scalaFutures._` wherever you had a Scala future? Does the cookbook example need to change?

here it is:

```scala
import io.finch._, io.finch.syntax.scalaFutures._
import scala.concurrent.Future
import scala.concurrent.ExecutionContext.Implicits.global

val e = get("foo") { Future.successful(Ok("bar")) }

e(Input.get("/foo")).awaitValueUnsafe()
```

I guess the suggestion is to now do:

```scala
import io.finch._, io.finch.syntax._
import scala.concurrent.Future
import scala.concurrent.ExecutionContext.Implicits.global

val e = get("foo") { 
  import scalaFutures._
  Future.successful(Ok("bar")) 
}

e(Input.get("/foo")).awaitValueUnsafe()
```

ImLiar:
@mtomko Nope, this import should be done on the top level because of how magnet pattern works and all implicits resolved

The change is that instead of having all syntax methods available with a single import of `scalaFutures._`, one should import syntax content AND `scalaFutures`.
Two imports instead of one.

ImLiar:
@vkostyukov LGTM

PR #904, opened 2018-01-20
Fix typo

robsonpeixoto:


rpless:
Thanks for catching this @robsonpeixoto!

PR #905, opened 2018-02-02
Content Type Negotiation

vkostyukov:
See #882.

### DIFF
--- a/core/src/main/scala/io/finch/Accept.scala
+++ b/core/src/main/scala/io/finch/Accept.scala
@@ -1,0 +1,22 @@
+package io.finch
+
+import cats.Eq
+import javax.activation.MimeType
+import scala.util.control.NonFatal
+
+case class Accept(primary: String, subtype: String)
+
+object Accept {
+
+  def apply(s: String): Option[Accept] = try {
+    val mt = new MimeType(s)
+    Some(Accept(mt.getPrimaryType, mt.getSubType))
+  } catch {
+    case NonFatal(_) => None
+  }
+
+  implicit val eq: Eq[Accept] = Eq.instance((a, b) => {
+    (a.primary == b.primary || a.primary == "*") &&
+      (a.subtype == b.subtype || a.subtype == "*")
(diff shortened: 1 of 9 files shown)


ImLiar:
👍

core/src/main/scala/io/finch/Accept.scala
```diff
@@ -0,0 +1,22 @@
...
+object Accept {
+
+  def apply(s: String): Option[Accept] = try {
+    val mt = new MimeType(s)
+    Some(Accept(mt.getPrimaryType, mt.getSubType))
+  } catch {
+    case NonFatal(_) => None
+  }
+
+  implicit val eq: Eq[Accept] = Eq.instance((a, b) => {
```
optician: * Symmetry and reflexivity don't hold. (*/*, text/html).
* If type is `*` then subtype should be `*` too. 
```
media-range    = ( "*/*"
                        | ( type "/" "*" )
                        | ( type "/" subtype )
                        ) *( ";" parameter )
```
from https://www.w3.org/Protocols/rfc2616/rfc2616-sec14.html

ImLiar: Hi

It's not supposed to be symmetric. We match client accept range against server range. While client could accept any wildcard media type, server can't produce wildcard one.

Though, I agree that using equality type class for it could be obscuring.

optician: Imho it worths efforts to introduce other type class in despite of only one possible instance. Violation of the law (especially such natural law) can become a time delayed bug.

vkostyukov: I think it was me who suggested using the `Eq` type class. Maybe it wasn't such a great idea (sorry for misleading suggestions). @ImLiar maybe we should/can roll back to `Accept.matches` (or something)?

vkostyukov: Let's do that as a follow-up PR.

PR #906, opened 2018-02-04
Fix Hello World example and update versions

rpless:
Currently, the example errors on compilation with:
```
[error] reference to get is ambiguous
[error] it is imported twice in the same scope by
[error] import io.finch.syntax._
[error] and import io.finch._
```

I also updated the versions in the README so that folks will be directed to the latest version of Finch.


vkostyukov:
Thanks @rpless!

PR #909, opened 2018-02-06
Cheaper Accept

vkostyukov:
Several things going on here:

 - `Accept.fromString` now parses string directly (w/o use of the `MimeType` API)
 - New type-class `Accept.Matcher` allows mapping existing content-types (i.e., JSON, text) onto corresponding `Accept` instances so we don't have to create/allocate those for each part of a coproduct.
 - This new type-class surely complicates things a bit but I think it models the matching more fairly as we shouldn't be parsing Accept headers from content-type values in the first place. Now there is a distinction.

@ImLiar @rpless Mind taking a quick look?

### DIFF
--- a/core/src/main/scala/io/finch/Accept.scala
+++ b/core/src/main/scala/io/finch/Accept.scala
@@ -1,26 +1,72 @@
 package io.finch
 
-import javax.activation.MimeType
-import scala.util.control.NonFatal
+import java.util.Locale
+
+import shapeless.Witness
 
 /**
  * Models an HTTP Accept header (see RFC2616, 14.1).
  *
+ * @note This API doesn't validate the input primary/sub types.
+ *
  * @see https://www.w3.org/Protocols/rfc2616/rfc2616-sec14.html
  */
-final case class Accept(primary: String, sub: String) {
-  def matches(that: Accept): Boolean =
-    (this.primary == that.primary || this.primary == "*") &&
-      (this.sub == that.sub || this.sub == "*")
+abstract class Accept {
+  def primary: String
+  def sub: String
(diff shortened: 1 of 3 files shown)


core/src/main/scala/io/finch/Accept.scala
```diff
@@ -1,26 +1,72 @@
...
+    implicit val html: Matcher[Text.Html] = fromWitness[Text.Html]
+
+    implicit def fromWitness[CT <: String](implicit w: Witness.Aux[CT]): Matcher[CT] = {
+      val slashIndex = w.value.indexOf(47)
+      if (slashIndex == 0 || slashIndex == w.value.length) Empty.asInstanceOf[Matcher[CT]]
+      else new Matcher[CT] {
+        private val primary: String = w.value.substring(0, slashIndex).trim.toLowerCase(Locale.ENGLISH)
+        private val sub: String = w.value.substring(slashIndex + 1, w.value.length).trim.toLowerCase(Locale.ENGLISH)
+        def apply(a: Accept): Boolean =
+          (a.primary == primary || a.primary == "*") && (a.sub == sub || a.sub == "*")
```
optician: Validation accepts a wrong case - `Accept(*, not_star_value)`. It seems like a minor problem but may become complicated due to the lack of logging inside Finch.

Suggestion - `(a.primary == "*" && a.sub == "*") || (a.primary == primary && (a.sub == sub || a.sub == "*"))` .

vkostyukov: Sounds good to me! @ImLiar thoughts?

ImLiar: Yes, it's reasonable

PR #910, opened 2018-02-06
Update to 18.2.0

travisbrown:
I've just published new releases for [Catbird](https://github.com/travisbrown/catbird/releases/tag/v18.2.0) and [iteratee-twitter](https://github.com/travisbrown/iteratee-twitter/releases/tag/v0.20.0).

### DIFF
--- a/core/src/main/scala/io/finch/endpoint/multipart.scala
+++ b/core/src/main/scala/io/finch/endpoint/multipart.scala
@@ -4,9 +4,9 @@ import scala.reflect.ClassTag
 
 import cats.Id
 import cats.data.NonEmptyList
-import com.twitter.finagle.http.exp.{Multipart => FinagleMultipart}
+import com.twitter.finagle.http.exp.{Multipart => FinagleMultipart, MultipartDecoder}
 import com.twitter.finagle.http.exp.Multipart.{FileUpload => FinagleFileUpload}
-import com.twitter.util.{Throw, Try}
+import com.twitter.util.Throw
 import io.catbird.util.Rerunnable
 import io.finch._
 import io.finch.internal.Rs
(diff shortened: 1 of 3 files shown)


core/src/main/scala/io/finch/endpoint/multipart.scala
```diff
@@ -22,7 +22,7 @@ private abstract class Attribute[F[_], A](val name: String, val d: DecodeEntity[
   protected def unparsed(errors: NonEmptyList[Throwable]): Rerunnable[Output[F[A]]]
 
   private def multipart(input: Input): Option[FinagleMultipart] =
-    Try(input.request.multipart).toOption.flatten
```
travisbrown: I changed this to be consistent with the `FileUpload` implementation below, since I had to touch both anyway.

vkostyukov: Nice! Thanks a lot!

I figured we have to do some special plumbing to support the new Multipart API (so we don't decode it twice for two separate endpoints, for example). I'd be happy to give it a shot in the upcoming PR. This looks great regardless - will merge when green!

vkostyukov:
Thanks again for the Catbird and Iteratee releases!

core/src/main/scala/io/finch/endpoint/multipart.scala
travisbrown: Oh, thanks—I didn't look in detail at the change, just made this switch and tests passed, so I figured it was fine. 😄

vkostyukov: Oh yeah, don't worry, it's more than fine!

Issue #902, opened 2018-01-14
path[String] parameters are URL-encoded


matskoff:
Is it supposed to keep path[String] url-encoded?
Here is an example:

```
import io.finch._
import com.twitter.finagle.Http
import com.twitter.util.Await

object Main {
    def main(args: Array[String]): Unit = {
        val api: Endpoint[String] = get("hello" :: path[String] :: param("param")) { (expr1: String, expr2: String) =>
            Ok(s"Passed as path: $expr1. Passed as param: $expr2." )
        }

        Await.ready(Http.server.serve(":8080", api.toServiceAs[Text.Plain]))
    }
}
```

Opening the url: http://localhost:8080/hello/1%2B1?param=1%2B1
I see: Passed as path: 1%2B1. Passed as param: 1+1.

Is different behavior on decoding path and url parameters expected? What is the best way to decode path parameters?

vkostyukov:
Interesting. I'd expect this to be decoded in the output. What do you think @rpless, @ImLiar?

ImLiar:
Yea, I think it should be decoded. 

RFC for URI says that
>   A percent-encoding mechanism is used to represent a data octet in a
   component when that octet's corresponding character is outside the
   allowed set or is being used as a delimiter of, or within, the
   component.

Since path and query parts both are components here, I see no reason to apply different rules

PR #917, opened 2018-02-10
Add path decoding for path[String]

ImLiar:
Addresses https://github.com/finagle/finch/issues/902

Though I suppose that use of a netty-provided solution in a long-run might be not the best pick, it solves the problem and does it _very_ efficiently.
See https://github.com/netty/netty/blob/d672a5a483d0c919279dc9d9724011b6d9dcfd9f/codec-http/src/main/java/io/netty/handler/codec/http/QueryStringDecoder.java#L290)

### DIFF
--- a/core/src/test/scala/io/finch/ExtractPathLaws.scala
+++ b/core/src/test/scala/io/finch/ExtractPathLaws.scala
@@ -1,6 +1,7 @@
 package io.finch
 
 import cats.instances.AllInstances
+import io.netty.handler.codec.http.QueryStringEncoder
 import org.scalacheck.{Arbitrary, Prop}
 import org.typelevel.discipline.Laws
 import scala.reflect.ClassTag
@@ -14,6 +15,7 @@ trait ExtractPathLaws[A]  extends Laws with MissingInstances with AllInstances {
     name = "all",
     parent = None,
     "extractOne" -> Prop.forAll { i: Input =>
+      val encodedInput = i.withRoute(i.route.map(s => new QueryStringEncoder(s).toString))
       val o = one(i)
       val v = i.route.headOption.flatMap(s => decode(s))
 
--- a/core/src/main/scala/io/finch/endpoint/path.scala
+++ b/core/src/main/scala/io/finch/endpoint/path.scala
@@ -3,6 +3,7 @@ package io.finch.endpoint
 import io.catbird.util.Rerunnable
 import io.finch._
 import io.finch.internal.Rs
+import io.netty.handler.codec.http.QueryStringDecoder
 import scala.reflect.ClassTag
 import shapeless.HNil
 
@@ -17,7 +18,7 @@ private class MatchPath(s: String) extends Endpoint[HNil] {
 
 private class ExtractPath[A](implicit d: DecodePath[A], ct: ClassTag[A]) extends Endpoint[A] {
   final def apply(input: Input): Endpoint.Result[A] = input.route match {
-    case s +: rest => d(s) match {
+    case s +: rest => d(QueryStringDecoder.decodeComponent(s)) match {
       case Some(a) =>
         EndpointResult.Matched(input.withRoute(rest), Rerunnable.const(Output.payload(a)))
       case _ =>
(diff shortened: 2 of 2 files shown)


core/src/main/scala/io/finch/DecodePath.scala
```diff
@@ -19,7 +20,7 @@ object DecodePath {
     def apply(s: String): Option[A] = fn( s)
   }
 
-  implicit val decodePath: DecodePath[String] = instance(Some.apply)
+  implicit val decodePath: DecodePath[String] = instance(s => Some(QueryStringDecoder.decodeComponent(s)))
```
vkostyukov: I'm thinking this should be part of `ExtractPath` endpoint (and not the decode-path type-class) so we can provide the same input guarantees across all instances (i.e., `path[Int]`, `path[Boolean]`, etc).

ImLiar: Do you think there is a need to decode special characters for components that are supposed to be `1`, `true`, `false` etc?

vkostyukov: I think so? For example, `path[Int]` should be able to parse `1%2B1` as `1`.

ImLiar: Sorry, didn't get last one

vkostyukov: So my guess is the following should match (and extract `1`):

```scala
scala> path[Int].apply(Input.get("/%31")).awaitValueUnsafe()
res0: Option[Int] = None
```

But I might be totally missing the fact peers never send integers url-encoded?

ImLiar: To my knowledge, only unsafe characters must be encoded within URL

There is the list:
http://www.ietf.org/rfc/rfc1738.txt

vkostyukov: Yeah, maybe also `-` (as in `/-123`)?

The rfc also mentions that peers are welcome to encode other chars as they see a fit:

```
On the other hand, characters that are not required to be encoded
   (including alphanumerics) may be encoded within the scheme-specific
   part of a URL, as long as they are not being used for a reserved
   purpose.
```

vkostyukov: @ImLiar Was wondering what we decided to do here. Maybe just support percent decoding for strings is a good start?

optician: It looks like a more fresh [standard](https://tools.ietf.org/html/rfc3986). 
From #section-2.3:
> URIs that differ in the replacement of an unreserved character with
   its corresponding percent-encoded US-ASCII octet are equivalent: they
   identify the same resource.  However, URI comparison implementations
   do not always perform normalization prior to comparison (see Section
   6).  For consistency, percent-encoded octets in the ranges of ALPHA
   (%41-%5A and %61-%7A), DIGIT (%30-%39), hyphen (%2D), period (%2E),
   underscore (%5F), or tilde (%7E) should not be created by URI
   producers and, when found in a URI, should be decoded to their
   corresponding unreserved characters by URI normalizers.

From #section-2.4
> When a URI is dereferenced, the components and subcomponents
   significant to the scheme-specific dereferencing process (if any)
   must be parsed and separated before the percent-encoded octets within
   those components can be safely decoded, as otherwise the data may be
   mistaken for component delimiters.  The only exception is for
   percent-encoded octets corresponding to characters in the unreserved
   set, which can be decoded at any time.

So suggested approach is to slice URL and then normalize each part if necessary.

ImLiar:
Moved decoding to `ExtractPath`

core/src/test/scala/io/finch/ExtractPathLaws.scala
```diff
@@ -14,6 +15,7 @@ trait ExtractPathLaws[A]  extends Laws with MissingInstances with AllInstances {
     name = "all",
     parent = None,
     "extractOne" -> Prop.forAll { i: Input =>
+      val encodedInput = i.withRoute(i.route.map(s => new QueryStringEncoder(s).toString))
```
vkostyukov: Unsed?

vkostyukov: I'll go ahead and merge this in - we can remove this later.

PR #918, opened 2018-02-15
Add CSS content-type

muuki88:


vkostyukov:
LGTM!

optician:
I wonder what is a criterion to choose predefined media types? I found a [list of types](https://www.iana.org/assignments/media-types/media-types.xhtml) and it's huge.

muuki88:
Adding as needed 😉
Will open a PR for image SVG soon

optician:
What do you think about to add all registered types at once? If the only usage is like constants then it would be straightforward.

muuki88:
Good point . I'll add them tomorrow ( already late in Germany )

rpless:
I'm not sure we want to be in the business of maintaining a huge list of content types. Especially, given that vendor content-types are thing, it might be better to document how to create custom types in the `Cookbook` section of the docs.

muuki88:
I agree on limiting the content types added. I would extend this PR with the [important content types](https://developer.mozilla.org/en-US/docs/Web/HTTP/Basics_of_HTTP/MIME_types) listed on Apache MDN.

IMHO the basic content types should already be provided.

vkostyukov:
I think I don't mind including these new types but I'd probably argue we might want to move all of them to a separate file (i.e., `contentTypes.scala`), while also including a link to that Mozilla doc.

muuki88:
Sounds reasonable. Should this be a trait and the package object extends it, or can it be a breaking change and live in a separate object?

vkostyukov:
We can just define those as top-level objects within a new file, i.e., `contentTypes.scala`. This should classify as a source compatible change so we should be fine.

vkostyukov:
Merging this. Thanks @muuki88!

muuki88:
You're welcome 🤩

PR #922, opened 2018-02-26
Endpoint composition (product) should use "/" instead of " :: " in toString()

slouc:
Proposing to use / instead of :: as character in endpoint string concatenation for endpoints composition.

I don't know of any use cases for endpoints toString() method other than logging purposes, and using "/" instead of " :: " doesn't take anything away from that. But it adds the URL-like property to it, so it can e.g. be directly put in the Location header upon 201 response. The way things are now, I have to do:

```
val root = "foo" :: "bar" :: "baz"

Created(someEntity)
  .withHeader(`Location` -> s"/${root.toString.replace(" :: ", "/")}/${someEntity.id}")
```

where `${root.toString.replace(" :: ", "/")}` could have been simply `$root`.

URL path is unlikely to change, and HTTP endpoints will be using forward slashes for a long time to come, but I can't say the same for :: which seems to be an arbitrary choice following a 3rd party library convention.

If this is not acceptable, perhaps :: could at least be defined as a constant, something along the lines of `Endpoint.compositionStringDelimiter`, so that I can use that in my replace function and not care should the symbol for composition suddenly change tomorrow.

### DIFF
--- a/core/src/main/scala/io/finch/Endpoint.scala
+++ b/core/src/main/scala/io/finch/Endpoint.scala
@@ -198,7 +198,7 @@ trait Endpoint[A] { self =>
       final def apply(input: Input): Endpoint.Result[pa.Out] = inner(input)
 
       override def item = items.MultipleItems
-      final override def toString: String = s"${other.toString} :: ${self.toString}"
+      final override def toString: String = s"${other.toString}/${self.toString}"
     }
 
   /**
--- a/core/src/test/scala/io/finch/EndpointSpec.scala
+++ b/core/src/test/scala/io/finch/EndpointSpec.scala
@@ -164,7 +164,7 @@ class EndpointSpec extends FinchSpec {
 
     check { (s: String, i: Int) => path(s).map(_ => i).toString === s }
     check { (s: String, t: String) => (path(s) :+: path(t)).toString === s"($s :+: $t)" }
-    check { (s: String, t: String) => (path(s) :: path(t)).toString === s"$s :: $t" }
+    check { (s: String, t: String) => (path(s) :: path(t)).toString === s"$s/$t" }
     check { s: String => path(s).product[String](*.map(_ => "foo")).toString === s }
     check { (s: String, t: String) => path(s).mapAsync(_ => Future.value(t)).toString === s }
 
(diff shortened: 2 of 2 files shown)


core/src/main/scala/io/finch/Endpoint.scala
```diff
@@ -198,7 +198,7 @@ trait Endpoint[A] { self =>
       final def apply(input: Input): Endpoint.Result[pa.Out] = inner(input)
 
       override def item = items.MultipleItems
-      final override def toString: String = s"${other.toString} :: ${self.toString}"
+      final override def toString: String = s"${other.toString}/${self.toString}"
```
vkostyukov: I wonder if we should also alter the `coproduct` to-string. I'd imagine, something like `\n` (a new line) would be quite human-readable. What do you think?

optician: Hm, little imagination
route - `root :: (("admin" :: header :: (addGoods :+: ...)) :+: (("goods" :: id) :+: ("cart" :: cookies) :+: "about" :+: ...))`
string representation - 
```
"root" / ( 
    "admin" / header / ( addGoods
    ... 
)
(
    ( "goods" / id)
    ( "cart" / cookies)
    "about"
    ...
)
```
I'm not sure it's somehow readable.

slouc: @vkostyukov I'm sorry, I wasn't paying attention and I wrote "coproduct" in the commit and PR description, where I obviously meant "product" (or "composition"). I edited the PR description and force pushed the amended commit with a proper description (I used the --force-with-lease flag which stops the push in case someone else already pushed something to the remote).

Yes, I think newline could work for coproduct. I feel like @optician tried to make a point of how clumsy such string representation would be, but I actually like it.

ImLiar:
Hi @slouc 
I don't think even your `Location` example might work for any endpoint with dynamic parameters, headers and so on, essentially anything aport constant strings.
Does it really make sense to change the delimiter here?
@vkostyukov I think 0.18 / 0.19 might be the right place to finally push our efforts towards endpoints metadata 😃

slouc:
@ImLiar I don't have a problem with dynamic parameters because I only convert the static part to string and manually append the dynamic part(s).

I think that this change, combined with what @vkostyukov suggested for coproduct, really adds more clarity to the string representation of Endpoint as it resembles a simple routes list, as opposed to the combination of `::` and `:+:` operators.

What's planned in the endpoints metadata change?

PR #924, opened 2018-03-06
Use a dedicated typeclass to guide coproduct induction

### LINKED ISSUE #923: toService with Circe takes forever to compile for large Coproducts
Looks like `valueToResponse` might send the compiler on a wild goose chase to look for `Encoder` instances at each step of the `Coproduct` induction, but we never really want to do that!
I.e. `Encoder` instance of the `Coproduct` tails.

joroKr21:
This disentangles the `Coproduct` induction logic in `ToResponse`
derivation from the actual encoding of values.

The net result is greatly reduced compilation times for large
`Coproducts`.

Fixes #923

### DIFF
--- a/core/src/main/scala/io/finch/ToResponse.scala
+++ b/core/src/main/scala/io/finch/ToResponse.scala
@@ -124,14 +124,29 @@ object ToResponse extends HighPriorityToResponseInstances {
     }
   }
 
-  implicit def cnilToResponse[CT <: String]: Aux[CNil, CT] =
-    instance((_, _) => Response(Version.Http10, Status.NotFound))
+  trait FromCoproduct[C <: Coproduct] extends ToResponse[C]
 
-  implicit def coproductToResponse[H, T <: Coproduct, CT <: String](implicit
-    trH: ToResponse.Aux[H, CT],
-    trT: ToResponse.Aux[T, CT]
-  ): Aux[H :+: T, CT] = instance {
-    case (Inl(h), cs) => trH(h, cs)
-    case (Inr(t), cs) => trT(t, cs)
+  object FromCoproduct {
+    type Aux[C <: Coproduct, CT] = FromCoproduct[C] { type ContentType = CT }
+
+    def instance[C <: Coproduct, CT <: String](fn: (C, Charset) => Response): Aux[C, CT] =
+      new FromCoproduct[C] {
+        type ContentType = CT
+        def apply(c: C, cs: Charset): Response = fn(c, cs)
+      }
+
+    implicit def cnilToResponse[CT <: String]: Aux[CNil, CT] =
+      instance((_, _) => Response(Version.Http10, Status.NotFound))
+
+    implicit def cconsToResponse[L, R <: Coproduct, CT <: String](
+      implicit trL: ToResponse.Aux[L, CT], fcR: Aux[R, CT]
+    ): Aux[L :+: R, CT] = instance {
+      case (Inl(h), cs) => trL(h, cs)
+      case (Inr(t), cs) => fcR(t, cs)
+    }
   }
+
+  implicit def coproductToResponse[C <: Coproduct, CT <: String](
+    implicit fc: FromCoproduct.Aux[C, CT]
+  ): Aux[C, CT] = fc
 }


joroKr21:
I will think about adding a test

vkostyukov:
Very interesting. How much of a compile time improvement we're talking about?

joroKr21:
WIP for now as I want to see if shapeless has a builtin combinator we can use without sacrificing performance. Maybe something like `LiftAll` and `Unifier`? or `LeftFolder` / `RightFolder`.

In one case I cared to measure the compile time improved from ~4min to ~6sec.

vkostyukov:
Nice! Thanks for the extra details @joroKr21! Yeah, I'd be interested in seeing if Shapeless has anything we'd be able to reuse. Also, I'm not minding this thin type-class at all. Just let me know when you think it's ready for review.

joroKr21:
It seems like `Folder` is exactly what we want, but the phantom `ContentType` parameter makes it difficult to express.

vkostyukov:
I'd personally be fine with this new type-class. Let's just make sure we get more people to take a look.

core/src/main/scala/io/finch/ToResponse.scala
```diff
@@ -124,14 +124,29 @@ object ToResponse extends HighPriorityToResponseInstances {
...
 
+  implicit def coproductToResponse[C <: Coproduct, CT <: String](
+    implicit ctr: CoproductToResponse.Aux[C, CT]
+  ): Aux[C, CT] = ctr
+}
+
+trait CoproductToResponse[C <: Coproduct] extends ToResponse[C]
+object CoproductToResponse {
+  type Aux[C <: Coproduct, CT] = CoproductToResponse[C] { type ContentType = CT }
+  def apply[C <: Coproduct](implicit ctp: CoproductToResponse[C]): Aux[C, ctp.ContentType] = ctp
```
vkostyukov: Should we drop this `apply` method? I don't see it's being used anywhere?

core/src/main/scala/io/finch/ToResponse.scala
```diff
@@ -124,14 +124,29 @@ object ToResponse extends HighPriorityToResponseInstances {
     }
   }
 
+  implicit def coproductToResponse[C <: Coproduct, CT <: String](
+    implicit ctr: CoproductToResponse.Aux[C, CT]
+  ): Aux[C, CT] = ctr
+}
+
+trait CoproductToResponse[C <: Coproduct] extends ToResponse[C]
```
vkostyukov: I'd add an extra new line here.

vkostyukov: Just an idea. How do you feel about moving this new typeclass under the `ToResponse` companion object? This way we can reduce the top-level/user-facing API surface. I'd imagine something like `ToResponse.FromCoproduct` should do it:

```scala
object ToReponse {
  type Aux = ...
  trait FromCoprodct extends ToResponse[...]
}

```

ImLiar:
LGTM  👍

core/src/main/scala/io/finch/ToResponse.scala
joroKr21: Fine by me, I was wondering the same thing :)

core/src/main/scala/io/finch/ToResponse.scala
joroKr21: Just habit, I will remove it.

core/src/main/scala/io/finch/ToResponse.scala
vkostyukov: Unless you already working on it, I'd be happy to see this merged in its current state and then addressing this comment later. Whatever you prefer - just le me know.

travisbrown:
It's surprising to me that this would have such a big impact on compile times, but 👍.

vkostyukov:
FWIW, I tried compiling `examples` project with it - no noticeable impact. Perhaps, the examples are too simple for this to "kick in".

joroKr21:
I addressed the comments. @vkostyukov as you noticed there is no easy way to test this.

I published locally the changes here and noticed a tiny slowdown of about 10sec compared to the workaround I have in our codebase, but it's still a huge improvement wrt the baseline. The difference in the workaround is a specialized `toCoproductService` that takes an implicit `ToResponse.FromCoproduct` instead of the more general `ToResponse`.

Anyway, I think it's fine like this.

vkostyukov:
Love it! Thank you @joroKr21!

PR #927, opened 2018-03-14
Support MethodNotAllowed via new EndpointResult

vkostyukov:
This is essentially an iteration on #883 that instead of capturing a matched method as part of `Input`, introduces a new `EndpointResult` case. I like this approach better because:

 - It's compliant with https://www.w3.org/Protocols/rfc2616/rfc2616-sec10.html (10.4.6)
 - It's more efficient (`Input` remains skinner)
 - It keeps a reasonable semantic for `Endpoint.apply(Input)` for endpoint-mappers so REPL/test environments will return "not-matched" when HTTP verb is different.

Example:

```scala
scala> val a = get("foo") { Ok("foo") }
a: io.finch.Endpoint[String] = GET /foo

scala> val b = post("foo") { Ok("post foo") }
b: io.finch.Endpoint[String] = POST /foo

scala> val c = put("foo") { Ok("put foo") }
c: io.finch.Endpoint[String] = PUT /foo

scala> val s = Bootstrap.configure(enabledMethodNotAllowed = true).serve[Text.Plain](a :+: b).serve[Text.Plain](c).toService
s: com.twitter.finagle.Service[com.twitter.finagle.http.Request,com.twitter.finagle.http.Response] = <function1>

scala> Http.server.serve(":8081", s)
```

And then:

```
[tw-mbp-vkostyukov finch (vk/match-wo-method)]$ http :8081/bar
HTTP/1.1 404 Not Found
Content-Length: 0
Date: Fri, 16 Mar 2018 20:59:30 GMT
Server: Finch



[tw-mbp-vkostyukov finch (vk/match-wo-method)]$ http POST :8081/foo
HTTP/1.1 200 OK
Content-Type: text/plain
Date: Fri, 16 Mar 2018 20:59:37 GMT
Server: Finch
content-encoding: gzip
content-length: 34

post foo

[tw-mbp-vkostyukov finch (vk/match-wo-method)]$ http PUT :8081/foo
HTTP/1.1 200 OK
Content-Type: text/plain
Date: Fri, 16 Mar 2018 20:59:48 GMT
Server: Finch
content-encoding: gzip
content-length: 33

put foo

[tw-mbp-vkostyukov finch (vk/match-wo-method)]$ http DELETE :8081/foo
HTTP/1.1 405 Method Not Allowed
Allow: PUT,GET,POST
Content-Length: 0
Date: Fri, 16 Mar 2018 20:59:56 GMT
Server: Finch
```

### DIFF
--- a/core/src/main/scala/io/finch/EndpointResult.scala
+++ b/core/src/main/scala/io/finch/EndpointResult.scala
@@ -1,5 +1,6 @@
 package io.finch
 
+import com.twitter.finagle.http.Method
 import com.twitter.util.{Await, Duration, Future, Try}
 import io.catbird.util.Rerunnable
 
@@ -7,8 +8,13 @@ import io.catbird.util.Rerunnable
  * A result returned from an [[Endpoint]]. This models `Option[(Input, Future[Output])]` and
  * represents two cases:
  *
- *  - Endpoint is matched so both `remainder` and `output` is returned.
- *  - Endpoint is skipped so `None` is returned.
+ *  - Endpoint is matched (think of 200).
+ *  - Endpoint is not matched (think of 404, 405, etc).
+ *
+ * In its current state, `EndpointResult.NotMatched` represented with two cases:
+ *
+ *  - `EndpointResult.NotMatched` (very generic result usually indicating 404)
+ *  - `EndpointResult.NotMatched.MethodNotAllowed` (indicates 405)
  *
  * API methods exposed on this type are mostly introduced for testing.
  *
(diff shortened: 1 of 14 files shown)


core/src/main/scala/io/finch/EndpointResult.scala
```diff
@@ -93,11 +94,15 @@ sealed abstract class EndpointResult[+A] {
 
 object EndpointResult {
 
-  case object Skipped extends EndpointResult[Nothing] {
+  final case class Matched[A](rem: Input, out: Rerunnable[Output[A]]) extends EndpointResult[A] {
+    def isMatched: Boolean = true
+  }
+
+  case object NotMatched extends EndpointResult[Nothing] {
```
vkostyukov: I want to look into this to see if it could be modeled better so we can avoid `asInstanceOf` casts in `Endpoint`.

vkostyukov:
@sergeykolbasov Do you mind taking a look?

vkostyukov:
See 10.4.6 here https://www.w3.org/Protocols/rfc2616/rfc2616-sec10.html.

Looks like we also need to add `Allow` header. This means changing the implementation a bit.

sergeykolbasov:
Sounds like a proper solution to me 👍

PR #937, opened 2018-05-09
Update to 18.5.0

travisbrown:
I've published releases for catbird and iteratee-twitter.


vkostyukov:
Thanks a lot, @travisbrown! Appreciate the quick releases of literature and catbird!

PR #944, opened 2018-06-10
Add finch-refined module with refined types support

sergeykolbasov:
This PR enables support for [refined](https://github.com/fthomas/refined) types in query parameters, path segments, and other request entities.

Personally, I find it very handy for validation on the type level. This library is also supported by 
circe, so users are able to use refined types in DTOs along with JSON decoders, but it's optional.

- [x] Decoders
- [x] Specs
- [x] Documentation

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -286,7 +286,7 @@ lazy val refined = project
   .settings(
     libraryDependencies ++= Seq(
       "eu.timepit" %% "refined" % refinedVersion,
-      "eu.timepit" %% "refined-cats" % refinedVersion,
+      "eu.timepit" %% "refined-cats" % refinedVersion % "test",
       "eu.timepit" %% "refined-scalacheck" % refinedVersion % "test"
     )
   )
--- a/build.sbt
+++ b/build.sbt
@@ -20,6 +20,7 @@ lazy val argonautVersion = "6.2.1"
 lazy val json4sVersion = "3.5.3"
 lazy val iterateeVersion = "0.17.0"
 lazy val iterateeTwitterVersion = "18.5.0"
+lazy val refinedVersion = "0.9.0"
 
 lazy val compilerOptions = Seq(
   "-deprecation",
@@ -106,7 +107,7 @@ lazy val publishSettings = Seq(
         <url>https://twitter.com/ryan_plessner</url>
       </developer>
       <developer>
-        <id>ImLiar</id>
+        <id>sergeykolbasov</id>
         <name>Sergey Kolbasov</name>
         <url>https://twitter.com/sergey_kolbasov</url>
       </developer>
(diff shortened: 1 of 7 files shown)


travisbrown:
👍 

This looks great to me!

docs/src/main/tut/user-guide.md
```diff
@@ -886,6 +886,31 @@ val bufEnumerator =
...
+
+val e: Endpoint[Int] = get("foo" :: param[Int Refined Positive]("int")) { (i: Int Refined Positive) =>
+  Ok(i.value)
+}
+
+val u: Endpoint[URL] = get("foo" :: param[String Refined Url]("url")) { (s: String Refined Url) =>
+  Ok(new URL(s.value))
+}
+
+//Throw(io.finch.Error.NotParsed: param 'int' cannot be converted to Refined: Predicate failed: (-1 > 0)..)
```
rpless: Probably don't need this comment here. `tut` will render the resulting `Throw` in the docs.

sergeykolbasov: Sure, thanks

clhodapp:
This is great but you really ought to make it work for any `RefType`. I know that I, for one, tend to use Refined with Shapeless tags.

Here's an example of how to make it generic in this way:
```
  implicit def decodeEntityRefined[F[_, _], T, P](implicit
    rt: RefType[F],
    decodeT: DecodeEntity[T],
    validate: Validate[T, P]
  ): DecodeEntity[F[T, P]] = { s =>
    decodeT(s).flatMap { t =>
      rt.refine[P](t) match {
        case Right(value) => Return(value)
        case Left(errString) => {
          Throw(new RuntimeException(errString))
        }
      }
    }
  }
```

The effect of doing it this way is that you get a `DecodeEntity` for both `T Refined P` and `T @@ P`.

build.sbt
```diff
@@ -279,6 +280,18 @@ lazy val sse = project
   .settings(allSettings)
   .dependsOn(core)
 
+lazy val refined = project
+  .settings(moduleName := "finch-refined")
+  .settings(allSettings)
+  .settings(
+    libraryDependencies ++= Seq(
+      "eu.timepit" %% "refined" % refinedVersion,
+      "eu.timepit" %% "refined-cats" % refinedVersion,
```
fthomas: It looks like refined-cats could be a test dependency, too.

sergeykolbasov: Good catch, thanks

sergeykolbasov:
@clhodapp Thanks for the example, sounds reasonable

refined/src/main/scala/io/finch/refined/PredicateFailed.scala
```diff
@@ -0,0 +1,9 @@
+package io.finch.refined
+
+import scala.util.control.NoStackTrace
+
+case class PredicateFailed(error: String) extends Exception with NoStackTrace {
```
vkostyukov: Maybe we can throw `Error.NotValid` for now? We can supply it with a generic `RequestItem` for now.

sergeykolbasov: That was actually a problem as I had no `RequestItem` in scope.

I think it's fine as it is now, because `param` and other endpoints will wrap it into `Error.NotParsed`:
https://github.com/finagle/finch/blob/master/core/src/main/scala/io/finch/endpoint/param.scala#L95

vkostyukov: Oh I see. That makes sense!

fommil:
btw, ever since https://github.com/fthomas/refined/pull/447 (and the cats equivalent) these modules aren't strictly necessary to get `Refined` support in your library.

PR #947, opened 2018-06-15
Update to 18.6.0

travisbrown:
I've just published catbird and iteratee-twitter, so this should be ready to go. I've also bumped a few other little things while I was here.


vkostyukov:
Thank you, @travisbrown! I will cut a Finch release today.

Issue #953, opened 2018-06-20
EndpointResult.Matched may need to contain the "matched route"


vkostyukov:
See [Gitter room](https://gitter.im/finagle/finch?at=5b2a8a717d3bca737af5f4cf) for more details.

If we had that, we would be able to employ this information in `ToService` and

 - Extract the matched route into a Response context (see #855)
 - Export some useful metrics (see #845)

Obviously, this comes at the cost of extra allocations.

vkostyukov:
Implemented in #957.

PR #957, opened 2018-07-05
Introduce Trace (what endpoint was matched) [RFC]

vkostyukov:
We've been talking quite a bit about the ways to provide users with a context of what endpoint was actually matched (see #953 for a top-level issue). This wasn't possible today.

Both of our attempts to built-in metrics/telemetry failed because of that (see #845 and #855) as we didn't have the right tools to implement it.

I've been playing with the idea of including the matched path into an `EndpointResult` so it could be propagated along the request/input path. This way, at the very bottom (in `ToService`) we can extract this matched path and do something useful with that, for example, report metrics.

This PR [RFC] introduces a new concept `io.finch.Trace` that represent a matched path of an endpoint and is optimized for low footprint and fast concat. Right now, we aren't doing anything useful with it in `ToService` as I wanted to make sure we settle on a general idea before I proceed with actually employing the new data.

Let me know what do you think! Here is a quick usage example.

```scala
scala> import io.finch._, io.finch.syntax._
import io.finch._
import io.finch.syntax._

scala> val foo = get("foo" :: "bar" :: path[String]) { s: String => Ok(s) }
foo: io.finch.Endpoint[String] = GET /foo :: bar :: :string

scala> val bar = get("bar" :: "foo" :: path[Int]) { i: Int => Ok(i) }
bar: io.finch.Endpoint[Int] = GET /bar :: foo :: :int

scala> val fooBar = foo :+: bar
fooBar: io.finch.Endpoint[String :+: Int :+: shapeless.CNil] = (GET /foo :: bar :: :string :+: GET /bar :: foo :: :int)

scala> fooBar(Input.get("/foo/bar/baz")).trace
res0: Option[io.finch.Trace] = Some(/foo/bar/:string)

scala> fooBar(Input.get("/bar/foo/10")).trace
res1: Option[io.finch.Trace] = Some(/bar/foo/:int)
```

### DIFF
--- a/core/src/main/scala/io/finch/Endpoint.scala
+++ b/core/src/main/scala/io/finch/Endpoint.scala
@@ -81,8 +81,8 @@ trait Endpoint[A] { self =>
       final def apply(oa: Output[A]): Future[Output[B]] = oa.traverse(fn)
 
       final def apply(input: Input): Endpoint.Result[B] = self(input) match {
-        case EndpointResult.Matched(rem, out) =>
-          EndpointResult.Matched(rem, out.flatMapF(this))
+        case EndpointResult.Matched(rem, trc, out) =>
+          EndpointResult.Matched(rem, trc, out.flatMapF(this))
         case skipped: EndpointResult.NotMatched => skipped
       }
 
@@ -104,8 +104,8 @@ trait Endpoint[A] { self =>
       final def apply(oa: Output[A]): Future[Output[B]] = oa.traverseFlatten(fn)
 
       final def apply(input: Input): Endpoint.Result[B] = self(input) match {
-        case EndpointResult.Matched(rem, out) =>
-          EndpointResult.Matched(rem, out.flatMapF(this))
+        case EndpointResult.Matched(rem, trc, out) =>
+          EndpointResult.Matched(rem, trc, out.flatMapF(this))
         case skipped: EndpointResult.NotMatched => skipped
       }
 
@@ -131,8 +131,8 @@ trait Endpoint[A] { self =>
   final def transform[B](fn: Future[Output[A]] => Future[Output[B]]): Endpoint[B] =
     new Endpoint[B] {
       final def apply(input: Input): Endpoint.Result[B] = self(input) match {
-        case EndpointResult.Matched(rem, out) =>
-          EndpointResult.Matched(rem, Rerunnable.fromFuture(fn(out.run)))
+        case EndpointResult.Matched(rem, trc, out) =>
+          EndpointResult.Matched(rem, trc, Rerunnable.fromFuture(fn(out.run)))
         case skipped: EndpointResult.NotMatched => skipped
       }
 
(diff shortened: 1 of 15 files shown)


sergeykolbasov:
Looks awesome, I was about to land the very similar PR but with trace replaced by `Endpoint.Meta` where meta is just a type alias to string.

My point is that `Trace` looks like the very lightweight & cut version of metadata ADT and I wonder if we are able to extend its use to the API users anyhow.

sergeykolbasov:
Maybe it makes sense to introduce `lazy val metadata` in Endpoint and during the compilation phase, all `HList` and `Coproduct` endpoints will compose their metadata together. It means, there will be fewer allocations spent on each call for `concat` operation as it is now.

@vkostyukov what are your thoughts?

vkostyukov:
@sergeykolbasov Yeah, I think exporting the `Trace` object under a response's record schema is a reasonable thing to do at this point.

Re: Meta. I'm perfectly fine if we keep metadata's purpose as a structure defining composed endpoints (built at endpoint creation), while trace could carry the information about what endpoints were matched during this evaluation (built when an endpoint is run). I hear you, we can populate `Trace` with metadata (similar to how we do it with `String`s) in case if users would want that information at the end. As of now, there is an immediate use-case a `Trace` structure is trying to address - figuring out what endpoint was matched so we can report telemetry. In my mind's eye, we should stick with it and see if people need anything more than that.

core/src/main/scala/io/finch/Endpoint.scala
```diff
@@ -174,9 +174,13 @@ trait Endpoint[A] { self =>
...
 
       final def apply(input: Input): Endpoint.Result[O] = self(input) match {
-        case a @ EndpointResult.Matched(_, _) => other(a.rem) match {
-          case b @ EndpointResult.Matched(_, _) =>
-            EndpointResult.Matched(b.rem, a.out.liftToTry.product(b.out.liftToTry).flatMapF(this))
+        case a @ EndpointResult.Matched(_, _, _) => other(a.rem) match {
+          case b @ EndpointResult.Matched(_, _, _) =>
+            EndpointResult.Matched(
+              b.rem,
+              a.trc.concat(b.trc),
```
sergeykolbasov: what if we move this `concat` to `private val trc = a.trc.concat(b.trc)` of the endpoint to generate it only once?

vkostyukov: Sorry, I'm not sure this is possible. Neither `a.trc` nor `b.trc` are available until we run an endpoint (`Endpoint.apply`).

sergeykolbasov: Ah, yeah

nvm then.

sergeykolbasov:
LGTM 👍

vkostyukov:
@rpless - that's a really good point about the failed request. We, however, have a top-level failure handler that always catches everything and converts it into a very basic 400 response. I think we should be fine with just setting the `response.ctx` field.

The new structure definitely comes at the cost yet I think we're in the position to pay it. Some recent performance optimizations should make it more-or-less even. I want to run some benchmarks to see the overhead.

vkostyukov:
There is a slight (and expected) increase for path endpoints but nothing concerning:

```
BEFORE:

[info] MatchPathBenchmark.stringSome                                avgt   10  207.146 ±   10.804   ns/op
[info] MatchPathBenchmark.stringSome:·gc.alloc.rate.norm            avgt   10  280.000 ±    0.001    B/op

AFTER:

[info] MatchPathBenchmark.stringSome                                avgt   10  216.268 ±  57.560   ns/op
[info] MatchPathBenchmark.stringSome:·gc.alloc.rate.norm            avgt   10  304.000 ±   0.001    B/op
```

PR #959, opened 2018-07-07
Add tracing to response context

sergeykolbasov:
- Add another option to `Bootstrap`
- If tracing is enabled, add `Trace` from `EndpointResult.Matched` to `Response.ctx`

### DIFF
--- a/core/src/main/scala/io/finch/Bootstrap.scala
+++ b/core/src/main/scala/io/finch/Bootstrap.scala
@@ -30,6 +30,9 @@ import shapeless._
  * - `enableMethodNotAllowed` (default: `false`): whether or not to enable 405 MethodNotAllowed HTTP
  *   response (see RFC2616, section 10.4.6)
  *
+ * - `enableTracing` (default: `false`): whether or not to enable tracing of called endpoint propagated
+ *   through Response.ctx
+ *
  * @see https://www.w3.org/Protocols/rfc2616/rfc2616-sec14.html
  * @see https://www.w3.org/Protocols/rfc2616/rfc2616-sec12.html
  * @see https://www.w3.org/Protocols/rfc2616/rfc2616-sec10.html
@@ -39,12 +42,18 @@ class Bootstrap[ES <: HList, CTS <: HList](
     val includeDateHeader: Boolean = true,
     val includeServerHeader: Boolean = true,
     val negotiateContentType: Boolean = false,
-    val enableMethodNotAllowed: Boolean = false) { self =>
+    val enableMethodNotAllowed: Boolean = false,
+    val enableTracing: Boolean = false) { self =>
 
   class Serve[CT] {
     def apply[E](e: Endpoint[E]): Bootstrap[Endpoint[E] :: ES, CT :: CTS] =
       new Bootstrap[Endpoint[E] :: ES, CT :: CTS](
-        e :: self.endpoints, includeDateHeader, includeServerHeader, negotiateContentType, enableMethodNotAllowed
+        e :: self.endpoints,
+        includeDateHeader,
+        includeServerHeader,
+        negotiateContentType,
+        enableMethodNotAllowed,
+        enableTracing
       )
     }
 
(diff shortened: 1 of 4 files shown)


vkostyukov:
As @rpless pointed out, if the endpoint fails, there is no response produced so the failure-path isn't covered. We could probably put `Trace` into `Request` as well, but I'm not sure I like this behavior: a user would need to make sure response completes before querying the request context.

I have a couple of ideas in mind with regards to alternatives:

 - Configure `Bootstrap` with a user-defined export side-effect, i.e. `onTrace: Trace => Unit` so they can export/log the trace themselves
 - Actually, export some metrics into a given `StatsReciever`. We would need to settle on some initial format but nothing stops us from allowing users to override it.

sergeykolbasov:
@vkostyukov Yep, I missed this one.
Personally, I'd prefer to have side-effect function. Going to implement it

sergeykolbasov:
On the other hand, `Trace => Unit` doesn't give any information about endpoint latency. It seems that stats receiver is the only meaningful option

vkostyukov:
Yeah, good point about latency. Maybe we should just settle on some format and export whatever we can to a given stats receiver? Again, we can have people opt-in for it and see how it goes. As I mentioned, figuring out the format may be non-trivial. Besides, I'm not super sure we need to offer latency at first (it maybe be too noisy and metrics aren't cheap).

How does the following sound for the `GET /foo/bar/:baz` endpoint:

 - `get/foo/bar/:baz/2xx` - succeses
 - `get/foo/bar/:baz/4xx`- failures
 - `get/foo/bar/:baz/5xx` - server errors

If we ever need to mix in latency there, we can do

 - `get/foo/bar/:baz/latency_ms` - latency historgram

rpless:
I'm not sure we'll be able to export an accurate latency (or at least the latency that most people would expect). Any latency that's measured here will not account for any latency introduced by Filters (which my guess is that users would prefer to measure the roundtrip as opposed to just inside the Finch endpoint). My preference here would be to provide users a hook into the Trace. We can still implement a default thing that uses this to dump a predetermined format to a StatsReceiver. But this might allow users to come up with their own mechanisms to get accurate latencies in the future.

I'll also mention again, I think using `Request.ctx` instead of `Response.ctx` would let us hook into this at the filter level even when Finch fails the future because the Filter would still have access to the request. Not sure if there is something I don't know about `Request.ctx` that makes this bad or undesirable, beyond the fact that it feels kinda weird to attach to the request instead of the response. With this method we could also provide a default filter with the suggested formats above.

Either way, I think we should leave it configureable at the bootstrap level.

vkostyukov:
Thinking out loud, what if we change Finch's default error handler to cover for all possible exceptions so we always have a `Response` within a derived service (`Bootstrap.toService`) we can attach a trace to. I'm trying to figure out if that would change any behavior and it seems to me, nothing's really changing as [Finagle does the very same thing a little later](https://github.com/twitter/finagle/blob/develop/finagle-base-http/src/main/scala/com/twitter/finagle/http/codec/HttpServerDispatcher.scala#L44).

rpless:
@vkostyukov That seems reasonable to me. Just add a catch-all that responds with a 500. The differences in behavior I can think are:
- If you don't have an error handling in your filters, you get a warning logged about `Exception propagated to the default monitor` and then finagle returns a 500. In this case I think the behavior is the same enough that it doesn't matter.
- If you do have error handling related to in your filters then it would probably break some expectations on the User's side.

Maybe its worth asking in the gitter channel to see if anyone does this?

vkostyukov:
As I think about this, the dispatcher feeds requests into a service that's materialized off an entire stack, which means it won't be observing errors if we handle them ourselves. This could possibly affect lots of different sub-systems including telemetry and nacking (as many of those kick-in on failed futures). So handling everything isn't probably a good option.

sergeykolbasov:
Can we somehow carry it in the future itself?

vkostyukov:
There are Future Locals but they work in the opposite (to a data flow) direction. In Finagle, we've been working this around with a top-level `Local.let` that carries some mutable container that's populated later. But as we're not controlling user-defined filters in Finch, there is no way to do that.

Honestly, sounds like we've got just two options:
 - Report something (see[ this comment](https://github.com/finagle/finch/pull/959#issuecomment-403305443))
 - Set `Request`s context as the endpoint is matched (perhaps we can also set the `Response` context)

vkostyukov:
Speaking of a workaround I've been mentioning. Perhaps it could be a reasonable middle ground allowing people to retrieve a `Trace` instance from an underlying endpoint assuming they opt-in for it explicitly. The trick is to set some _mutable_ container that earlier in the stack that we can mutate later (in `ToService`). This would mean, users would need to supply that container in a local, in one of the filters. We, however, can hide it behind a pretty reasonable API.

Here is a sketch:

```
object Trace {
  private class Captured(var trace: Trace)
  private val capturedLocal: Local[Captured]

  def capture[A](f: => A): A = capturedLocal.let(new Captured(empty))(f)
  def captured: Trace = capturedLocal() match {
    case Some(c) => c.trace
    case None => empty
  }
}
```

Within `ToService`, we can mutate it:

```
Trace.capturedLocal() {
  case Some(c) => // a user supplied captured instance so we can mutate it
    c.trace = endpoint's trace
  case None => // a user didn't do `Trace.capture` so there is no place for us to put trace in
}
```

And then, in a user-defined filter the usage would look:

```
class Filter {
  def apply(req, service) = {
    Trace.capture {
      service(req).respond { _ =>
        Console.println(Trace.captured)
      }
    }
  }
}
```

Given users would need to call `Trace.capture` explicitly, we probably don't have a need for an extra boostrap option so there is already a knob that can turn it on/off.

I think finagle-mux uses very similar idea to propagate it's contexts (headers).

vkostyukov:
I put a quick prototype together in #960. Let me know what do you think!

spockz:
It is really similar to what we had. Just added one question, otherwise LGTM.

vkostyukov:
Suppressed by #960.

PR #960, opened 2018-07-12
Capture Traces

vkostyukov:
As discussed in #959. Here is how it works (from the user perspective).

```scala
scala> import io.finch._, io.finch.syntax._
import io.finch._
import io.finch.syntax._

scala> val foo = get("foo" :: path[String]) { s: String => Ok(s) }
foo: io.finch.Endpoint[String] = GET /foo :: :string

scala> import com.twitter.finagle.http.Request
import com.twitter.finagle.http.Request

scala> val s = foo.toServiceAs[Text.Plain]
s: com.twitter.finagle.Service[com.twitter.finagle.http.Request,com.twitter.finagle.http.Response] = io.finch.ToService$$anon$4$$anon$2

scala> Trace.capture { s(Request("/foo/bar")).map(_ => Trace.captured) }
res0: com.twitter.util.Future[io.finch.Trace] = Promise@406161512(state=Done(Return(/foo/:string)))
```

### DIFF
--- a/core/src/main/scala/io/finch/ToService.scala
+++ b/core/src/main/scala/io/finch/ToService.scala
@@ -101,9 +101,9 @@ object ToService {
         private[this] val underlying = es.head.handle(respond400OnErrors)
 
         def apply(req: Request): Future[Response] = underlying(Input.fromRequest(req)) match {
-          case EndpointResult.Matched(rem, _, out) if rem.route.isEmpty =>
+          case EndpointResult.Matched(rem, trc, out) if rem.route.isEmpty =>
 
-            // TODO: Use Trace to report telemetry.
+            Trace.captureIfNeeded(trc)
 
             val accept =
               if (opts.negotiateContentType) req.accept.map(a => Accept.fromString(a)) else Nil
--- a/core/src/main/scala/io/finch/Trace.scala
+++ b/core/src/main/scala/io/finch/Trace.scala
@@ -1,5 +1,6 @@
 package io.finch
 
+import com.twitter.util.Local
 import scala.annotation.tailrec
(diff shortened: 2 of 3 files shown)


sergeykolbasov:
👍 

Let's merge it and release. We have instant cases that will be solved using it

rpless:
LGTM. I'm 👍 to getting this into the next release too.

vkostyukov:
@sergeykolbasov I'm debating myself on whether there is a need for a bootstrap option. Having both (bootstrap option and `Trace.capture`) seems to be redundant for me. I will run some benchmarks to see how expensive that local lookup (based on what we've seen in FInagle it could be quite expensive).

vkostyukov:
Nothing stands out of the benchmark runs (just a drop in the ocean). Also, it's pretty volatile so somehow the new code is faster than old. Again, I think it's just noise, nothing to be concerned about.

```
BEFORE:

[info] Benchmark                                                   Mode  Cnt      Score      Error   Units
[info] TextBootstrapBenchmark.foos                                thrpt   10  68824.182 ± 2454.004   ops/s
[info] TextBootstrapBenchmark.foos:·gc.alloc.rate.norm            thrpt   10  47888.021 ±    0.008    B/op

AFTER:

[info] Benchmark                                                   Mode  Cnt      Score      Error   Units
[info] TextBootstrapBenchmark.foos                                thrpt   10  70165.207 ± 2442.619   ops/s
[info] TextBootstrapBenchmark.foos:·gc.alloc.rate.norm            thrpt   10  47904.100 ±    0.308    B/op
```

sergeykolbasov:
@vkostyukov ok, that's nice. Have no questions

PR #962, opened 2018-07-13
fixes #955 adding client recipe to cookbook

### LINKED ISSUE #955: A Cookbook recipe of using a Finagle client
This has been asked several times and we should have a cookbook example already.

erikwj:


docs/src/main/tut/cookbook.md
```diff
@@ -351,9 +351,33 @@ val policy: Cors.Policy = Cors.Policy(
...
+import com.twitter.finagle.http._
+import com.twitter.util.{Await, Future}
+
+val host = "api.nasa.gov"
+val client: Service[Request, Response] = Http.client.withTls(host).newService(s"$host:443")
+val request = Request(Method.Get, "/planetary/apod?api_key=NNKOjkoul8n1CH18TWA9gwngW1s1SmjESPjNoUFo")
+val response: Future[Response] = client(request)
+
+Await.result(response.onSuccess { rep: Response =>
+  //println("GET success: " + rep.contentString)
```
rpless: We might want to remove this commented out line.

rpless:
@erikwj Thanks for putting this together! I left a couple comments. Curious to hear your thoughts.

docs/src/main/tut/cookbook.md
```diff
@@ -351,9 +351,33 @@ val policy: Cors.Policy = Cors.Policy(
...
+documentation to get a full understanding of the configuration options for the client like its [loadbalancing][loadbalancing] abilities. A simple example is given below.
+
+```tut:silent
+import com.twitter.finagle.Http
+import com.twitter.finagle.http._
+import com.twitter.util.{Await, Future}
+
+val host = "api.nasa.gov"
+val client: Service[Request, Response] = Http.client.withTls(host).newService(s"$host:443")
+val request = Request(Method.Get, "/planetary/apod?api_key=NNKOjkoul8n1CH18TWA9gwngW1s1SmjESPjNoUFo")
```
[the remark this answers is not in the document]
vkostyukov: Can we use `DEMO_KEY` instead? This https://api.nasa.gov/planetary/apod?api_key=DEMO_KEY works for me. They have some limits about the `DEMO_KEY` usage but I think we should be good: https://api.nasa.gov/api.html#demo_key-rate-limits

vkostyukov:
Thanks, @erikwj! This LGTM (after @rpless's concern about the API key is addressed).

docs/src/main/tut/cookbook.md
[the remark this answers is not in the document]
rpless: :+1: to using the demo key.

docs/src/main/tut/cookbook.md
```diff
@@ -351,9 +351,33 @@ val policy: Cors.Policy = Cors.Policy(
 val corsService: Service[Request, Response] = new Cors.HttpFilter(policy).andThen(service)
 ```
 
+### Creating a Finagle Client
```
vkostyukov: Can you also add a TOC item for this? I don't think we generate automatically.

docs/src/main/tut/cookbook.md
erikwj: @rpless Agree on all above will update

docs/src/main/tut/cookbook.md
erikwj: will do

Issue #954, opened 2018-06-21
Rerunnable -> Arrow/Task


vkostyukov:
As discussed in #926, we want to try replacing `Rerrunable` with `Arrow`/`Task` (https://github.com/traneio/arrows).

This is an internal API change and should not affect users.

vkostyukov:
This is implemented in #964. We'll release both Finch with Rerunnable and Finch with Arrows in 0.23 (under different artifacts but same version).

PR #964, opened 2018-07-13
Rerunnable => Arrows / Task

erikwj:
A first attempt to solve #954, learning Arrow/Task on the side. Code compiles but tests fail with `java.lang.NoSuchMethodError: com.twitter.util.Local$.save()[Lscala/Option;`

Comments welcome


core/src/main/scala/io/finch/Endpoint.scala
```diff
@@ -82,7 +84,7 @@ trait Endpoint[A] { self =>
 
       final def apply(input: Input): Endpoint.Result[B] = self(input) match {
         case EndpointResult.Matched(rem, trc, out) =>
-          EndpointResult.Matched(rem, trc, out.flatMapF(this))
+          EndpointResult.Matched(rem, trc, out.run.flatMap(this))
```
vkostyukov: Maybe we shouldn't run it until the very last step (in `ToService`). I'd imagine, we can wrap user-defined function `A => Future[B]` wit `Task` and use just a normal `Task.flatMap`.

Also, I don't understand how it compiles. `out.run` returns `Future`, which is then flat-mapped onto a function. Not sure from where we get a new `Task` to pass into `EndpointMatched`.

vkostyukov:
Really cool!  @erikwj 

I bet you see a failure b/c we recently changed the API for Future Locals in Finagle but Arrows project hasn't updated yet to the most recent version of util.

core/src/main/scala/io/finch/Endpoint.scala
vkostyukov: Oh I think the conversion happens implicitly: https://github.com/traneio/arrows/blob/master/arrows-twitter/src/main/scala/arrows/twitter/Arrow.scala#L619

We should probably avoid this bouncing and stay in a `Task` as long as possible.

erikwj:
Thanks for your comments @vkostyukov I'll continue with it next week

core/src/main/scala/io/finch/Endpoint.scala
erikwj: did some changes to remove the runs, it compiles and the implicit conversions are gone. Will try tomorrow to fix the test

core/src/main/scala/io/finch/Endpoint.scala
```diff
@@ -82,7 +84,7 @@ trait Endpoint[A] { self =>
 
       final def apply(input: Input): Endpoint.Result[B] = self(input) match {
         case EndpointResult.Matched(rem, trc, out) =>
-          EndpointResult.Matched(rem, trc, out.flatMapF(this))
+            EndpointResult.Matched(rem, trc, out.map(o => o.traverse(fn)).flatMap(f => Task.async(f)))
```
vkostyukov: I wonder if we can keep `flatMap(this)` in place for the sake of saving allocations.

If we modify `apply` as follows, maybe it should work:

```scala
final def apply(oa: Output[A]): Task[Output[B]] = Task.async(oa.traverse(fn))
... 

// and then 
out.flatMap(this)
```

core/src/main/scala/io/finch/Endpoint.scala
```diff
@@ -105,14 +107,15 @@ trait Endpoint[A] { self =>
 
       final def apply(input: Input): Endpoint.Result[B] = self(input) match {
         case EndpointResult.Matched(rem, trc, out) =>
-          EndpointResult.Matched(rem, trc, out.flatMapF(this))
+            EndpointResult.Matched(rem, trc, out.flatMap(o => Task.async(fn(o.value))))
```
vkostyukov: Same q about keeping `flatMap(this)` in place.

core/src/main/scala/io/finch/Endpoint.scala
```diff
@@ -176,11 +179,13 @@ trait Endpoint[A] { self =>
...
-            EndpointResult.Matched(
-              b.rem,
-              a.trc.concat(b.trc),
-              a.out.liftToTry.product(b.out.liftToTry).flatMapF(this)
-            )
+            {
+              EndpointResult.Matched(
+                b.rem,
+                a.trc.concat(b.trc),
+                a.out.liftToTry.join(b.out.liftToTry).map( this).flatMap(f => Task.async(f))
```
vkostyukov: Can we move `Task.async` wrapping under `def apply` so we can keep `flatMap(this)` in place?

core/src/main/scala/io/finch/Endpoint.scala
erikwj: I'll give it a go later today

erikwj: I think I found a solution closest to your suggestion, couldn't edit the `final def apply(oa: Output[A]): Task[Output[B]]` because it should remain returning a `Future[_]` to not break the interface (or am I missing something)

erikwj:
Updating to twitter-util 18.7.0 helped a lot

core/src/main/scala/io/finch/Endpoint.scala
vkostyukov: You can change the interface - it's only used to reduce closure allocations. It's a trick I learned from Flavio (Arrows author) and he called it "merge interfaces". Here is an example:

```scala
class Foo extends AtomicInteger with (Try[Int] => Unit) {
  def apply(ti: Try[Int]): Unit = ti match {
    case Return(i) => set(i) // AtomicInteger.set
    case Throw(_) => 
  }
}

val i: Future[Int] = ???
i.respond(atomicInteger)
```

So instead of allocating an extra closure for `i.respond { case Return(_) => }` and updating a given atomic integer we just use a single object (a smash of both, a closure and an atomic integer) as both things. This helps to save quite some allocations.

erikwj: Thanks for the pointer, it worked

core/src/main/scala/io/finch/Endpoint.scala
```diff
@@ -1,11 +1,11 @@
 package io.finch
 
+import arrows.twitter.{Arrow, Task}
```
vkostyukov: Arrow is unused?

vkostyukov:
Hey @erikwj! This looks pretty decent already! What's failing?

I looked at the CI job and it seems like scalastyle isn't super happy. You can validate the style (and everything else) locally with `sbt validate`.

build.sbt
```diff
@@ -1,17 +1,21 @@
 import microsites.ExtraMdFileConfig
 
+
 lazy val buildSettings = Seq(
   organization := "com.github.finagle",
   version := "0.22.0",
   scalaVersion := "2.12.6",
+  scalafixSemanticdbVersion := "4.0.0-M1",
```
vkostyukov: I'd assume we only needed semanicdb/scalafix for migration purposes? If so let's remove these before merging.

build.sbt
```diff
@@ -33,9 +37,14 @@ lazy val compilerOptions = Seq(
   "-Ywarn-dead-code",
   "-Ywarn-numeric-widen",
   "-Xfuture",
-  "-Xlint"
+  "-Xlint",
+  "-Xplugin-require:semanticdb"
 )
 
+
+scalafixSettings
```
vkostyukov: Same q. about scalafix.

project/plugins.sbt
```diff
@@ -13,4 +13,4 @@ addSbtPlugin("com.jsuereth" % "sbt-pgp" % "1.1.1")
 addSbtPlugin("org.scalastyle" % "scalastyle-sbt-plugin" % "1.0.0")
 addSbtPlugin("pl.project13.scala" % "sbt-jmh" % "0.3.4")
 addSbtPlugin("com.47deg"  % "sbt-microsites" % "0.7.20")
-
+addSbtPlugin("ch.epfl.scala" % "sbt-scalafix" % "0.5.10")
```
vkostyukov: Same q. about scalafix.

erikwj: Yes I will clean up, but I also had to use a local snapshot version of Arrow to use `twitter-util % 18.7.0`

Currently the exception handling is failing: 
- `should give back bad request if we divide by 0 *** FAILED ***`
- `should rescue the exception occurred in it *** FAILED ***`
- `should handle both Error and Errors *** FAILED ***`

erikwj: Need to wrap my head around it, but at least the failures seem consistent

vkostyukov: Oh interesting. Let's clean up everything else and make sure scalastyle is happy. I can take a look at these failures later this week unless you figure out what's going before that.

build.sbt
```diff
@@ -12,6 +13,7 @@ lazy val twitterServerVersion = "18.7.0"
 lazy val circeVersion = "0.9.3"
 lazy val circeIterateeVersion = "0.9.0"
 lazy val catbirdVersion = "18.6.0"
```
erikwj: I see imports are ordered alphabetically, agree on sorting these versions as well? or leave as is?

vkostyukov: Don't have a strong opinion - feel free to sort if you feel like it.

build.sbt
```diff
@@ -49,7 +52,8 @@ val baseSettings = Seq(
     "org.typelevel" %% "cats-core" % catsVersion,
     "com.twitter" %% "finagle-http" % finagleVersion,
     scalaOrganization.value % "scala-reflect" % scalaVersion.value,
-    "io.catbird" %% "catbird-util" % catbirdVersion
+    "io.catbird" %% "catbird-util" % catbirdVersion,
```
vkostyukov: I'm trying to figure out if we still need a catbird dependency? I'd assume `finch-iteratee` would need it but it seems to depend on `iteratee-twitter` already. What happens if we drop it?

erikwj: yes I had the same idea, will see how it works, also removing added whitespaces

core/src/main/scala/io/finch/Endpoint.scala
```diff
@@ -132,7 +132,7 @@ trait Endpoint[A] { self =>
     new Endpoint[B] {
       final def apply(input: Input): Endpoint.Result[B] = self(input) match {
         case EndpointResult.Matched(rem, trc, out) =>
-          EndpointResult.Matched(rem, trc, Rerunnable.fromFuture(fn(out.run)))
+          EndpointResult.Matched(rem, trc, out.flatMap(a => Task.async(fn(Future.value(a)))))
```
erikwj: this seems to be the cause

erikwj:
Thanks @vkostyukov for your support, learned a lot along the way.

core/src/main/scala/io/finch/Endpoint.scala
vkostyukov: Oh cool. Yeah, maybe it's fine to bounce thorough Future/Scheduler here as it's pretty close to what we're doing right now.

vkostyukov:
No, thank you @erikwj! I think the most exciting part is to run some benchmarks and see how `Task` compares to `Returnable`. Do you think you can try it?

I'd suggest to run out wrk benchmark as a first step to make sure the big picture still looks good. After that we can narrow it down to more specific workloads: there is a couple of `Bootstrap*` benchamrks we can run as well.

erikwj:
Yes, I'll give it a try, but probably on Friday

vkostyukov:
That sounds good to me. By all means, there is no rush on this.

fwbrasil:
LGTM. The new release is available: http://search.maven.org/#artifactdetails%7Cio.trane%7Carrows-twitter_2.11%7C0.1.21%7Cjar

travisbrown:
Very simple benchmark results here (in a WIP PR for the catbird instances for arrows): https://github.com/travisbrown/catbird/pull/70

vkostyukov:
That looks very promising @travisbrown!

vkostyukov:
Here is a sneak peek into the product (`::`) benchmark results (the difference is pretty bizarre):

```
BEFORE:

[info] Benchmark                                                   Mode  Cnt     Score     Error   Units
[info] ProductBenchmark.bothMatched                                avgt   10   984.884 ±  35.849   ns/op
[info] ProductBenchmark.bothMatched:·gc.alloc.rate.norm            avgt   10  1472.000 ±   0.001    B/op
[info] ProductBenchmark.leftMatched                                avgt   10    27.908 ±   1.950   ns/op
[info] ProductBenchmark.leftMatched:·gc.alloc.rate.norm            avgt   10    80.000 ±   0.001    B/op
[info] ProductBenchmark.rightMatched                               avgt   10     5.059 ±   0.179   ns/op
[info] ProductBenchmark.rightMatched:·gc.alloc.rate.norm           avgt   10    ≈ 10⁻⁶              B/op


AFTER

[info] ProductBenchmark.bothMatched                                avgt   10   361.867 ±   8.955   ns/op
[info] ProductBenchmark.bothMatched:·gc.alloc.rate.norm            avgt   10   712.000 ±  12.749    B/op
[info] ProductBenchmark.leftMatched                                avgt   10    29.446 ±   2.453   ns/op
[info] ProductBenchmark.leftMatched:·gc.alloc.rate.norm            avgt   10    80.000 ±   0.001    B/op
[info] ProductBenchmark.rightMatched                               avgt   10     5.152 ±   0.326   ns/op
[info] ProductBenchmark.rightMatched:·gc.alloc.rate.norm           avgt   10    ≈ 10⁻⁶              B/op
```

vkostyukov:
Here is the coproduct (`:+:`) benchmark:

```
BEFORE:

[info] Benchmark                                                      Mode  Cnt     Score     Error   Units
[info] CoproductBenchmark.bothMatched                                 avgt   10   242.465 ±  11.592   ns/op
[info] CoproductBenchmark.bothMatched:·gc.alloc.rate.norm             avgt   10   424.000 ±   0.001    B/op
[info] CoproductBenchmark.leftMatched                                 avgt   10   209.721 ±  15.328   ns/op
[info] CoproductBenchmark.leftMatched:·gc.alloc.rate.norm             avgt   10   344.000 ±   0.001    B/op
[info] CoproductBenchmark.rightMatched                                avgt   10   204.683 ±   6.273   ns/op
[info] CoproductBenchmark.rightMatched:·gc.alloc.rate.norm            avgt   10   336.000 ±  12.749    B/op

AFTER:

[info] CoproductBenchmark.bothMatched                                 avgt   10    83.733 ±   1.724   ns/op
[info] CoproductBenchmark.bothMatched:·gc.alloc.rate.norm             avgt   10   272.000 ±   0.001    B/op
[info] CoproductBenchmark.leftMatched                                 avgt   10    49.858 ±   1.122   ns/op
[info] CoproductBenchmark.leftMatched:·gc.alloc.rate.norm             avgt   10   168.000 ±   0.001    B/op
[info] CoproductBenchmark.rightMatched                                avgt   10    50.779 ±   1.250   ns/op
[info] CoproductBenchmark.rightMatched:·gc.alloc.rate.norm            avgt   10   168.000 ±   0.001    B/op
```

erikwj:
That's pretty impressive

vkostyukov:
I changed the base branch from `master` to `arrows` as we want to release this a part of its own artifact: `finch-arrows-x`.

Issue #963, opened 2018-07-13
Add docs for Trace


vkostyukov:
We can pretty much move release notes under the user guide.

vkostyukov:
Done in #965.

PR #965, opened 2018-07-13
Add docs for Trace

vkostyukov:
Addresses #963.


sergeykolbasov:
Awesome.
I think it'd be useful to add some filter example into cookbook also

Issue #958, opened 2018-07-05
Dispatching to a decoder based on Content-Type header


vkostyukov:
This came out of [this Gitter discussion](https://gitter.im/finagle/finch?at=5b3e7ce0bd92d80782971b48).

As of now, Finch's `body*` endpoints don't really care about input's content-type as it's only being used to lookup decoders at compile time. It would be awesome if we can employ the fact we can resolve multiple decoders at compile time, and dispatch input to them later, when endpoint is being run.

Something along the lines:

```scala
body[Application.Json :+: Text.Plain :+: CNil] <- dispatches to a JSON or text decoder based on the input's content type
```

See #905 for an inspiration.

vkostyukov:
That's good with me, @sergeykolbasov!

I want to aim to ship this in the next release. It /seems/ possible given we already did a very similar trick on the outbound path.

vkostyukov:
Implemented in #966.

PR #966, opened 2018-07-14
Add support for content-type-based decoding of request body

sergeykolbasov:
https://github.com/finagle/finch/issues/958

- Add new abstraction `AdaptableDecode` that is used only for `body` and all the related endpoints. This abstraction enables coproduct support for `Content-Type`:
```scala
body[A, Application.Json :+: Application.Csv :+: CNil]
```
Backward-compatibility is there with a very small overhead of one more instance per call.

- Add specs and related documentation

### DIFF
--- a/core/src/main/scala/io/finch/Decode.scala
+++ b/core/src/main/scala/io/finch/Decode.scala
@@ -10,14 +10,14 @@ import shapeless.{:+:, CNil, Coproduct, Witness}
  * an arbitrary type `A`.
  */
 trait Decode[A] {
-  type ContentType
+  type ContentType <: String
 
   def apply(b: Buf, cs: Charset): Try[A]
 }
 
 object Decode {
 
-  type Aux[A, CT] = Decode[A] { type ContentType = CT }
+  type Aux[A, CT <: String] = Decode[A] { type ContentType = CT }
 
   type Json[A] = Aux[A, Application.Json]
   type Text[A] = Aux[A, Text.Plain]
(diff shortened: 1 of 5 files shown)


sergeykolbasov:
@vkostyukov I'm not sure about calling headers (Content-Type) each time, but taking in account that we already consume the whole body of a request, I think it should be relatively fast.

core/src/main/scala/io/finch/Decode.scala
```diff
@@ -40,3 +41,36 @@ object Decode {
    */
   @inline def apply[A, CT <: String](implicit d: Aux[A, CT]): Aux[A, CT] = d
 }
+
+trait AdaptableDecode[A, CT] {
```
vkostyukov: Maybe we can move this under `Decode` companion object, by analogy with `ToResponse.Negotiable`?

Also, what do you think about renaming this to `Decode.Dispatchable`?

core/src/main/scala/io/finch/Decode.scala
```diff
@@ -40,3 +41,36 @@ object Decode {
...
+}
+
+object AdaptableDecode {
+
+  def instance[A, CT](fn: Option[String] => Decode.Aux[A, CT]):  AdaptableDecode[A, CT] = new AdaptableDecode[A, CT] {
+    def apply(contentType: Option[String]): Decode.Aux[A, CT] = fn(contentType)
+  }
+
+  implicit def mkCnil[A, CTH <: String](implicit decode: Decode.Aux[A, CTH]): AdaptableDecode[A, CTH :+: CNil] =
+    instance(_ => decode.asInstanceOf[Decode.Aux[A, CTH :+: CNil]])
```
vkostyukov: I like that this preserves the behavior, but maybe in future, we'd need to change this and make sure to check if the given content-type matches.

vkostyukov: Same thing about `mkSingle`.

core/src/main/scala/io/finch/endpoint/body.scala
```diff
@@ -125,15 +125,15 @@ private[finch] trait Bodies {
    * (non-streamed) requests.
    */
   def bodyOption[A, CT <: String](implicit
```
vkostyukov: We should probably remove `<: String` here too.

core/src/main/scala/io/finch/Decode.scala
vkostyukov: Thinking out loud. I sort of like that we had `CT <: String` requirement on the `Decode` type-class. Perhaps, we can work it around and keep it in place for the sake of symmetry with `Encode`?

One thing we can do is to have a new type-class wrapping `Decode` (not returning it). Similar to how `ToResponse` and `Encode` relation is modeled.

Quick sketch. Let me know if you like it.

```scala
object Decode {
  
  trait Dispatchable[A, CT] {
    def apply(ct: Option[String], b: Buf, cs: Charset): Try[A]
  }
}
```

sergeykolbasov: Sounds good

sergeykolbasov:
@vkostyukov there is a good article about compiler profiling:
https://www.scala-lang.org/blog/2018/06/04/scalac-profiling.html

Probably, we could set up something on Travis CI, but it should be different issue already

core/src/main/scala/io/finch/endpoint/body.scala
```diff
@@ -55,33 +55,33 @@ private object FullBody {
   }
 }
 
-private abstract class Body[A, B, CT <: String](
-  d: Decode.Aux[A, CT],
-  ct: ClassTag[A]
+private abstract class Body[A, B, CT](
+                                       ad: Decode.Dispatchable[A, CT],
```
vkostyukov: What happened to formatting here?

core/src/main/scala/io/finch/Decode.scala
```diff
@@ -39,4 +40,40 @@ object Decode {
...
+      instance((contentType, b, cs) => {
+        if (contentType.exists(ct => ct.equalsIgnoreCase(witness.value))) {
+          decode(b, cs)
+        } else {
+          tail(contentType, b, cs)
+        }
+      })
+
+    implicit def mkSingle[A, CT <: String](implicit
+                                           decode: Decode.Aux[A, CT]
```
vkostyukov: Formatting seems off.

core/src/main/scala/io/finch/Decode.scala
```diff
@@ -39,4 +40,40 @@ object Decode {
...
+
+    def instance[A, CT](fn: (Option[String], Buf, Charset) => Try[A]):  Dispatchable[A, CT] = new Dispatchable[A, CT] {
+      def apply(contentType: Option[String], b: Buf, cs: Charset): Try[A] = fn(contentType, b, cs)
+    }
+
+    implicit def mkCnil[A, CTH <: String](implicit decode: Decode.Aux[A, CTH]): Dispatchable[A, CTH :+: CNil] =
+      instance((_, b, cs) => decode(b, cs))
+
+    implicit def mkCons[A, CTH <: String, CTT <: Coproduct](implicit
+                                                            decode: Decode.Aux[A, CTH],
```
vkostyukov: Same q about formatting.

rpless:
Selecting the last decoder when the coproduct doesn't contain a Content-Type is interesting. I'm not sure this is the behavior I would expect. Part of me feels like the endpoint should not get matched in that case. Reading through the [spec](https://tools.ietf.org/html/rfc7231#section-3.1.1.5) though seems to indicate that this behavior is fine. I think we'll need to make sure we document this clearly in the User Guide.
Other than that, :+1: from me.

vkostyukov:
Yeah, I somewhat agree with @rpless that perhaps we should adjust this behavior in the future and either return 404 (not matched) or 400 (not parsed). We, however, may need to keep the current logic in place and do not inspect content-type before applying a decoder when there is only one decoder in the list (when we reach the last).  I filed #967 so we can discuss this.

PR #969, opened 2018-07-25
Aggregate refined

rider-yi:
I just noticed that the new `finch-refined` module wasn't being published.

```
[warn] 	::::::::::::::::::::::::::::::::::::::::::::::
[warn] 	::          UNRESOLVED DEPENDENCIES         ::
[warn] 	::::::::::::::::::::::::::::::::::::::::::::::
[warn] 	:: com.github.finagle#finch-refined_2.12;0.22.0: not found
[warn] 	::::::::::::::::::::::::::::::::::::::::::::::
```

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -187,7 +187,7 @@ lazy val finch = project.in(file("."))
     "io.spray" %%  "spray-json" % sprayVersion
   ))
   .aggregate(
-    core, iteratee, generic, argonaut, json4s, circe, sprayjson, benchmarks, test, jsonTest, examples, sse
+    core, iteratee, generic, argonaut, json4s, circe, sprayjson, benchmarks, test, jsonTest, examples, sse, refined
   )
   .dependsOn(core, iteratee, generic, circe)
 
--- a/refined/src/main/scala/io/finch/refined/package.scala
+++ b/refined/src/main/scala/io/finch/refined/package.scala
@@ -9,7 +9,7 @@ package object refined {
     ad: DecodePath[A],
     v: Validate[A, B],
     rt: RefType[F]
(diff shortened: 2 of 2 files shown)


vkostyukov:
Oh that's a good catch. Thanks @rider-yi!

CI fails with the following error on 2.11:

```
/home/travis/build/finagle/finch/refined/src/main/scala/io/finch/refined/package.scala:12:88: value toOption is not a member of Either[String,F[A,B]]
[error]   ): DecodePath[F[A, B]] = DecodePath.instance(s => ad(s).flatMap(p => rt.refine[B](p).toOption))
```

Do you have any idea why? I restarted the job just in case.

rpless:
@vkostyukov I think its Scala 2.11 v 2.12 related. Pretty sure in Scala 2.11 `Either` doesn't have `toOption` in Scala 2.11. Cats has an implicit syntax for it, we'll either have to use that or manually fold the `Either` here.

vkostyukov:
I see. Sounds like we never compiled it for 2.11 hence never saw an error. Importing Cats' syntax works for me, fwiw. @rider-yi any chance you want to fix this as part of this PR?

rider-yi:
I just fixed the error with [RightProjection](https://www.scala-lang.org/api/2.12.x/scala/util/Either$$RightProjection.html). It seems that importing Cats's syntax causes unused warning:
```
[warn] /Users/yu_ishikawa/finch/refined/src/main/scala/io/finch/refined/package.scala:3:27: Unused import
[warn] import cats.syntax.either._
[warn]                           ^
```

vkostyukov:
Changes look good to me! Thanks again, @rider-yi!

vkostyukov:
Cherry-picked this into `arrows`: 3ffa8808e7d12fea4efc4f1b49582d20c42794fb and bd92d61a7bc5aa22b6d7f2463b3dd4411bca0c4a.

PR #972, opened 2018-08-01
Remove ToTwitterFuture trait and replace it with cats.arrow.FunctionK

sergeykolbasov:
cats library provides sufficient for our need abstraction out of the box to map `F[A] => G[A]`.
Remove ToTwitterFuture as obsolete and replace it with type alias for the sake of compatibility


vkostyukov:
How does the instance for Scala Future work/look? I don't see changes for it.

sergeykolbasov:
@vkostyukov What's the difference? Of what I see that's going to be the same anonymous instance of `FunctionK`. Taking in account that it's `val` in the end, there is almost no additional price to pay.

vkostyukov:
I'd assume we won't actually need `ToTwitterFuture` should go with cats-effect? Maybe it's not really worth improving on them then?

sergeykolbasov:
Still need it once we convert `F[_]` to `Future` in `ToService`

PR #973, opened 2018-08-09
Update to 18.8.0 for Finagle, etc.

travisbrown:
These are out!


vkostyukov:
Awesome! Thanks a lot @travisbrown!

Issue #926, opened 2018-03-07
Adopt cats-effect type-classes?


vkostyukov:
I figured we should consider this as a way to abstract over concrete IO type for `Endpoint`.

As @travisbrown mentioned, there is no way to provide law-full instances of effect type-classes for Twitter Futures. Maybe doing so for `Rerunnable` is a fine middle-ground?

sergeykolbasov:
I wonder how it'd affect performance

vkostyukov:
We could also try https://github.com/traneio/arrows as a drop-in replacement for `Rerunnable`.

travisbrown:
@vkostyukov I like this idea. I'm planning to add a catbird-arrows module with Cats and cats-effect instances for `Arrow` / `Task`, and I think it might make sense for Finch to replace the catbird dependency with arrows, and users who want Cats instances can pull in catbird themselves.

vkostyukov:
@sergeykolbasov I was hesitant about doing this before 1.0 but maybe I'm being too conservative and it's just a perfect time to do that today. How about we tried to talk to our adopters and see how they feel about `Endpoint[F[_], A]`?

sergeykolbasov:
@vkostyukov I can land a PR with `experimental` API being under the separate subproject. In that case, there will be no need to keep 100% backward-compatibility.

vkostyukov:
@sergeykolbasov I'm not sure we absolutely have to keep the API unchanged. It's one of those changes when the net positive impact in a long run may be worth breaking an API a bit.

I'm wondering if we can do this work in a separate branch for now and publish it under a separate artifact - so we give people time to opt-in (should they want to).

sergeykolbasov:
In that case I already started an experiment quite a while and could land a PR soon.

vkostyukov:
As of now I'm thinking of several ways to deliver cats-effect integration to Finch's users:

 - Release separate artifacts in 0.25 (similar to finch-arrows) and wait several releases to allow people to migrate before switching to cats-effect indefinitely
 - Switch to cats-effect in 0.25 and never look back (users would need to migrate with the upgrade)

I'd be curious to hear what people think our course of action should be. Perhaps I missed an opportunity for a middle-ground solution if so, please, let me know. I'd assume we would be able to move a slider between very aggressive (1-2 cross-published releases before switching to cats-effect) and very conservative (6-12 cross-published releases) should we pick the first approach.

sergeykolbasov:
[...]

PR #979, opened 2018-08-22
Switch to effect-agnostic Endpoint

sergeykolbasov:
This is quite huge PR related to #926 -, though I'm going to explain every big change, and there are a lot of them:

- `Endpoint[A]` now is `Endpoint[F[_], A]` where `F` should have `Effect` type class
- `EndpointResult[A]` now is `EndpointResult[F[_], A]`
- All the built-in endpoints such as `path`, `header`, `param` and so on were reworked to respect `F[_]` type of `Endpoint`
- Same with syntax endpoints to match methods: `get`, `post` etc
- Same for iteratee streams and all related endpoints
- Instead of `io.finch._` customers are supposed to use `io.finch.%effectName%._`. By default there is the `catsEffect` object provided

More in PR


core/src/main/scala/io/finch/Endpoint.scala
```diff
@@ -238,127 +239,136 @@ trait Endpoint[A] { self =>
...
-   * Recovers from any exception occurred in this endpoint by creating a new endpoint that will
-   * handle any matching throwable from the underlying future.
-   */
-  final def rescue[B >: A](pf: PartialFunction[Throwable, Future[Output[B]]]): Endpoint[B] =
-    transform(foa => foa.rescue(pf))
+    * Recovers from any exception occurred in this endpoint by creating a new endpoint that will
+    * handle any matching throwable from the underlying future.
+    */
+  final def rescue(pf: PartialFunction[Throwable, F[Output[A]]]): Endpoint[F, A] =
+    transform(foa => foa.handleErrorWith(pf))
```
sergeykolbasov: Due to the fact that `F[_]` is invariant and `MonadError.handleErrorWith` doesn't give any lower bounds, we can't use `Output[B]` where `B >: A` anymore. That's quite sad

core/src/main/scala/io/finch/Endpoint.scala
```diff
@@ -368,21 +378,21 @@ trait Endpoint[A] { self =>
...
+    final def apply(input: Input): Result[F, A] = EndpointResult.NotMatched
   }
 
   /**
    * Creates an [[Endpoint]] that always matches and returns a given value (evaluated eagerly).
    */
-  def const[A](a: A): Endpoint[A] = new Endpoint[A] {
-    final def apply(input: Input): Result[A] =
-      EndpointResult.Matched(input, Trace.empty, Rerunnable.const(Output.payload(a)))
+  def const[F[_] : Effect, A](a: A): Endpoint[F, A] = new Endpoint[F, A] {
```
sergeykolbasov: That necessity of putting effect type together with value type is striking me a bit, so I was thinking if it's possible to provide syntax like `Endpoint[F].const(a)`, so `A` will be derived automatically.

Ideally, both types here are inferred automatically if there is only one implicit `Effect` in scope, but it's a bit unclear behavior.

core/src/main/scala/io/finch/Endpoint.scala
```diff
@@ -449,42 +459,48 @@ object Endpoint {
...
      */
-    def orElse[B >: A](alternative: => Option[B]): Endpoint[Option[B]] =
+    def orElse[B >: A](alternative: => Option[B]): Endpoint[F, Option[B]] =
       self.map(_.orElse(alternative))
   }
 
-  implicit val endpointInstance: Alternative[Endpoint] = new Alternative[Endpoint] {
-    final override def ap[A, B](ff: Endpoint[A => B])(fa: Endpoint[A]): Endpoint[B] =
-      ff.productWith(fa)((f, a) => f(a))
+  implicit def endpointInstances[F[_] : Effect]: Alternative[({type T[B] = Endpoint[F, B]})#T] = {
```
sergeykolbasov: type lambda allows to hack around lack of partial type application

core/src/main/scala/io/finch/EndpointResult.scala
```diff
@@ -60,69 +46,49 @@ sealed abstract class EndpointResult[+A] {
...
-   * Awaits for an [[Output]] wrapped with [[Try]] (indicating if the [[com.twitter.util.Future]]
-   * is failed).
-   *
-   * @note This method is blocking. Never use it in production.
-   *
-   * @return `Some(output)` if this endpoint was matched on a given input, `None` otherwise.
-   */
-  final def awaitOutput(d: Duration = Duration.Top): Option[Try[Output[A]]] = this match {
-    case EndpointResult.Matched(_, _, out) => Some(Await.result(out.liftToTry.run, d))
+  def awaitOutput(d: Duration = Duration.Top)(implicit e: Effect[F]): Option[Try[Output[A]]] = this match {
```
sergeykolbasov: Ideally, I'd like to move those methods to the test scope and use them implicitly.

core/src/main/scala/io/finch/EndpointResult.scala
```diff
@@ -60,69 +46,49 @@ sealed abstract class EndpointResult[+A] {
...
     def isMatched: Boolean = false
   }
 
-  object NotMatched extends NotMatched {
-    final case class MethodNotAllowed(allowed: List[Method]) extends NotMatched
+  object NotMatched extends NotMatched[Pure] {
+    final case class MethodNotAllowed[F[_]](allowed: List[Method]) extends NotMatched[F]
   }
+
+  implicit def covaryEndpointResult[F[_], A](result: EndpointResult[Pure, A]): EndpointResult[F, A] =
```
sergeykolbasov: Anything with `Pure` effect (all notmatched) could be converted the same thing with `F` effect, `F` is not supposed to happen anyway. It's used in pattern matching inside of `Endpoint`

core/src/main/scala/io/finch/endpoint/effect/Bodies.scala
```diff
@@ -0,0 +1,81 @@
+package io.finch.endpoint.effect
+
+import com.twitter.concurrent.AsyncStream
+import com.twitter.io.Buf
+import io.finch._
+import scala.reflect.ClassTag
+
+trait Bodies[F[_]] { self: EffectInstances[F] =>
```
sergeykolbasov: I had to copypaste all the methods once again into `effect` subpackage because from a customer perspective writing something like `body[Rerunnable, A]` for every endpoint is too much, so when `io.finch.%effectName%._` is imported, only all those copy-pasted endpoints are available with `F` already defined by `%effectName%`

sergeykolbasov:
I think it'd be better to split it into different commits making a review easier.

vkostyukov:
@sergeykolbasov Whoa this is huge! Both in terms of a footprint and the impact!

I'm surely glad seeing us making the first steps and excited to review this!

I, however, might be a little slow on reviews as I got flu yesterday but I will start leaving my feedback as I feel any better.

Let's keep it as a giant review for now and work through the big picture first. I'm personally fine if we just merge that as a big PR - this should go into its own branch anyway (not master).

core/src/main/scala/io/finch/Endpoint.scala
```diff
@@ -54,108 +55,108 @@ import shapeless.ops.hlist.Tupler
  *   Http.server.serve(foobar.toService)
  * }}}
  */
-trait Endpoint[A] { self =>
+abstract class Endpoint[F[_], A](implicit E: Effect[F]) { self =>
```
vkostyukov: I'm interested in hearing your reasoning for this approach vs. threading `Effect[F]` into `map`, `product`, etc?

I've seen many libraries doing the later. For example, both Monix's [Iterant](https://github.com/monix/monix/blob/master/monix-tail/shared/src/main/scala/monix/tail/Iterant.scala) and FS2's [Stream](https://github.com/functional-streams-for-scala/fs2/blob/series/1.0/core/shared/src/main/scala/fs2/Stream.scala) (and many more) just require an implicit effect for their combinators.

core/src/main/scala/io/finch/Endpoint.scala
```diff
@@ -238,127 +239,136 @@ trait Endpoint[A] { self =>
...
+  final def liftToTry: Endpoint[F, Try[A]] = attempt.map({
+    case Right(r) => Return(r)
+    case Left(t) => Throw(t)
+  })
+
+  /**
+    * Lifts this endpoint into one that always succeeds, with [[Either[Throwable, A]] representing both success and
+    * failure cases.
+    */
+  final def attempt: Endpoint[F, Either[Throwable, A]] =
```
vkostyukov: Side note: I support the idea of renaming this. I think we can even do that outside of this PR.

core/src/main/scala/io/finch/Endpoint.scala
```diff
@@ -368,21 +378,21 @@ trait Endpoint[A] { self =>
...
 
-  type Result[A] = EndpointResult[A]
+  type Result[F[_], A] = EndpointResult[F, A]
 
   /**
    * Creates an empty [[Endpoint]] (an endpoint that never matches) for a given type.
    */
-  def empty[A]: Endpoint[A] = new Endpoint[A] {
-    final def apply(input: Input): Result[A] = EndpointResult.NotMatched
+  def empty[F[_] : Effect, A]: Endpoint[F, A] = new Endpoint[F, A] {
```
vkostyukov: I liked how Monix's `Iterant` provides [its builders](https://github.com/monix/monix/blob/master/monix-tail/shared/src/main/scala/monix/tail/Iterant.scala#L1998) to enable the following syntax:

```
Endpoint[IO].empty[Foo]
```

Can we adopt this here?

core/src/main/scala/io/finch/Endpoint.scala
vkostyukov: Oh yeah, we should do that. See my comment above about Iterant's builders.

core/src/main/scala/io/finch/EndpointResult.scala
```diff
@@ -60,69 +46,49 @@ sealed abstract class EndpointResult[+A] {
...
-   *
-   * @note This method is blocking. Never use it in production.
-   *
-   * @return `Some(output)` if this endpoint was matched on a given input, `None` otherwise.
-   */
-  final def awaitOutput(d: Duration = Duration.Top): Option[Try[Output[A]]] = this match {
-    case EndpointResult.Matched(_, _, out) => Some(Await.result(out.liftToTry.run, d))
+  def awaitOutput(d: Duration = Duration.Top)(implicit e: Effect[F]): Option[Try[Output[A]]] = this match {
+    case EndpointResult.Matched(_, _, out) =>
+      val promise = Promise[Try[Output[A]]]
```
vkostyukov: Is there a way to avoid bouncing through `Promise` here? I think `unsafeRunTimed` already does what we want.

```scala
Effect[F].toIO(out).unsafeRunTimed(durection).get
```

core/src/main/scala/io/finch/EndpointResult.scala
vkostyukov: I actually like we have them. It makes experimenting with REPL a lot easier. I'd, perhaps, argue we should keep them in core.

core/src/main/scala/io/finch/endpoint/Paths.scala
```diff
@@ -0,0 +1,87 @@
+package io.finch.endpoint
+
+import cats.effect.Effect
+import io.finch._
+import io.netty.handler.codec.http.QueryStringDecoder
+import scala.reflect.ClassTag
+import shapeless.HNil
+
+private[finch] class Paths[F[_] : Effect] {
```
vkostyukov: I'm not yet sold on this new way of providing endpoint instances. I think there are good arguments could be made for both approaches.

```
// iteratee-style (what's currently implemented)
import io.finch.arrows._
val b: Endpoint[Task, Foo] = jsonBody[Foo]

// fs2 (and others) style
val b = jsonBody[Task, Foo]
```

I feel like adopting the former somewhat defeats the purpose of `cats-effect` as we'd need to take dependencies on different effect implementations to provide people with `io.finch.yourFancyEffect` With just `jsonBody[Task, Foo]` we would not need to do that.

An alternative (and a middle ground) solution would be to allow people create their own modules given the `Effect` instance.

```
val f = io.Finch[Task]
import f._
val b = jsonBody[Foo]
```

I want to spend some time to look around and see what other libraries are doing before I can lean towards any of these approaches.

core/src/main/scala/io/finch/ToService.scala
```diff
@@ -91,12 +93,13 @@ object ToService {
       }
   }
 
-  implicit def hlistTS[A, EH <: Endpoint[A], ET <: HList, CTH, CTT <: HList](implicit
+  implicit def hlistTS[F[_], A, EH <: Endpoint[F, A], ET <: HList, CTH, CTT <: HList](implicit
     ntrA: ToResponse.Negotiable[A, CTH],
     ntrE: ToResponse.Negotiable[Exception, CTH],
+    effect: Effect[F],
```
vkostyukov: I'm thinking we should/might? require `ConcurrentEffect` as Finagle can cancel its promises/futures so we can propagate that back to the users' code. This would definitely shrink the space of effective implementation we can adapt so we should be really careful with our reasoning here.

core/src/main/scala/io/finch/ToService.scala
```diff
@@ -107,9 +110,16 @@ object ToService {
 
             val accept =
               if (opts.negotiateContentType) req.accept.map(a => Accept.fromString(a)) else Nil
-            out.map(oa =>
-              conformHttp(oa.toResponse(ntrA(accept), ntrE(accept)), req.version, opts)
-            ).run
+
+            val promise = Promise[Output[A]]
+            val io = effect.runAsync(out) {
```
vkostyukov: Can we just do `effect.unsafeRunAsync` instead and avoid calling `unsafeRunSync` later?

core/src/main/scala/io/finch/ToService.scala
```diff
@@ -107,9 +110,16 @@ object ToService {
...
+              case Left(t) => IO.pure(promise.setException(t))
+              case Right(value) => IO.pure(promise.setValue(value))
+            }
+            io.unsafeRunSync()
+            promise.map(oa =>
+              conformHttp(oa.convertToResponse(ntrA(accept).apply, ntrE(accept).apply), req.version, opts)
```
vkostyukov: I think we should move this map continuation under `Effect.unsafeRun*` and avoid bouncing through Future.

core/src/main/scala/io/finch/Endpoint.scala
sergeykolbasov: Yes, I like the idea

core/src/main/scala/io/finch/endpoint/Paths.scala
sergeykolbasov: There is no need for us to provide `io.finch.yourFancyEffect` as people might implement their own `object yourFancyEffect` using same approach as https://github.com/finagle/finch/pull/979/files#diff-5c1246ec310c1d425749036da9b6bb61
A lot of traits would be simplified to something like `abstract class Finch[F[_] : Effect]`

And they will get all the endpoints for free

core/src/main/scala/io/finch/ToService.scala
sergeykolbasov: Sadly no, `unsafeRunAsync` is a method of `IO` only

core/src/main/scala/io/finch/Endpoint.scala
sergeykolbasov: I've done it for the sake of hack & try it quickly. Going to do it per combinator, yes

core/src/main/scala/io/finch/endpoint/Paths.scala
sergeykolbasov: @vkostyukov check out `abstract class Finch` in the latest commit plox

sergeykolbasov:
@vkostyukov I've updated to latest cats & cats-effect and fixed some of the previous comments.

core/src/main/scala/io/finch/ToService.scala
sergeykolbasov: I see the reason but from the top of my head there are only two async implementations available that satisfy `ConcurrentEffect` laws: `cats.effect.IO` and `monix.Task`. Neither `Rerunnable` nor `Arrow` support it

vkostyukov:
Hey @sergeykolbasov! I'm going to get another look tonight. Any chance you can rebase? Also, looks like there 1.0 release for cats-effect - we should probably update the dependency.

sergeykolbasov:
@vkostyukov sure, I just came back. Going to take care of it this week

core/src/main/scala/io/finch/ToService.scala
```diff
@@ -121,14 +121,16 @@ object ToService {
...
-            effect.toIO(out).unsafeRunAsync {
-              case Left(t) => IO.pure(promise.setException(t))
-              case Right(value) => IO.pure {
-                  promise.setValue(
-                    conformHttp(value.convertToResponse(ntrA(accept).apply, ntrE(accept).apply), req.version, opts)
-                  )
-                }
-            }
+
+            effect.runAsync(out) {
```
sergeykolbasov: I feel it's better to skip redundant conversion to `IO`

core/src/main/scala/io/finch/EndpointResult.scala
```diff
@@ -1,8 +1,9 @@
 package io.finch
 
+import cats.effect.Effect
 import com.twitter.finagle.http.Method
-import com.twitter.util.{Await, Duration, Future, Try}
-import io.catbird.util.Rerunnable
+import com.twitter.util._
+import scala.concurrent.duration.{Duration => ScalaDuration}
```
vkostyukov: It looks like the next item on our agenda should be migrate off Twitter's `Duration` to Scala `Duration`.

core/src/main/scala/io/finch/Endpoint.scala
```diff
@@ -54,108 +55,111 @@ import shapeless.ops.hlist.Tupler
...
+  def apply(input: Input): EndpointResult[F, A]
 
   /**
-   * Maps this endpoint to the given function `A => B`.
-   */
-  final def map[B](fn: A => B): Endpoint[B] =
-    mapAsync(fn.andThen(Future.value))
+    * Maps this endpoint to the given function `A => B`.
+    */
+  final def map[B](fn: A => B)(implicit F: Monad[F]): Endpoint[F, B] =
```
vkostyukov: I see we require deferent instances depending on what we do. I'd assume your reasoning was to require the least-powerful abstraction suitable for a job. I like this in general, but maybe we should just fix everything into `Effect` and stop worrying about this?

I will try to see what others doing, but if you have some insights on why having different effects could be helpful, please share.

core/src/main/scala/io/finch/Endpoint.scala
```diff
@@ -238,127 +244,146 @@ trait Endpoint[A] { self =>
...
   /**
     * Lifts this endpoint into one that always succeeds, with [[Try]] representing both success and
     * failure cases.
     */
-  final def liftToTry: Endpoint[Try[A]] =
-    new Endpoint[Try[A]] with (Try[Output[A]] => Output[Try[A]]) {
-      final def apply(toa: Try[Output[A]]): Output[Try[A]] = toa match {
-        case Return(oo) => oo.map(Return.apply)
-        case t @ Throw(_) => Output.payload(t.cast[A])
+  final def liftToTry(implicit F: MonadError[F, Throwable]): Endpoint[F, Try[A]] = attempt.map({
```
vkostyukov: We should deprecate this and point people to `attempt`.

This may or may not happen as part of this PR - whatever you feel appropriate.

core/src/main/scala/io/finch/Endpoint.scala
```diff
@@ -368,79 +393,84 @@ trait Endpoint[A] { self =>
...
-  type Result[A] = EndpointResult[A]
+  type Result[F[_], A] = EndpointResult[F, A]
 
-  /**
-   * Creates an empty [[Endpoint]] (an endpoint that never matches) for a given type.
-   */
-  def empty[A]: Endpoint[A] = new Endpoint[A] {
-    final def apply(input: Input): Result[A] = EndpointResult.NotMatched
-  }
+  def apply[F[_]]: EndpointBuilder[F] = new EndpointBuilder[F]
```
vkostyukov: Love this!

core/src/main/scala/io/finch/ToService.scala
vkostyukov: I feel it would be great if we could still support cancellable endpoints. Maybe there is way to distinguish those (cancellable vs. non-cancelable) in `Bootstrap`?

Again, this falls outside of the scope of this PR. I agree with you starting with just simple `Effect` is totally reasonable.

examples/src/main/scala/io/finch/wrk/Finch.scala
```diff
@@ -17,5 +19,5 @@ import io.finch.circe._
  *   c = t * n * 1.5
  */
 object Finch extends Wrk {
-  serve(Endpoint.lift(Payload("Hello, World!")).toServiceAs[Application.Json])
+  serve(Endpoint[IO].lift[Finch.Payload](Payload("Hello, World!")).toServiceAs[Application.Json])
```
vkostyukov: I think we can drop `Finch.Payload` type argument. It should be inferred.

core/src/main/scala/io/finch/EndpointResult.scala
sergeykolbasov: Agree, it's confusing

core/src/main/scala/io/finch/Endpoint.scala
sergeykolbasov: I think it's a matter of approach, but i.e. testing & calling some separate methods sometimes doesn't require you to have `Effect` instance, so it's possible to use any valid `F[_]` such as `Try`

core/src/main/scala/io/finch/Endpoint.scala
sergeykolbasov: Let's deprecate it as a separate PR

build.sbt
```diff
@@ -11,11 +11,12 @@ lazy val twitterVersion = "18.8.0"
 lazy val circeVersion = "0.9.3"
 lazy val circeIterateeVersion = "0.9.0"
 lazy val shapelessVersion = "2.3.3"
-lazy val catsVersion = "1.1.0"
+lazy val catsVersion = "1.3.1"
```
vkostyukov: You should probably rebase.

build.sbt
```diff
@@ -45,7 +46,9 @@ val baseSettings = Seq(
     "org.typelevel" %% "cats-core" % catsVersion,
     "com.twitter" %% "finagle-http" % twitterVersion,
     scalaOrganization.value % "scala-reflect" % scalaVersion.value,
-    "io.catbird" %% "catbird-util" % twitterVersion
+    "io.catbird" %% "catbird-util" % twitterVersion,
```
vkostyukov: Do we still need catbird dependencies from finch-core?

circe/src/test/scala/io/finch/circe/test/package.scala
```diff
@@ -0,0 +1,16 @@
+package io.finch.circe
+
+import cats.Comonad
+import com.twitter.util.Try
+
+package object test {
+
+  implicit val comonadEither: Comonad[Try] = new Comonad[Try] {
```
vkostyukov: Can we defined this within `AbstractJsonSpec`? It seems to be a place where it's used.

core/src/main/scala/io/finch/Finch.scala
```diff
@@ -0,0 +1,16 @@
+package io.finch
+
+import cats.effect.Effect
+import io.finch.endpoint.effect.EffectEndpoints
+import io.finch.syntax.effect.EffectEndpointMappers
+
+abstract class Finch[F[_]](implicit val E: Effect[F])extends EffectEndpoints[F]
```
vkostyukov: How about this name it `Module[F]`? I think that's what iteratee is doing.

I'd imagine something like this should work:

```scala
object Main extends App with Module[IO] {
  def foo = param("a")
}
```

core/src/main/scala/io/finch/Finch.scala
```diff
@@ -0,0 +1,16 @@
...
+  with Outputs
+  with ValidationRules
+  with EffectInstances[F] {
+
+  type Endpoint[A] = io.finch.Endpoint[F, A]
```
vkostyukov: Do we need this?

core/src/main/scala/io/finch/EndpointResult.scala
```diff
@@ -60,69 +47,45 @@ sealed abstract class EndpointResult[+A] {
...
     trc: Trace,
-    out: Rerunnable[Output[A]]
-  ) extends EndpointResult[A] {
+    out: F[Output[A]]
+  ) extends EndpointResult[F, A] {
     def isMatched: Boolean = true
   }
 
-  abstract class NotMatched extends EndpointResult[Nothing] {
+  abstract class NotMatched[F[_]] extends EndpointResult[F, Nothing] {
```
vkostyukov: Can we do?

```scala
abstract class NotMatched extends EndpoointResult[Pure, Nothing]
```

Also, is there anything in cats / stdlib we can use instead of defining our own `Pure`?

core/src/main/scala/io/finch/EndpointResult.scala
vkostyukov: Let's create a ticket so we don't forget.

core/src/main/scala/io/finch/Pure.scala
```diff
@@ -0,0 +1,3 @@
+package io.finch
+
+sealed trait Pure[+A]
```
vkostyukov: Can we use cat's `Id`?

core/src/main/scala/io/finch/EndpointResult.scala
vkostyukov: Can we do following instead?

```scala
abstract class NotMatched extends EndpointResult[Pure, Nothing]
```

Also, as I asked bellow, can we use `Id` instead of `Pure`?

core/src/main/scala/io/finch/Finch.scala
vkostyukov: I think we should name this `Module` by analogy with iteratee.

core/src/main/scala/io/finch/EffectInstances.scala
```diff
@@ -0,0 +1,9 @@
+package io.finch
+
+import cats.effect.Effect
+
+trait EffectInstances[F[_]] {
```
vkostyukov: Given we rename `Finch` to `Module`, can we call this `Component[F[_]]`? This way a module would look like:

```scala
abstract class Module[F](implicit effect) extends Component[F[_]] with Endpoints with Boies with ect
```

sergeykolbasov: Honestly, it sounds way too generic for something that just guarantees the `Effect` instance for type `F[_]`

core/src/main/scala/io/finch/Finch.scala
sergeykolbasov: `Module` is 👍

core/src/main/scala/io/finch/Finch.scala
sergeykolbasov: Again, it's just to make life easier for customers. In some cases, I believe it wouldn't even require to change anything besides the `import` line:
```scala
val helloWorld: Endpoint[String] = get("/hello") { Ok("hello world") }
//instead of
val helloWorld: Endpoint[MyFancyEffect, String] = get("/hello") { Ok("hello world") }
```

build.sbt
sergeykolbasov: No, actually we don't

core/src/main/scala/io/finch/EndpointResult.scala
sergeykolbasov: > Can we do following instead? ...

It was my first intention, actually. Sadly, it's impossible due to the poor GADT support in Scala. Once you start to pattern-match `Endpoint.Result[F, A]` against `NotMatched extends EndpointResult[Pure, Nothing]` (and we do it a lot in `Endpoint`), everything turns into compile-errors nightmare and I spent around a week to find a work-around you see.

Other approach would be to write interpeters like [here](https://pchiusano.github.io/2014-05-20/scala-gadts.html), but it turns the nested pattern matching into spagetti.

> Can we use `Id` instead of `Pure`  

It's possible, though it's my extra precaution to avoid surprising implicit conversion of `EndpointResult[Id, ?]` into `EndpointResult[F[_], ?]` in customer codebase if users by some reasons use `Id` as an effect type. May be it's just my paranoia.

core/src/main/scala/io/finch/ToService.scala
vkostyukov: Yeah, that could a one way to solve this. I think it's totally fine if we just focus on `Effect` for now and figure out cancellations later.

core/src/main/scala/io/finch/Endpoint.scala
vkostyukov: That's a good reason to use least powerful typeclass. Let's keep it as is for now. We can always revisit later if we decide that sticking to `Effect` everywhere is a better approach.

core/src/main/scala/io/finch/Endpoint.scala
vkostyukov: I agree.

core/src/main/scala/io/finch/EndpointResult.scala
vkostyukov: Interesting. Let me play with it in a repl.

core/src/main/scala/io/finch/EffectInstances.scala
vkostyukov: Yeah, you're right. How about `ModuleComponent` or even `Module.Component`. I feel like the name should reflect that fact we use it to build a `Module` out of it.

vkostyukov:
Hey @sergeykolbasov! I'm going to take another closer look today. I think I'm convinced `Module[IO]` is a reasonable way for us to adopt cats-effect. In my head something like this reads nicely:

```scala
class Main extends App with io.finch.Module[IO] {

}
```

I hope that works given `Module` has to be a an abstract class.

I want to spend some time to understand the workaround you had to use for `NotMatched`. After that I feel like we can focus on polishing the smaller bits and asking more people to review.

I'm going to cut a 0.24 release today so the new effect endpoints could end up in 0.25.

sergeykolbasov:
@vkostyukov 

Okay, so what I did is to move all the built-in stuff we have under the `EndpointBuilder`.
It's much better now as I removed all the duplicating code for endpoint functions with `F[_]` type and without, but it's still possible to use `Module` approach _and_ builder one. I.e.:

```scala
Endpoint[IO].get(Endpoint[IO].path("foo")) { Ok("bar" }

//or

import io.finch.catsEffect._
get("foo") { Ok("bar") }

//or

object App extends io.finch.Module[IO] {
  get("foo") { Ok("bar") }
}
``` 

Ah, also removed `Pure` thingy.

sergeykolbasov:
Now I'd like to run benchmarks and scratch my head over some small optimizations like having constant `emptyOutput` and stuff

core/src/main/scala/io/finch/ToService.scala
```diff
@@ -118,11 +120,26 @@ object ToService {
...
 
-            val response = out.map(oa =>
-              conformHttp(oa.toResponse(ntrA(accept), ntrE(accept)), req.version, opts)
-            )
-
-            response.run
+            val promise = Promise[Response]
+
+            (effect match {
+              case concurrent: ConcurrentEffect[F] =>
```
vkostyukov: I'm still not super sure we have to do that just now. I'd be happy to roll out an initial version w/o interrupts.

core/src/main/scala/io/finch/package.scala
```diff
@@ -1,12 +1,14 @@
...
 /**
  * This is a root package of the Finch library, which provides an immutable layer of functions and
  * types atop of Finagle for writing lightweight HTTP services.
  */
-package object finch extends Endpoints
-    with Outputs
-    with ValidationRules {
+package object finch {
+
+  object catsEffect extends Module[IO]
```
vkostyukov: Maybe we don't have to provide this?

core/src/main/scala/io/finch/Module.scala
```diff
@@ -0,0 +1,15 @@
+package io.finch
+
+import cats.effect.Effect
+import io.finch.syntax.EndpointMappers
+
+abstract class Module[F[_]](implicit val E: Effect[F]) extends Endpoints[F]
```
vkostyukov: I'm thinking. Maybe we don't have to capture `implicit val E: Effect[F]` here? Monix's builders work this way - only `F[_]` needs to be captured, which means we can resolve the implicit later.

What's interesting is this would allow us to downgrade `Module` to a trait, which gives users the ability to mix it into any existing class should they need to.

This is, obviously, something we can address later, when we move this work under a separate branch.

core/src/main/scala/io/finch/Module.scala
```diff
@@ -0,0 +1,15 @@
...
+  with Outputs
+  with ValidationRules
+  with EffectInstances[F] {
+
+  type Endpoint[A] = io.finch.Endpoint[F, A]
+
+  object syntax extends EndpointMappers[F]
```
vkostyukov: One more thing. I believe we can get rid of `syntax` now - let's just mix it with other things.

core/src/main/scala/io/finch/package.scala
sergeykolbasov: I think having some out-of-the-box example should be useful. `IO` comes for free as a data type inside of `cats-effect` anyway

core/src/main/scala/io/finch/Module.scala
sergeykolbasov: In terms of the performance it should be better. `Effect[F]` is captured once even if it's resolved from `def`

core/src/main/scala/io/finch/ToService.scala
sergeykolbasov: I think it's low-hanging fruit and we can do it right from the beginning

vkostyukov:
> @vkostyukov Sounds good, but I still want to run benchmarks first 😄

I'm personally not too worried about performance just yet. We can track down any regression down the road.

rpless:
> Future can't be expressed in terms of Effect
ugh, right. We'll probably have to provide something (at least in the short term) using `Rerunnable` and `Arrow` then.

sergeykolbasov:
@vkostyukov @rpless
Benchmark results:
https://gist.github.com/sergeykolbasov/a0bbb20777fdf6da371a28ce707a08e3

Looks quite promising 😄

vkostyukov:
@sergeykolbasov Wow looks great so far! 

And the `with Twitter Futures` is Finch in master?

I'm copying the product benchmark results here for better discoverability (the improvement reminds me what we got with Arrows):

Before:

```
[info] ProductBenchmark.bothMatched                                                 avgt   10    1058.658 ±    78.951   ns/op
[info] ProductBenchmark.bothMatched:·gc.alloc.rate.norm                             avgt   10    1472.000 ±     0.001    B/op
[info] ProductBenchmark.leftMatched                                                 avgt   10      30.222 ±     1.718   ns/op
[info] ProductBenchmark.leftMatched:·gc.alloc.rate.norm                             avgt   10      80.000 ±     0.001    B/op
[info] ProductBenchmark.rightMatched                                                avgt   10       5.572 ±     0.408   ns/op
[info] ProductBenchmark.rightMatched:·gc.alloc.rate.norm                            avgt   10      ≈ 10⁻⁶                B/op
```

After:
```
[info] ProductBenchmark.bothMatched                                                 avgt   10     272.480 ±    11.060   ns/op
[info] ProductBenchmark.bothMatched:·gc.alloc.rate.norm                             avgt   10     720.000 ±     0.001    B/op
[info] ProductBenchmark.leftMatched                                                 avgt   10      31.837 ±     0.571   ns/op
[info] ProductBenchmark.leftMatched:·gc.alloc.rate.norm                             avgt   10      80.000 ±     0.001    B/op
[info] ProductBenchmark.rightMatched                                                avgt   10       6.055 ±     0.168   ns/op
[info] ProductBenchmark.rightMatched:·gc.alloc.rate.norm                            avgt   10      ≈ 10⁻⁶                B/op
```

sergeykolbasov:
@vkostyukov yes, it's the master branch I compared with

\*slaps the roof of the PR\*
should I push this bad guy into separate branch?

vkostyukov:
I think we can declare a victory and just squash-and-merge it into an `cats-effect` branch. Unless there is anything else you wanted to clean up - just let me know.

sergeykolbasov:
@vkostyukov done

alexandru:
Folks this is awesome.

Let me know if you need help with anything. You can contact us on the [Gitter channel](http://gitter.im/typelevel/cats-effect), or me in private (usually with a high latency, but I get back).

sergeykolbasov:
@alexandru Sure thing, thanks for the support! 😄

Issue #967, opened 2018-07-15
Determine if we should inspect Content-Type before applying a (single) decoder


vkostyukov:
See #966 for a discussion.

vkostyukov:
I'm giving this a shot.

PR #980, opened 2018-08-25
Support 415

vkostyukov:
This is addressing #967.

The easiest way of supporting 415 (UnsupportedMediaType) is to return a special (singleton) exception from `Decode.Dispatchable[CT, CNil]` (the very last decoder in the list) so we can later match on that within a derived service.

Here is the proposed behavior:

 - By default (a standard `Bootstrap`), a body endpoint that sees an unknown media type will resolve into `Throw(Error.NotParsed)` (fail-fast) with an underlying cause of `Decode.UnsupportedMediaType`.
 - A `Bootstrap` option `enableUnsupportedMediaType` (false by default), would turn those errors into 415 responses. Note the default response remains 400.

I'm not entirely in love with encoding this logic in a special exception but it seems a very natural way to implement it.  I'm eager to hear what people think about this.

As a bonus, this PR fixes a bug with dispatchable decoders where content-type also contains a charset (i.e., `application/json;charset=utf-8`). The new logic only relies on media-type when dispatching to a decoder (as it should).

### DIFF
--- a/core/src/main/scala/io/finch/endpoint/body.scala
+++ b/core/src/main/scala/io/finch/endpoint/body.scala
@@ -13,7 +13,7 @@ import scala.reflect.ClassTag
 private abstract class FullBody[A] extends Endpoint[A] {
 
   protected def missing: Future[Output[A]]
-  protected def present(contentType: Option[String], content: Buf, cs: Charset): Future[Output[A]]
+  protected def present(contentType: String, content: Buf, cs: Charset): Future[Output[A]]
 
   final def apply(input: Input): Endpoint.Result[A] =
     if (input.request.isChunked) EndpointResult.NotMatched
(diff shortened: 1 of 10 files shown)


core/src/main/scala/io/finch/endpoint/body.scala
```diff
@@ -21,7 +21,11 @@ private abstract class FullBody[A] extends Endpoint[A] {
       val output = Rerunnable.fromFuture {
         val contentLength = input.request.contentLengthOrNull
         if (contentLength == null || contentLength == "0") missing
-        else present(input.request.contentType, input.request.content, input.request.charsetOrUtf8)
+        else present(
+          Option(input.request.mediaTypeOrNull),
```
rpless: Given that we always wrap this in an option, maybe we want to push that down into `mediaTypeOrNull` and have that function return an `Option[String]`?

vkostyukov: Yeah, I was thinking about this. Maybe even eliminate the need for an option wrapper and just treat empty strings as none? This would shove some allocations from the request path.

vkostyukov:
@sergeykolbasov Mind taking a quick look? I'm thinking to ship this today.

sergeykolbasov:
Sorry, I'm on vacation.

vkostyukov:
Hey @sergeykolbasov, no worries (also, sorry to nag you). Feel free to take a look when you're back!

sergeykolbasov:
👍

PR #981, opened 2018-09-13
Update Twitter deps and iteratee.io

travisbrown-stripe:
I've just published a new batch of releases for 18.9.0 and cats-effect 1.0.0.


vkostyukov:
Awesome! Thank you, Travis!

Issue #196, opened 2015-02-16
Code formatter?


vkostyukov:
There is a code formatter for Scala: https://github.com/mdr/scalariform. We might think about enabling it.

rpless:
I'm not sure we should do this. Code formatters are not that flexible and will require a lot of tuning of its options to get it to a place we are comfortable with. I enabled it on the core project locally and it does some interesting things. For example it renders:
```scala
  def flatMap[S, B](fn: A => PRequestReader[S, B])(
    implicit ev: R %> S
  ): PRequestReader[R, B] = new PRequestReader[R, B] {
```
as
```scala
  def flatMap[S, B](fn: A => PRequestReader[S, B])(
    implicit
    ev: R %> S
  ): PRequestReader[R, B] = new PRequestReader[R, B] {
```
I'm imagine there is an option to change this behavior, and I'm going play with it a little to see if I can get it in a more reasonable state, but I'm not really sold on it. I think Wartremover and scalastyle are better options for enforcing format.

vkostyukov:
Yeah. That's a weird style for sure. Maybe we can grab an initial config from some open source project that uses scalariform? That's basically what I did wit scalastyle (grabbed it from cats).

vkostyukov:
I think we should do that now. Everybody seems to be using scalafmt nowadays.

joroKr21:
Scalafmt added in #1267

PR #988, opened 2018-10-02
Refine Endpoint Module

vkostyukov:
This PR refines what `EndpointModule` is and where `Endpoint` instances are constructed. It doesn't invent anything new but reuses ideas from other libraries solving similar problems.

The following ways of constructing endpoints are now supported.

1. If you prefer being explicit:
```scala
import io.finch._, cats.effect.IO
val foo = Endpoint.path[IO, Int] // Endpoint[IO, Int]
```

2. If you prefer being explicit and also would like type-inference work for you:
```scala
import io.finch._, cats.effect.IO
val foo = Endpoint[IO].path[Int] // Endpoint[IO, Int]
```

3. If you prefer do not repeat yourself and just opt-in for Cats Effect in this source file:
```scala
import io.finch._, io.finch.catsEffect._
val foo = path[Int] // Endpoint[IO, Int]
```

4. If you prefer to have an extension point for your own effect.
```scala
object myIO extends Endpoint.Module[MyIO]

import io.finch._, myIO._
val foo = path[Int] // Endpoint[MyIO, Int]
```

5. If you prefer an extension point for your own effect and want to keep the imports minimal.
```scala
import io.finch._

object Foo extends Endpoint.Module[MyIO] {
  val foo = path[Int] // Endpoint[MyIO, Int]
}
```

### What's changed
 - *All* endpoint instances are now defined on `Endpoint.*` (think `Endpoint.param`, `Endpoint.body`) and `EndpointModule.*` just redirects its calls there similar to how iteratee and Monix do that.
 - `Module[F[_]]` no longer embeds `Outputs` and `ValidationRules` are they are not depending on the effect type (`io.finch._` imports them as it does today).
 - There is no longer `tried` module implementation, `IO` is used in tests.
 - The syntax package is removed and merged with the rest of the endpoints (note `Endpoint.Mappable`).
 - We're no longer providing the `Endpoint` type alias from within `EndpointModule`.
 -  `/` is renamed to `zero` and `*` is renamed to `pathAny` (the fewer symbolic APIs we have the better).
 - Style has been broken. I want to keep it that way until we replace scalastyle with scalafmt in #196.
 - Implicit conversions from `Int` and `Boolean` to path-endpoints were removed. Only `String` is supported now (the less we implicitly convert the better).


vkostyukov:
> What is a benifit of having two almost equivalent method definitions like `Endpoint.const` and `EndpointModule.const` comparing to what it was before with `Endpoint[F].const`?

I feel like this structure promote consistency. `Endpoint.*` is the only place where instances are defined and it uses the very direct API (a call-site has to supply both `F[_]` and `A`). And the `EndpointModule` enables users to drop the `F[_]` argument assuming it's always the same thing in their apps.

The very same approach is used in several places where I looked-up. Notable, Monix's `Iterant` and iteatee libraries are structured in the same.

PR #992, opened 2018-10-06
removes duplicate Xss and Xms options

arron-green:
minor modification to `build.sbt` which removes duplicated `javaOptions` from `benchmarks`


vkostyukov:
Thank you, @arron-green!

PR #995, opened 2018-10-06
starts the effort to simplify artifact release

### LINKED ISSUE #977: Simplify Artifacts Release
I'm eager to see if we can adopt what @fwbrasil is doing for https://github.com/traneio/arrows (and perhaps other projects). The idea of being able to trigger a release via `git commit` to `version.sbt` seems quite appealing to me. 

As of now, it takes about 10 mins in total for me to cut a release (no including writing release notes) -that's like 2h/year (12 releases) that I could have spent playing with my dog or, at least, wasting my time in Overwatch.

This commit seems to enable the magic for Trane Arrows: https://github.com/traneio/arrows/commit/310b1862fbe62579af1081107540358a997e4dce

arron-green:
In the spirit of hacktoberfest, this PR is a start to simplify the effort required to release finch artifacts.

@vkostyukov I'm going to need your help with finishing this PR as i'm sure you have the keys to publish these artifacts.

as suggested i used https://www.theguardian.com/info/developer-blog/2014/sep/16/shipping-from-github-to-maven-central-and-s3-using-travis-ci as a reference

Requires the following files to be added to the `./build` directory

| File  | Description |
| ------------- | ------------- |
| deploy_key.pem.enc | Encrypted github deploy key to trigger commits, when non PR branches are merged into the `master` branch when `version.sbt` does not contain `SNAPSHOT` |
| secring.gpg.enc | Encrypted `gpg` secret ring |
| pubring.gpg.enc | Encrypted `gpg` public ring |
| credentials.sbt.enc | Encrypted sbt file containing credentials to sonatype |

```scala
// example credentials.sbt
import com.typesafe.sbt.SbtPgp._

credentials += Credentials(
  "Sonatype Nexus Repository Manager",
  "oss.sonatype.org",
  "<SONATYPE_USERNAME>",
  "<SONATYPE_PASSWORD>"
)

pgpPassphrase := Some("<PGP_PASSPHRASE>".toCharArray)
```

I also wasn't sure on what to add for the github commits for releasing so i figured we could take advantage of environment variables set through travis. I defined `CI_DEPLOY_USERNAME` and `CI_DEPLOY_EMAIL`

Also worth noting that the matrix style building of `2.11` and `2.12` might not make sense with this PR.  I don't think the `sbt release` command used in the `build.sh` makes sense if it kicks off two separate travis jobs. Therefore i left the execution of `build.sh` commented out for now

We might need to iterate on this a few times to find the right balance of auto release vs semi-automatic release

Fixes #977


vkostyukov:
Thank, @arron-green! I will take a look into this today.

arron-green:
@vkostyukov no problem. glad to help on next steps to find the right balance in what you're looking for :)

build.sbt
```diff
@@ -64,6 +65,33 @@ val baseSettings = Seq(
...
+    import scala.io.Source
+    import java.io.PrintWriter
+
+    // files containing version to update upon release
+    val filesToUpdate = Seq(
+      "docs/src/main/tut/index.md"
+    )
+    val pattern = """"com.github.finagle" %% "finch-.*" % "(.*)"""".r
+
+    filesToUpdate.foreach{ fileName =>
```
vkostyukov: This needs a space after `foreach`

build/build.sh
```diff
@@ -0,0 +1,36 @@
...
+    openssl aes-256-cbc -pass env:ENCRYPTION_PASSWORD -in ./build/pubring.gpg.enc -out local.pubring.gpg -d
+    openssl aes-256-cbc -pass env:ENCRYPTION_PASSWORD -in ./build/credentials.sbt.enc -out local.credentials.sbt -d
+    openssl aes-256-cbc -pass env:ENCRYPTION_PASSWORD -in ./build/deploy_key.pem.enc -out local.deploy_key.pem -d
+
+    if [[ "${TRAVIS_BRANCH}" == "master" && $(cat version.sbt) != *"SNAPSHOT"* ]]; then
+        eval "$(ssh-agent -s)"
+        chmod 600 local.deploy_key.pem
+        ssh-add local.deploy_key.pem
+        git config --global user.name "${CI_DEPLOY_USERNAME}"
+        git config --global user.email "${CI_DEPLOY_EMAIL}"
```
vkostyukov: What are those `CI_DEPLOY` values by default? Should we set them into something?

build/build.sh
```diff
@@ -0,0 +1,36 @@
...
+SBT_CMD="sbt ++${TRAVIS_SCALA_VERSION} validate"
+
+if [[ "${TRAVIS_PULL_REQUEST}" == "false" ]]; then
+    SBT_CMD+=" +coverageOff +publish"
+    openssl aes-256-cbc -pass env:ENCRYPTION_PASSWORD -in ./build/secring.gpg.enc -out local.secring.gpg -d
+    openssl aes-256-cbc -pass env:ENCRYPTION_PASSWORD -in ./build/pubring.gpg.enc -out local.pubring.gpg -d
+    openssl aes-256-cbc -pass env:ENCRYPTION_PASSWORD -in ./build/credentials.sbt.enc -out local.credentials.sbt -d
+    openssl aes-256-cbc -pass env:ENCRYPTION_PASSWORD -in ./build/deploy_key.pem.enc -out local.deploy_key.pem -d
+
+    if [[ "${TRAVIS_BRANCH}" == "master" && $(cat version.sbt) != *"SNAPSHOT"* ]]; then
```
vkostyukov: What does commit the new version back? Or maybe we aren't doing it yet?

.travis.yml
```diff
@@ -14,5 +17,6 @@ cache:
     - $HOME/.sbt/boot/
 script:
   - sbt ++$TRAVIS_SCALA_VERSION validate
+# - "./build/build.sh"
```
vkostyukov: Maybe we should just give up on built matrix and start using `build.sh`? It seems to be what Flavio is doing for Arrows (https://github.com/traneio/arrows/blob/master/.travis.yml).

build/build.sh
vkostyukov: Perhaps I should somehow use the same email/user name while generating a github deploy key: https://help.github.com/articles/generating-a-new-ssh-key-and-adding-it-to-the-ssh-agent/#generating-a-new-ssh-key

vkostyukov: As we don't really have a Finch domain yet, I made an email alias from my personal domain: `ci@kostyukov.net`. Let's just use it for now as a hard-coded value.

arron-green: Okay cool. How about `Finch CI` for the username so it's easy to distinguish automated CI deploys vs deploys made by you?

arron-green: borrowing ideas from https://github.com/traneio/arrows/blob/master/build/build.sh#L21

vkostyukov: Yep. `Finch CI` sounds good!

build/build.sh
arron-green: The magic seems to be part of the `sbt-release` functionality defined here
https://github.com/sbt/sbt-release/blob/v1.0.9/src/main/scala/ReleaseExtra.scala#L77-L93

Used as part of the release process defined here https://github.com/finagle/finch/pull/995/files#diff-fdc3abdfd754eeb24090dbd90aeec2ceR131

TBH, i haven't tested the release process yet.  I have an internal jfrog account where i can test with a sandbox repo. Maybe i can create a dummy repo to test out the versioning. I also need to read up a bit more on `sbt-release` and understand how the next version gets promoted.  In theory it should be `0.26.0-SNAPSHOT` when a new release is performed.  Maybe setting to `1.0.0-SNAPSHOT` is a manual step before committing to the `1.0.0` milestone

Wasn't super familiar with this plugin but seems like it can do a bit more than what is currently being done.  Seems like it can apparently automate release notes?

.travis.yml
arron-green: makes sense to me!

arron-green: @fwbrasil i see you went through a couple of different iterations on doing cross builds to finally hardcoding `2.12` in your build script. Any helpful gotchas that you discovered in the process?

build.sbt
```diff
@@ -75,6 +103,10 @@ lazy val publishSettings = Seq(
       Some("releases"  at nexus + "service/local/staging/deploy/maven2")
   },
   publishArtifact in Test := false,
+  pgpSecretRing := file("local.secring.gpg"),
```
vkostyukov: Should we add `local.*` into `.gitignore`?

vkostyukov: Also, what's interesting is I have gpg 2.2.10 installed locally and it doesn't use the `secring.gpg` file (according to this [doc](https://www.gnupg.org/documentation/manuals/gnupg/GPG-Configuration.html)):

> A secret keyring as used by GnuPG versions before 2.1. It is not used by GnuPG 2.1 and later. I'm not sure which version of gpg utility sbt-gpg depends on so perhaps I should downgrade to make sure I generate both files.

vkostyukov:
I think I did everything right: #999 (although I haven't really checked anything).

build.sbt
vkostyukov: I decided to use the older version of GPG I installed via `brew`. I'm pretty sure that version works just fine with `sbt-gpg` plugin:

```
$ gpg --version
gpg (GnuPG) 1.4.18
Copyright (C) 2014 Free Software Foundation, Inc.
```

PR #997, opened 2018-10-07
[WIP]: Test out scalafmt

rpless:
This is a WIP branch for `scalafmt`. I have not removed `scalastyle` yet because there are a number of cases that it can handle that `scalafmt` does not appear to cover.I tried to pick rules that I thought made sense in terms of how the project has been written thus far, but I'm open to adding/removing or changing any of the settings I've put in the config.

One large source of changes generated by scalafmt is our inconsistent use of doc style. I stuck with scalafmt's default which is `Scaladocs`, but either way that rule in particular generates a lot of changes. Another large source of changes comes from the default `verticalMultiline` rules which I kinda just guessed with `6`, so we may want to go higher with that. One last thing to note is that scalafmt greatly increases the size of the Hlist instance of `ToService` to the point where it started to fail the scalastyle function size limit. So I bumped that up a little.

I would be curious to here others' thoughts on scalafmt and if we should keep scalastyle around.

### DIFF
--- a/argonaut/src/main/scala/io/finch/argonaut/Decoders.scala
+++ b/argonaut/src/main/scala/io/finch/argonaut/Decoders.scala
@@ -9,13 +9,13 @@ import io.finch.internal.HttpContent
 trait Decoders {
 
   /**
-   * Maps Argonaut's [[DecodeJson]] to Finch's [[Decode]].
-   */
+    * Maps Argonaut's [[DecodeJson]] to Finch's [[Decode]].
+    */
   implicit def decodeArgonaut[A](implicit d: DecodeJson[A]): Decode.Json[A] =
     Decode.json { (b, cs) =>
       Parse.parse(b.asString(cs)).flatMap(_.as[A].result.leftMap(_._1)) match {
         case Right(result) => Right(result)
-        case Left(error) => Left(new Exception(error))
+        case Left(error)   => Left(new Exception(error))
       }
-  }
+    }
 }
--- a/core/src/main/scala/io/finch/Endpoint.scala
+++ b/core/src/main/scala/io/finch/Endpoint.scala
@@ -6,12 +6,7 @@ import cats.effect.{Effect, Sync}
 import cats.syntax.all._
 import com.twitter.concurrent.AsyncStream
 import com.twitter.finagle.Service
-import com.twitter.finagle.http.{
-  Cookie => FinagleCookie,
-  Method => FinagleMethod,
-  Request,
-  Response
-}
+import com.twitter.finagle.http.{Cookie => FinagleCookie, Method => FinagleMethod, Request, Response}
 import com.twitter.finagle.http.exp.{Multipart => FinagleMultipart}
 import com.twitter.io.Buf
 import io.finch.endpoint._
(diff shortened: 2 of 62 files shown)


rpless:
`scalafmt` doesn't seem to like `Endpoint`. I'm getting an "error" from that file in that scalafmt prints an error, but appears to format everything correctly. There was a closed [issue](https://github.com/scalameta/scalafmt/issues/1123) on `scalafmt`.

vkostyukov:
Whoa this is huge! Thanks a lot @rpless for looking into this. I will review the changes today (specifically the places you mentioned). I personally don't mind the big diff and feel like it's worth doing anyway as long as we're happy to the output.

Quick question: would `sbt scalafmt` format the project for us?

rpless:
@vkostyukov Yes `sbt scalafmt` and `sbt scalafmtSbt` will format the codebase. I've also added `scalafmtCheck` and `scalafmtSbtCheck` to the validate command which will tell us if the code has been properly formatted. When I did that, I realized the problem I was seeing if `Endpoint` was actually a problem and the code was not fully being formatted. It was suggested that I try the 1.6.0-RC4, which fixes the issue, but has changed the default style slightly (and thus changed the formatting of the project again). There are a couple of odd ones now, but overall I think the new defaults are an improvement on what was in 1.5.1.

argonaut/src/main/scala/io/finch/argonaut/Decoders.scala
```diff
@@ -10,13 +10,13 @@ import io.finch.internal.HttpContent
 trait Decoders {
 
   /**
```
vkostyukov: I wonder what happened to my (and others) IJ so it generates these wrong scaladoc comments. Let me see if I can find where it's configurable.

vkostyukov:
@rpless Sorry I messed your PR (merge conflicts). I'm thinking to merge `cats-effect` into master so we can just continue working against master as we normally do. Would you mind rebasing this against master when I'm done?

vkostyukov:
@rpless, the `cats-effect` branch is merged into master. Let me know when you rebase and I will take a look.

core/src/main/scala/io/finch/Endpoint.scala
```diff
@@ -271,39 +277,45 @@ trait Endpoint[F[_], A] { self =>
...
+    ): Service[Request, Response] = Bootstrap.serve[CT](this).toService
 
   /**
     * Recovers from any exception occurred in this endpoint by creating a new endpoint that will
     * handle any matching throwable from the underlying future.
     */
-  final def rescue(pf: PartialFunction[Throwable, F[Output[A]]])(implicit
-    F: ApplicativeError[F, Throwable]
-  ): Endpoint[F, A] = transform(foa => foa.handleErrorWith(pf))
+  final def rescue(
```
vkostyukov: Is it just me or this is now much harder to grok?

rpless: yeah I agree. I suspect this is related to `verticalMultiline`. I can try shutting it off and see how it looks.

rpless:
Our options for the doccstrings are `ScalaDoc`, `JavaDoc`, and `preserve`. I think the last option just leaves them as they are.

vkostyukov:
> Our options for the doccstrings are `ScalaDoc`, `JavaDoc`, and `preserve`. I think the last option just leaves them as they are.

Maybe we can start with `preserve` for now and roll from there?

Issue #989, opened 2018-10-02
Get rid of Twitter Try in the APIs


vkostyukov:
As we're not polymorphic over the effect type, lets also drop the `Try` types in the API (use Scala's `Either` instead).

prayagupd:
I see couple of usage of `twitter.Try` - https://github.com/finagle/finch/search?l=Scala&q=Try

With scala equivalent `scala.util.Try` or `Either[Throwable, A]`, I like `Either[E, A]` instead of `Either[Throwable, A]` but it might be lot of refactor that way.

vkostyukov:
For `Decode`, for example, we could probably use `Either[Error, A]`.

Also, try looking at `cats-effect` branch - there where we need to get rid of `Try`s (as that's the twitter-util-free branch).

vkostyukov:
Dine in #998.

PR #998, opened 2018-10-07
replace twitter.Try[A] with scala.Either[Throwable, A]

prayagupd:
This is a WIP for https://github.com/finagle/finch/issues/989

- [x]  replaced all occurances of `twitter.Try[A]` with `Either[Throwable, A]`
- [x] need to fix `DecodeEntityLaws`, providing view for `cats.laws.IsEq[Either[Throwable,A]]`, 

```scala
[error] /Users/prayagupd/buybest/sc212/finch/core/src/test/scala/io/finch/DecodeEntityLaws.scala:20:32: No implicit view available from cats.laws.IsEq[Either[Throwable,A]] => org.scalacheck.Prop.
[error]     "roundTrip" -> Prop.forAll { (a: A) => roundTrip(a) }
[error] 
```

### DIFF
--- a/core/src/main/scala/io/finch/Decode.scala
+++ b/core/src/main/scala/io/finch/Decode.scala
@@ -1,7 +1,6 @@
 package io.finch
 
 import com.twitter.io.Buf
-import com.twitter.util.{Throw, Try}
 import java.nio.charset.Charset
 import scala.util.control.NoStackTrace
 import shapeless.{:+:, CNil, Coproduct, Witness}
@@ -13,7 +12,7 @@ import shapeless.{:+:, CNil, Coproduct, Witness}
 trait Decode[A] {
   type ContentType <: String
 
-  def apply(b: Buf, cs: Charset): Try[A]
+  def apply(b: Buf, cs: Charset): Either[Throwable, A]
 }
 
 object Decode {
@@ -32,15 +31,15 @@ object Decode {
   /**
    * Creates an instance for a given type.
    */
-  def instance[A, CT <: String](fn: (Buf, Charset) => Try[A]): Aux[A, CT] = new Decode[A] {
+  def instance[A, CT <: String](fn: (Buf, Charset) => Either[Throwable, A]): Aux[A, CT] = new Decode[A] {
     type ContentType = CT
-    def apply(b: Buf, cs: Charset): Try[A] = fn(b, cs)
+    def apply(b: Buf, cs: Charset): Either[Throwable, A] = fn(b, cs)
   }
 
-  def json[A](fn: (Buf, Charset) => Try[A]): Json[A] =
+  def json[A](fn: (Buf, Charset) => Either[Throwable, A]): Json[A] =
     instance[A, Application.Json](fn)
 
-  def text[A](fn: (Buf, Charset) => Try[A]): Text[A] =
+  def text[A](fn: (Buf, Charset) => Either[Throwable, A]): Text[A] =
     instance[A, Text.Plain](fn)
 
   /**
@@ -53,14 +52,14 @@ object Decode {
    * value.
    */
   trait Dispatchable[A, CT] {
-    def apply(ct: String, b: Buf, cs: Charset): Try[A]
+    def apply(ct: String, b: Buf, cs: Charset): Either[Throwable, A]
   }
 
   object Dispatchable {
 
     implicit def cnilToDispatchable[A]: Dispatchable[A, CNil] = new Dispatchable[A, CNil] {
-      def apply(ct: String, b: Buf, cs: Charset): Try[A] =
-        Throw(Decode.UnsupportedMediaTypeException)
+      def apply(ct: String, b: Buf, cs: Charset): Either[Throwable, A] =
+        Left(Decode.UnsupportedMediaTypeException)
     }
 
     implicit def coproductToDispatchable[A, CTH <: String, CTT <: Coproduct](implicit
@@ -68,7 +67,7 @@ object Decode {
       witness: Witness.Aux[CTH],
       tail: Dispatchable[A, CTT]
     ): Dispatchable[A, CTH :+: CTT] = new Dispatchable[A, CTH :+: CTT] {
-      def apply(ct: String, b: Buf, cs: Charset): Try[A] =
+      def apply(ct: String, b: Buf, cs: Charset): Either[Throwable, A] =
         if (ct.equalsIgnoreCase(witness.value)) decode(b, cs)
         else tail(ct, b, cs)
     }
--- a/core/src/main/scala/io/finch/DecodeEntity.scala
+++ b/core/src/main/scala/io/finch/DecodeEntity.scala
@@ -63,5 +63,5 @@ trait LowPriorityDecode {
     gen: Generic.Aux[A, H],
     ev: (E :: HNil) =:= H,
     de: DecodeEntity[E]
-  ): DecodeEntity[A] = instance(s => de(s).map(b => gen.from(b :: HNil)))
+  ): DecodeEntity[A] = instance(s => de(s).right.map(b => gen.from(b :: HNil)))
 }
--- a/core/src/main/scala/io/finch/DecodeEntity.scala
+++ b/core/src/main/scala/io/finch/DecodeEntity.scala
@@ -1,7 +1,8 @@
 package io.finch
 
-import com.twitter.util.{Return, Throw, Try}
 import java.util.UUID
+
+import scala.util.Try
 import shapeless._
 
 /**
@@ -9,7 +10,7 @@ import shapeless._
  * an arbitrary type `A`.
  */
 trait DecodeEntity[A] {
-  def apply(s: String): Try[A]
+  def apply(s: String): Either[Throwable, A]
 }
 
 object DecodeEntity extends HighPriorityDecode {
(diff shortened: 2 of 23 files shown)


core/src/main/scala/io/finch/Decode.scala
```diff
@@ -13,7 +12,7 @@ import shapeless.{:+:, CNil, Coproduct, Witness}
 trait Decode[A] {
   type ContentType <: String
 
-  def apply(b: Buf, cs: Charset): Try[A]
+  def apply(b: Buf, cs: Charset): Either[Throwable, A]
```
vkostyukov: I wonder if we can make it more specific `Either[io.finch.Error, A]`? If it's not easy, no big deal, we can revisit that later.

core/src/main/scala/io/finch/DecodeEntity.scala
```diff
@@ -19,25 +20,25 @@ object DecodeEntity extends HighPriorityDecode {
...
 
-  implicit val decodeString: DecodeEntity[String] = instance(s => Return(s))
+  implicit val decodeString: DecodeEntity[String] = instance(s => Right(s))
 
 }
 
 trait HighPriorityDecode extends LowPriorityDecode {
 
-  implicit val decodeInt: DecodeEntity[Int] = instance(s => Try(s.toInt))
+  implicit val decodeInt: DecodeEntity[Int] = instance(s => Try(s.toInt).toEither)
```
vkostyukov: We would need to use `Either` directly to avoid that extra allocations for `Try`.

core/src/main/scala/io/finch/endpoint/header.scala
```diff
@@ -11,23 +10,23 @@ private[finch] abstract class Header[F[_], G[_], A](name: String)(implicit
...
   }
 
   final def apply(input: Input): EndpointResult[F, G[A]] = {
     val output: F[Output[G[A]]] = F.suspend {
       input.request.headerMap.getOrNull(name) match {
         case null => missing(name)
-        case value => d(value).transform(self) match {
-          case Return(r) => F.pure(r)
-          case Throw(t) => F.raiseError(t)
+        case value => apply(d(value)) match {
```
vkostyukov: Re `apply`: looks like we don't need to mix-in `Function1` anymore. Any chance you want to make it a simple method instead? Obviously, we can revisit that later in a sense it doesn't need to happen as part of this PR.

core/src/main/scala/io/finch/endpoint/param.scala
```diff
@@ -3,31 +3,30 @@ package io.finch.endpoint
...
-import com.twitter.util.{Return, Throw, Try}
 import io.finch._
 import scala.reflect.ClassTag
 
 private[finch] abstract class Param[F[_], G[_], A](name: String)(implicit
   d: DecodeEntity[A],
   tag: ClassTag[A],
   protected val F: Effect[F]
-) extends Endpoint[F, G[A]] with (Try[A] => Try[Output[G[A]]]) { self =>
+) extends Endpoint[F, G[A]] with (Either[Throwable, A] => Either[Throwable, Output[G[A]]]) { self =>
```
vkostyukov: Same question about `with (Either[Throwable, A] => Either[Throwable, Output[G[A]]])`. It seems to me we can get rid of this now.

core/src/main/scala/io/finch/DecodeEntity.scala
prayagupd: Yeah. I used `scala.util.Try` to catch exceptions. If not I might have to use plain old try/catch

```scala
try {
   Right(s.toInt)
} catch {
 case e: Throwable => Left(e)
}
```

vkostyukov: Yeah, that's perfect.

core/src/main/scala/io/finch/Decode.scala
vkostyukov: I'm perfectly happy if we just do `Either[Throwable, A]` for now and revisit this later. So, please, use your own judgment.

core/src/main/scala/io/finch/DecodeEntity.scala
```diff
@@ -19,35 +19,44 @@ object DecodeEntity extends HighPriorityDecode {
...
+  implicit val decodeBoolean: DecodeEntity[Boolean] = instance(s => toEither(s)(_.toBoolean))
 
   implicit val decodeUUID: DecodeEntity[UUID] = instance(s =>
-    if (s.length != 36) Throw(new IllegalArgumentException(s"Too long for UUID: ${s.length}"))
-    else Try(UUID.fromString(s))
+    if (s.length != 36) Left(new IllegalArgumentException(s"Too long for UUID: ${s.length}"))
+    else toEither(s)(UUID.fromString)
   )
+
+  private def toEither[A](s: String)(fn: String => A): Either[Throwable, A] = {
```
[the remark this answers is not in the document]
prayagupd: Yeah makes sense. I kept `try catch` block on each operation. Thank you for the feedback

core/src/main/scala/io/finch/Decode.scala
prayagupd: Okay, I will keep it `Either[Throwable, A]` for now but will look into later.

core/src/test/scala/io/finch/FinchSpec.scala
```diff
@@ -266,7 +266,7 @@ trait FinchSpec extends FlatSpec
       val resultX = await(x(input))
       val resultY = await(y(input))
 
-      Eq[Option[(Input, Try[Output[A]])]].eqv(resultX, resultY)
+      Eq[Option[(Input, Either[Throwable, Output[A]])]].eqv(resultX, resultY)
```
prayagupd: @vkostyukov 

This guy is failing to find an implicit value here, I'm not sure if I can import from somewhere? Please let me know if you have any idea. 

```scala
[error] /Users/prayagupd/buybest/finch/core/src/test/scala/io/finch/FinchSpec.scala:269:9: could not find implicit value for parameter ev: cats.kernel.Eq[Option[(io.finch.Input, Either[Throwable,io.finch.Output[A]])]]
[error]       Eq[Option[(Input, Either[Throwable, Output[A]])]].eqv(resultX, resultY)
[error]         ^
```

vkostyukov: Maybe instance for `Eq[Throwable]` is missing?

vkostyukov: You can probably try defining it under `io.finch.MissingInstances`.

prayagupd: Ahh I see. Let me try that. 

Also, I rebased from cats-effect yesterday with more `twitter.Try` to `Either[Throwable, A]`, just push few minutes ago. Please let me know if you see anything.

vkostyukov: Yeah, once CI is happy I'm happy.

prayagupd: figured out that scala 2.11.x did not have `.map`/ `.flatMap` in it - https://www.scala-lang.org/api/2.11.8/#scala.util.Either, finch still supports `2.11.x` i believe :)

core/src/main/scala/io/finch/EndpointResult.scala
```diff
@@ -48,20 +48,24 @@ sealed abstract class EndpointResult[F[_], +A] {
...
   }
 
   def awaitOutputUnsafe(d: Duration = Duration.Inf)(implicit e: Effect[F]): Option[Output[A]] =
-    awaitOutput(d).map(toa => toa.get)
+    awaitOutput(d).map(toa => toa.right.get)
 
-  def awaitValue(d: Duration = Duration.Inf)(implicit e: Effect[F]): Option[Try[A]] =
-    awaitOutput(d).map(toa => toa.flatMap(oa => Try(oa.value)))
+  def awaitValue(d: Duration = Duration.Inf)(implicit e: Effect[F]): Option[Either[Throwable, A]] =
+    awaitOutput(d).map(toa => toa.flatMap(oa => Right(oa.value)))
```
vkostyukov: This fails on 2.11. You need to select a `right` projection here: `toa.right.map(oa => oa.value)`.

vkostyukov: Also, you can verify a build locally via `sbt +validate`.

core/src/test/scala/io/finch/FinchSpec.scala
vkostyukov: Yeah, or just use `either.right.map`.

core/src/main/scala/io/finch/endpoint/header.scala
```diff
@@ -11,23 +10,23 @@ private[finch] abstract class Header[F[_], G[_], A](name: String)(implicit
...
-) extends Endpoint[F, G[A]] with (Try[A] => Try[Output[G[A]]]) { self =>
+) extends Endpoint[F, G[A]] { self =>
 
   protected def missing(name: String): F[Output[G[A]]]
   protected def present(value: A): G[A]
 
-  final def apply(ta: Try[A]): Try[Output[G[A]]] = ta match {
-    case Return(a) => Return(Output.payload(present(a)))
-    case Throw(e) => Throw(Error.NotParsed(items.HeaderItem(name), tag, e))
+  final def apply(ta: Either[Throwable, A]): Either[Throwable, Output[G[A]]] = ta match {
```
vkostyukov: This seems unused?

core/src/test/scala/io/finch/MissingInstances.scala
```diff
@@ -8,11 +8,18 @@ import com.twitter.util.{Return, Throw, Try}
  * Type class instances for non-Finch types.
  */
 trait MissingInstances {
+
   implicit def eqTry[A](implicit A: Eq[A]): Eq[Try[A]] = Eq.instance {
```
vkostyukov: I don't think we need this instance anymore.

core/src/main/scala/io/finch/endpoint/header.scala
prayagupd: Yep, REMOVED now.

core/src/test/scala/io/finch/MissingInstances.scala
prayagupd: Yeah, my "Fear Driven Development" :)

core/src/main/scala/io/finch/Endpoint.scala
```diff
@@ -14,7 +14,6 @@ import com.twitter.finagle.http.{
 }
 import com.twitter.finagle.http.exp.{Multipart => FinagleMultipart}
 import com.twitter.io.Buf
-import com.twitter.util.{Return, Throw, Try}
```
vkostyukov: Love it!

core/src/test/scala/io/finch/FinchSpec.scala
prayagupd: `.right` looks deprecated in 2.13.x, so i think sticking with pattern match would be nice, 

```scala
scala> data.right
            ^
       warning: method right in class Either is deprecated (since 2.13.0): Either is now right-biased
res1: scala.util.Either.RightProjection[String,Int] = RightProjection(Left(error))
```

core/src/main/scala/io/finch/Endpoint.scala
prayagupd: Thank you for the love @vkostyukov :) . I'm glad to learn from you and contribute to this awesome piece of work. hope to contribute more :)

PR #999, opened 2018-10-09
Add encoded keys/credentials/etc

vkostyukov:
To accomodate #995.


vkostyukov:
> @vkostyukov should we change this PR to merge into the WIP branch in #995 rather than master?

Let's merge these two PRs into the same branch (off of master) and keep iterating there.

PR #1000, opened 2018-10-09
Enable auto releases

### LINKED ISSUE #977: Simplify Artifacts Release

vkostyukov:
@arron-green and I spent some time working on this. We're putting everything together to see how it looks.

Fixes #977.

Follow up work:
 - docs generation #1002

Also 🎉 for the 1000th PR!


vkostyukov:
@arron-green Thanks again for instigating this!

What our next steps should be in your opinion?

.travis.yml
```diff
@@ -1,18 +1,25 @@
...
+- find $HOME/.ivy2 -name "ivydata-*.properties" -delete
+- find $HOME/.sbt -name "*.lock" -delete
 cache:
   directories:
-    - $HOME/.ivy2/cache
-    - $HOME/.sbt/boot/
+  - "$HOME/.ivy2/cache"
+  - "$HOME/.sbt/boot/"
 script:
   - sbt ++$TRAVIS_SCALA_VERSION validate
```
vkostyukov: I'd assume we should just toggle into `build.sh` here and we're done?

build/build.sh
```diff
@@ -0,0 +1,36 @@
+#!/usr/bin/env bash
+set -e
+
+SBT_CMD="sbt ++${TRAVIS_SCALA_VERSION} validate"
```
vkostyukov: I'm not entirely sure what implications it might have to lock into 2.12 here, but I'm willing to try assuming if what people already doing in other projects.

vkostyukov: Or, perhaps, using just `release with-defaults` (drop `cross`) would allow us to keep build matrix?

vkostyukov: What I mean is:

```
sbt ++${TRAVIS_SCALA_VERSION} 'release with-defaults'
```

arron-green: this feels problematic but we can try it.  If `release` command triggers an updates, which bumps the version and commits the next version then i'm not entirely sure how this will effect two builds going at once.  My gut feeling says that two commits will result from this and feels a bit dangerous

.travis.yml
```diff
@@ -1,18 +1,25 @@
...
-sudo: false
+branches:
+  except:
+  - release
+sudo: true
 scala:
-  - 2.11.12
-  - 2.12.6
+- 2.11.12
+- 2.12.6
```
arron-green: Should this be changed to `2.12.7`? `build.sbt` currently points to `2.12.7`

.travis.yml
arron-green: assuming the build matrix stuff is set correctly, then yes

build/build.sh
vkostyukov: Oh I see now. That's a good point. I'm less optimistic now. Maybe an alternative would be to keep build matrix and place for PRs and "skip" the 2.11 build when we're releasing? I'd guess we can just check the string against `{TRAVIS_SCALA_VERSION}`.

.travis.yml
vkostyukov: Oh yeah, we totally should.

vkostyukov: One pro for tossing away a build matrix is we never have to do that again.

build/build.sh
vkostyukov: Anyways, I think we should keep it simple for now and just give up on build matrix. Let me give it a try.

.travis.yml
```diff
@@ -1,18 +1,25 @@
 language: scala
-sudo: false
+branches:
+  except:
+  - release
```
vkostyukov: @arron-green Do you know what's `release` branch is for?

vkostyukov:
I think it's ready to go. The only question I have is how do we test it? Perhaps we could just wait for the next release and see how it goes.

vkostyukov:
I rebase that against `cats-effect` - it will be merged there.

arron-green:
> I think it's ready to go. The only question I have is how do we test it? Perhaps we could just wait for the next release and see how it goes.

Sounds good.  Glad i could help motivate this issue! Excited to see the result of this

PR #1005, opened 2018-10-11
Model NotMatched result more explicitly

vkostyukov:
This PR gets rid of implicit conversion we had to go from `EndpointResult[Id, A]` to `EndpointResult[F, A]`. This type-casting is now happening explicitly either via pattern matching `_: NotMatched[F]` or via `NotMatched.apply[F]` method.


sergeykolbasov:
👍

PR #1008, opened 2018-10-20
attempt to fix broken build

arron-green:
appears that dropping `iteratee-twitter` also dropped the transitive dependency for `catbird`

`catbird` still has a dependency on the previous finagle release so i created https://github.com/travisbrown/catbird/pull/75 to match that version

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -9,6 +9,8 @@ lazy val buildSettings = Seq(
 )
 
 lazy val twitterVersion = "18.10.0"
+lazy val catbirdVersion = "18.9.1"
+//lazy val catbirdVersion = "18.10.0"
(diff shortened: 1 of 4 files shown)


vkostyukov:
Thanks for the PR @arron-green! I think shouldn't need to depend on catbird anymore. My understanding is we needed that for `MonadError` instance for Twitter Try. By simply replacing `c.t.Try` with `scala.util.Try`, we can avoid the catbird dependency.

We do, however, need both catbird and iteratee-twitter for a regular Finch release. I will open a separate PR for that.

arron-green:
@vkostyukov no problem! i'll close the PR then

PR #1012, opened 2018-10-20
Fix for #1009 - bumps minor version on next auto release

### LINKED ISSUE #1009: Bump minor version in auto-release, instead of patch
Looks like we need to refer to these settings: https://github.com/sbt/sbt-release/blob/0d79966dfe65a60ac97d8084691ce396bbaaede7/README.md#custom-versioning

/cc @arron-green

arron-green:
bump minor version for next auto release

fixes #1009

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -118,6 +118,7 @@ lazy val publishSettings = Seq(
       "scm:git:git@github.com:finagle/finch.git"
     )
   ),
+  releaseVersionBump := sbtrelease.Version.Bump.Minor,
   releaseProcess := {
     Seq[ReleaseStep](
       checkSnapshotDependencies,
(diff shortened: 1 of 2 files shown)


arron-green:
@vkostyukov one minor thing i thought about adding to this PR is that intellij complains about the order of `import ReleaseTransformations._` saying it should come first.  
Also that `sbtrelease.ReleasePlugin` is unused

vkostyukov:
Thanks for the PR @arron-green! If you want to fix these issues you mentioned, just add them as separate commits to this branch. Otherwise, I'm happy to merge this as is if you prefer to leave them for later.

vkostyukov:
That's great! Thanks again @arron-green!

PR #1015, opened 2018-10-25
Introduce pathEmpty

vkostyukov:
It does what you think it does.

```scala
import io.finch._, io.finch.catsEffect._

val e = get(pathEmpty) { Ok("foo") }  // matches GET /
```


sergeykolbasov:
👍

Issue #1013, opened 2018-10-23
Mapper conversions for Futures


sergeykolbasov:
I had some thoughts on the easy migration to new `finchx` and one of them is to provide few additional implicit conversions for `Mapper` that will work with `Future` and automatically pack it into `E : Effect`.

So as an example:
```scala
import io.finch._
import io.finch.catsEffect._
import io.finch.futures._

val endpoint: Endpoint[IO, String] = get("hello") { Future.value(Ok("world"))) } //this apply will be automatically lifted into IO and converted to Mapper, so it should work
```

What do you think people? Is there any need for it?

rpless:
I think there is definitely a need for this for both Twitter and Scala Futures (assuming we can support both). Changing to Cats Effect is a big change and not everyone is going to be able to move all of their codebases to it immediately (or even over the 6 months or whatever time frame that we're still going to publish Finch with futures). Easing the migration path will be beneficial for everyone.

vkostyukov:
I like that. I wonder if we can do this w/o exporting `io.finch.futures._` and just provide mapping from both Twitter Future and Scala Future to whatever is the current effect type.

sergeykolbasov:
No, I think `Mapper` is enough atm

sergeykolbasov:
Closed with #1016

PR #1016, opened 2018-10-27
Add syntax support for Scala & Twitter Futures in Mapper

sergeykolbasov:
Related to #1013

### DIFF
--- a/core/src/main/scala/io/finch/internal/Mapper.scala
+++ b/core/src/main/scala/io/finch/internal/Mapper.scala
@@ -47,7 +47,7 @@ private[finch] trait LowPriorityMapperConversions {
       f.onComplete {
         case Success(s) => cb(Right(s))
         case Failure(t) => cb(Left(t))
-      }(dummieExecutionContext)
+      }(DummyExecutionContext)
     }
   }
 
@@ -78,6 +78,7 @@ private[finch] trait LowPriorityMapperConversions {
   /**
     * @group LowPriorityMapper
     */
+  @deprecated("com.twitter.util.Future use is deprecated in Endpoints. Consider to use cats-effect compatible effect", "0.25.0")
   implicit def mapperFromFutureOutputFunction[A, B, F[_] : Effect](f: A => TwitterFuture[Output[B]]): Mapper.Aux[F, A, B] =
     instance(_.mapOutputAsync(a =>
       twitterFutureToEffect(f(a))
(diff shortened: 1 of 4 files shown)


vkostyukov:
I like it @sergeykolbasov! What's your thoughts on how long we have to keep these around? For the record, I don't think  "indefinitely" is unreasonable.

core/src/main/scala/io/finch/internal/package.scala
```diff
@@ -121,4 +122,9 @@ package object internal {
       new String(array, begin, end - begin, cs.name)
     }
   }
+
+  private[internal] val dummieExecutionContext: ExecutionContext = new ExecutionContext {
```
vkostyukov: I'd probably prefer this being its own file under `internal` package (perhaps as an `object`). Also no need to limit its visibility as it has `internal` in its name.

sergeykolbasov:
I think as long as we support the previous endpoint version.
I'm going to add `deprecated` flag to those implicit conversions to throw some warnings during compile phase

core/src/main/scala/io/finch/internal/Mapper.scala
```diff
@@ -106,11 +181,37 @@ object Mapper extends HighPriorityMapperConversions {
...
+  implicit def mapperFromFutureResponseHFunction[F[_] : Effect, A, FN, FR](f: FN)(implicit
+    ftp: FnToProduct.Aux[FN, A => FR],
+    ev: FR <:< TwitterFuture[Response]
+  ): Mapper.Aux[F, A, Response] = instance(_.mapOutputAsync { value =>
+    val fr = twitterFutureToEffect(ev(ftp(f)(value)))
+    fr.map(r => Output.payload(r, r.status))
+  })
+
+  implicit def mapperFromScFutureOutputHFunction[F[_] : Effect, A, B, FN, FOB](f: FN)(implicit
+                                                                                    ftp: FnToProduct.Aux[FN, A => FOB],
```
vkostyukov: Formatting here is somewhat crazy.

core/src/main/scala/io/finch/internal/Mapper.scala
```diff
@@ -106,11 +181,37 @@ object Mapper extends HighPriorityMapperConversions {
...
+                                                                                    ev: FOB <:< ScalaFuture[Output[B]]
+  ): Mapper.Aux[F, A, B] = instance(_.mapOutputAsync(value => scalaFutureToEffect(ev(ftp(f)(value)))))
+
+  implicit def mapperFromScFutureResponseHFunction[F[_] : Effect, A, FN, FR](f: FN)(implicit
+                                                                                  ftp: FnToProduct.Aux[FN, A => FR],
```
vkostyukov: Here as well.

sergeykolbasov: I hope we gonna have scalafmt soon 😞

vkostyukov: Yea that would be rad. I've seen cats recently adopted it too. Sounds like we're going to be a good company after all.

PR #1017, opened 2018-10-30
Introduce Endpoint.resource

vkostyukov:
Turns out after we deprecated Twitter Futures in the mapper it became problematic to serve static files (although people generally not supposed to do that).

This PR leverages cats-effect `Resource` to read Java resources and serve them as a static content. The scaladoc has a example.

I'm not super sure about the name yet, maybe `liftResource` or `staticResource` could be a better alternative. Let me know how do you feel about this.


rpless:
I like this. For the name, maybe `classpathResource` or `resourceFromClasspath` could work? I also like `liftResource`.

My only other thought is maybe we should replace the example of serving static files in the cookbook with this.

vkostyukov:
> I like this. For the name, maybe `classpathResource` or `resourceFromClasspath` could work? I also like `liftResource`.
> 
> My only other thought is maybe we should replace the example of serving static files in the cookbook with this.

Yes, I need to update the cookbook. Also I like your `classpath` idea.

PR #1020, opened 2018-11-02
Refine TODO App example

vkostyukov:
I was playing around the idea of doing better job with our examples to make them more appealing. This PR adds an HTML frontend for a Finch app. It looks like this.

screen shot 2018-11-01 at 9 06 50 pm

Additionally this:

 - Introduces `Endpoint.resource` to serve static content (see #1017)
 - Makes use of `Ref` instead of `synchronized`
 - Adds `EndpointResult.output` as a syntax (it's a bummer variance doesn't give us another option; I will file a ticket - #1021)

Curious to hear what people think about this.

### DIFF
--- a/core/src/main/scala/io/finch/EndpointResult.scala
+++ b/core/src/main/scala/io/finch/EndpointResult.scala
@@ -31,9 +31,6 @@ sealed abstract class EndpointResult[F[_], +A] {
 
   /**
    * Returns the remainder of the [[Input]] after an [[Endpoint]] is matched.
-   *
-   * @return `Some(remainder)` if this endpoint was matched on a given input,
-   *         `None` otherwise.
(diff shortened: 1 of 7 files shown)


sergeykolbasov:
I'd play with fonts, but otherwise, it's good 😄  

👍

core/src/main/scala/io/finch/EndpointModule.scala
```diff
@@ -2,7 +2,7 @@ package io.finch
 
 import cats.Applicative
 import cats.data.NonEmptyList
-import cats.effect.{Effect, Sync}
+import cats.effect.{ContextShift, Effect, Sync}
```
sergeykolbasov: Unused import?

Issue #1019, opened 2018-10-31
Do we want to support `/foo/bar/baz` as a single path?


vkostyukov:
Basically,

```
get("/foo/bar/baz") ==
get("foo" :: "/bar/baz") ==
get("/foo/bar" :: "baz") == 
get("foo" :: "bar" :: "baz")
```

Very simple rule we can employ, if the path starts with a slash `/`, if can contain multiple (zero or more) segments that we need to match against.

This way `get("foo")` is effectively the same as `get("/foo")`. Moreover, we can make `get("/")` being the same as `get(pathEmpty)`.

rpless:
Yeah I'm with @sergeykolbasov on this one. Maybe if we a construct like `compositePath` that would behave this way as opposed to the existing `path` that could work? But I think that still has the issues with `?` and `&`.

vkostyukov:
I'm convinced. Let's cancel this for now. We can reopen should users ask for this.

spockz:
Whoops. It appears that my comment got lost. Jersey and Spring MVC support this feature and I haven’t seen any developers trying to enter query parameters in the path annotations. Just some anecdotal evidence. :)

PR #1034, opened 2018-11-14
Matching multiple segments with path

vkostyukov:
This PR makes it possible to match multiple path segments with a single endpoint, `path`. The change is backward compatible with the current behavior: `path("foo")` or just `"foo"` works as expected.

In addition it enables a couple interesting use cases:
 
 - `get("/")` to match empty path 
 - `get("/foo/bar")` to match multiple segments (I've seen people doing this)
 - this would simplify #1017


vkostyukov:
Cancelled in #1019.

PR #1036, opened 2018-11-14
seq to list with scalafix

### LINKED ISSUE #1035: Use List[String] instead of Seq[String] internally

erikwj:
Fixes #1035

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -1,8 +1,6 @@
 import ReleaseTransformations._
 import microsites.ExtraMdFileConfig
 
-addCompilerPlugin("org.scalameta"  % "semanticdb-scalac" % "4.0.0" cross CrossVersion.full)
-
 
 lazy val buildSettings = Seq(
   organization := "com.github.finagle",
--- a/build.sbt
+++ b/build.sbt
@@ -1,6 +1,9 @@
 import ReleaseTransformations._
 import microsites.ExtraMdFileConfig
 
+addCompilerPlugin("org.scalameta"  % "semanticdb-scalac" % "4.0.0" cross CrossVersion.full)
+
+
 lazy val buildSettings = Seq(
   organization := "com.github.finagle",
   scalaVersion := "2.12.7",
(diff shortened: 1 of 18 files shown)


build.sbt
```diff
@@ -1,6 +1,9 @@
 import ReleaseTransformations._
 import microsites.ExtraMdFileConfig
 
+addCompilerPlugin("org.scalameta"  % "semanticdb-scalac" % "4.0.0" cross CrossVersion.full)
```
vkostyukov: Do we need this enabled going forward?

erikwj: it's needed for scalafix, I like to keep scalafix available, but I'm fine to clean up if you want to

erikwj: This did the trick for 90% of the code: 
```scalafix https://gist.githubusercontent.com/erikwj/1c576d0f78a095dc7ff2e14ebb71c8a9/raw/c139a6b92f9ee3a7b9edfa7dd6856dcff0010afa/SeqToList```

vkostyukov:
@rpless @sergeykolbasov How do you feel about this?

build.sbt
vkostyukov: Right. I just wasn't sure what's the reason to keep it around, until we have another rewrite. I'm voting for omitting it.

vkostyukov: Getting back to this. @erikwj, can we omit this for now (as it's no longer needed)?

rpless:
I'm mostly fine with this, although there are 3 cases I'm not sure about:
- The `paths` method in Endpoint feels like it should be a Seq because there is an order
- The `accept` argument in `Negotiable` also feels like it should remain a sequential
- the `r` in `Trace.fromRoute` also is ordered

I'm kinda on the fence about changing these, but if people are okay with there being an implied order for the lists in these cases, then I can go along with that.

build.sbt
erikwj: ok @vkostyukov will submit the changes

rpless:
> Seq is that it could be substituted with mutable structures

That's a good point. Also didn't realize `List` extended `Seq`, thought it list weren't necessarily ordered, but I guess they are, so its not a problem then.

Issue #1001, opened 2018-10-09
Support to fs2


vkostyukov:
So we can parse/print stream Json (via Circe, for example) with it.

sergeykolbasov:
Sure thing, I've been dealing with it for the last month or two. Will be happy to land a PR

vkostyukov:
Done in #1042.

PR #1042, opened 2018-11-18
fs2 support

sergeykolbasov:
Related to #1001 

Idea is to have `streamBody` and `jsonStreamBody` endpoints in core, along with `StreamingLaws`, so modules like iteratee and fs2 could implement only neccessary type classes and couple of specs while the rest of machinery is done inside of the core.

- [x] Refactoring
- [x] fs2 module + circe instances for it


build.sbt
```diff
@@ -62,7 +64,8 @@ val baseSettings = Seq(
   },
   scalacOptions in (Compile, console) += "-Yrepl-class-based",
   fork in Test := true,
-  javaOptions in ThisBuild ++= Seq("-Xss2048K")
+  javaOptions in ThisBuild ++= Seq("-Xss2048K"),
+  addCompilerPlugin("org.spire-math" % "kind-projector" % "0.9.8" cross CrossVersion.binary)
```
vkostyukov: What's this for?

core/src/main/scala/io/finch/internal/package.scala
```diff
@@ -122,4 +124,12 @@ package object internal {
     }
   }
 
+  def futureToEffect[F[_] : Effect, A](future: => Future[A]): F[A] = {
```
vkostyukov: I like idea. We needed it in a couple of other places so better to reuse it.

What do you think about a type class `LiftFuture` (by analogy with https://typelevel.org/cats-effect/typeclasses/liftio.html) that would let us lift both Scala futures and Twitter Futures into `F[_]: Effect`. I'd imagine we can simplify some code in Mapper with it.

core/src/main/scala/io/finch/streaming/StreamDecode.scala
```diff
@@ -0,0 +1,48 @@
...
+
+import com.twitter.io.Buf
+import io.finch.Application
+import scala.annotation.implicitNotFound
+
+/**
+  * Stream HTTP streamed payload represented as S[F, Buf] into
+  * a S[F, A] of arbitrary type `A`.
+  */
+trait StreamDecode[S[_[_], _], F[_], A] {
```
vkostyukov: I believe this belongs in the top-level namespace.

core/src/main/scala/io/finch/streaming/StreamDecode.scala
```diff
@@ -0,0 +1,48 @@
...
+  type ContentType <: String
+
+  def apply(stream: S[F, Buf], cs: Charset): S[F, A]
+
+}
+
+object StreamDecode {
+
+  @implicitNotFound(
+"""An Enumerator endpoint requires implicit Enumerate instance in scope, probably decoder for ${A} is missing.
```
vkostyukov: Outdated warning.

core/src/main/scala/io/finch/streaming/StreamDecode.scala
vkostyukov: Also, by analogy with `DecodePath`, `DecodeEntity`, I'd probably vote naming this `DecodeStream`.

core/src/main/scala/io/finch/streaming/StreamFromReader.scala
```diff
@@ -0,0 +1,20 @@
+package io.finch.streaming
+
+import com.twitter.io.{Buf, Reader}
+
+/**
+  * Create stream S[F, Buf] from com.twitter.io.Reader[Buf]
+  */
+trait StreamFromReader[S[_[_], _], F[_]] {
```
vkostyukov: I'm trying to understand why we need this. Is there a reason, `StreamDecode` can't take `Reader[Buf]` as an input?

core/src/main/scala/io/finch/EndpointModule.scala
```diff
@@ -289,6 +290,22 @@ trait EndpointModule[F[_]] {
   def asyncBody(implicit F: Effect[F]): Endpoint[F, AsyncStream[Buf]] =
```
vkostyukov: My questions about this:

 - Can we somehow allow users to work with `AsyncStream[Buf]` body stream via the new API (provide appropriate instances)
 - If we can, let's deprecate this method.

core/src/main/scala/io/finch/EndpointModule.scala
```diff
@@ -289,6 +290,22 @@ trait EndpointModule[F[_]] {
...
     Endpoint.asyncBody[F]
 
+  /**
+    * An alias for [[Endpoint.streamBody]]
+    */
+  def streamBody[S[_[_],_], A, CT <: String](implicit
```
vkostyukov: I like this a lot.

My only question would be, we should figure out if we can make this work with `AsyncStream[Buf]` as a stream or a plain `Reader[Buf]` as a stream.

build.sbt
sergeykolbasov: To simplify lambda types. It's required for sane use of type classes like `Functor[F[_]]` for things like `F[_, _]`

core/src/main/scala/io/finch/internal/package.scala
sergeykolbasov: Yeah, it's possible

core/src/main/scala/io/finch/streaming/StreamFromReader.scala
sergeykolbasov: It's a good question. I spent some time on this structure already and believe it's reasonable.

So the way how modules depend on each other now is following:

- iteratee depends on core and iteratee library
- fs2 depends on core and fs2 library
- circe depends on core, circe-iteratee, and circe-fs2 libraries.

If we remove `StreamReader` frome core it raises some interesting questions.
Who is responsible for transforming `Reader[Buf]` into `S[F[_], Buf]`?
If we merge it with`DecodeStream` and push to JSON modules, they're (only circe now) start concerning about low-level stuff about reading streamed HTTP body.
Also people would have to use `circe` module even if they need only to get simple `Buf` stream out of request. There is no way to default `DecodeStream` for `Buf` as in core we don't know which library is used, and there is now common specific type classes.

core/src/main/scala/io/finch/EndpointModule.scala
sergeykolbasov: I need to think about it. My gut feeling is that it's possible, but type aliases should be involved.

build.sbt
sergeykolbasov: Example:
https://github.com/finagle/finch/blob/219c0a147ac27dc2c3f59ce4a9cf3ceb97902519/json-test/src/main/scala/io/finch/test/JsonLaws.scala#L45

core/src/main/scala/io/finch/streaming/StreamFromReader.scala
rpless: > Also people would have to use circe module even if they need only to get simple Buf stream out of request. 

Trying to understand this bit in particular. Is this referring to the current state of streaming or the state the PR puts it in? My gut feeling is that we want to be able to stream raw `Buf`s from core and have various libraries utilize that as they need.
Seems to relate to:
> There is no way to default DecodeStream for Buf as in core we don't know which library is used

I'm not understanding this restriction. Feels like we should not need this info in the same way we don't for regular http requests, but I think I'm not understanding something about Streaming as it exists

sergeykolbasov: @rpless Sorry if I'm unclear.

In the current "master" status of streaming, one should add `finch-iteratee` dependency to enable endpoints like `enumerateBody` and `enumerateJsonBody` as they don't exist in core module. And circe module also provides decoder for `iteratee.Enumerator`.

In this PR _definition_ of endpoints was moved to core, so methods `streamBody` and `streamJsonBody`. Both of them rely on two type classes: `StreamFromReader` and `StreamDecode`. But due to the fact that core module itself doesn't depend on any streaming library, there is no default implementation of `StreamFromReader` could be provided as there is no such type `S[_[_], _]` available in core. There is default identity defition of `StreamDecode` for `Buf=>Buf` transformation available, though, because the only thing it does is to return `S[_[_], _]` back to caller.
These two type-classes are implemented in additional modules. `StreamFromReader` comes with one of the streaming modules: `fs2` or `iteratee` depending on what you'd like to use.
`StreamDecode` comes from JSON module, because it's the place where we put all the JSON related stuff.

What @vkostyukov proposes is to remove `StreamFromReader` and merge it with `StreamDecode` into single type class. My argument is that it would violate separation of concerns, because JSON library would need to understand all that HTTP `Reader[Buf]` stuff, and users will be forced to use this library as dependency, because it would be the only way to get this type class, even if the only thing you need is to stream bytes.

iteratee/src/main/scala/io/finch/iteratee/package.scala
```diff
@@ -15,82 +14,38 @@ import shapeless.Witness
...
-
-      final override def item: RequestItem = items.BodyItem
-      final override def toString: String = "enumeratorBody"
-  }
-
-  /**
-    * An evaluating [[Endpoint]] that reads a required chunked streaming JSON body, interpreted as
-    * an `Enumerator[Future, A]`. The returned [[Endpoint]] only matches chunked (streamed) requests.
-    */
-  def enumeratorJsonBody[F[_] : Effect, A](implicit
```
vkostyukov: Should we keep this around and deprecate so we can remove it next release?

fs2/src/main/scala/io/finch/fs2/package.scala
```diff
@@ -0,0 +1,62 @@
...
+    w: Witness.Aux[Application.Json]
+  ): ToResponse.Aux[Stream[F, A], Application.Json] = {
+    mkToResponse[F, A, Application.Json](delimiter = Some(ToResponse.NewLine))
+  }
+
+}
+
+trait StreamInstances {
+
+  implicit def streamToResponse[F[_] : Effect, A, CT <: String](implicit
```
vkostyukov: I believe these two instances can now go under `ToResponse` companion object?

core/src/main/scala/io/finch/EndpointModule.scala
vkostyukov: Another question about this. I'm trying to figure out if we can simply the type signature given we have the `io.finch.fs2` import in place? I'm thinking along the lines of `EndpointModule` but for streaming types. Don't have anything concrete in mind, just wanted to pointed out I'm thinking if there anything we can do.

fs2/src/main/scala/io/finch/fs2/package.scala
sergeykolbasov: Sadly, nope. `ToResponse` exists in the core, there is no `fs2` stuff there

iteratee/src/main/scala/io/finch/iteratee/package.scala
sergeykolbasov: Yes, we can do so.

core/src/main/scala/io/finch/EndpointModule.scala
sergeykolbasov: Sure
https://github.com/finagle/finch/pull/1042/commits/9e86108b117e517b2193989199bf33d3a3712449

sergeykolbasov: Check https://github.com/finagle/finch/pull/1042/commits/b03b88ad2339cebc1a106bb4f203d64895fde56e

It's possible using type alias `AsyncStreamF`. Also, I don't see much of a value to add straight `Reader[Buf]` support as one could always use `root` endpoint to extract it from request.

core/src/main/scala/io/finch/internal/package.scala
vkostyukov: Should we rebase this PR as we now have `ToEffect`?

sergeykolbasov: I just realized that `ToEffect` we have now is not lazy on the argument, so `Future` got executed immediately instead of being suspended in `Effect.async`.

sergeykolbasov: Here is what I mean:

```
import cats.effect.IO
import com.twitter.util.{Future, FuturePool}
import fs2._
import io.finch.internal._

def future: Future[Int] = FuturePool.unboundedPool {
  val i = 5
  println(i)
  i
}

Stream.repeatEval(futureToEffect[IO, Future, Int](future)) //it prints 5, but we didn't run stream yeat
```

sergeykolbasov: Yes, we can. But it would mean that we penalize the lawful effect uses with the additional allocation of `Function0` in `Mapper`. Actually, we're doing it already.

https://github.com/finagle/finch/blob/master/core/src/main/scala/io/finch/internal/Mapper.scala#L83-L84

https://github.com/finagle/finch/blob/master/core/src/main/scala/io/finch/internal/Mapper.scala#L83-L84

Shall we specialize it to make Futures called-by-name and `E : Effect` to be called by value?
Sorry, that was a too late observation from my side 😞

vkostyukov: I'm starting to think the signature of `ToEffect` may not be out problem. These got evaluated only when the request comes in anyway, and more importantly, only when `Endpoint.map*` is being run. When dealing with `Endpoint[IO, ?]`, for example, this only happens when we run the endpoint at the end, which pretty much neglects the difference between eager/lazy evaluation of that type-class.

But I agree, we should strive for consistency. I'm not super sure though, it's possible to get rid of call-by-name args in mapper. This [very simple test](https://github.com/finagle/finch/blob/master/core/src/test/scala/io/finch/EndpointSpec.scala#L263-L270) would likely to fail.

sergeykolbasov: @vkostyukov 
Probably, I've explained it not clear enough.

So my point is:
- Lawful effects `F[_]` _should not_ be called by name. We run them inside of Service, but it's enough to instantiate this effect once.
- Futures should be _should_ be called by name, because evaluation starts once Future is instantiated.
- Responses & Outputs _should_ be called by name, for the very same reason as for Future.

vkostyukov: Yeah, I get it. That's my point, even with `F[_]: Effect` and call-by-value, the linked test would fail - give it a try.

core/src/main/scala/io/finch/streaming/StreamFromReader.scala
vkostyukov: >  My argument is that it would violate separation of concerns, because JSON library would need to understand all that HTTP Reader[Buf] stuff, and users will be forced to use this library as dependency, because it would be the only way to get this type class, even if the only thing you need is to stream bytes.

I think we're mostly on the same page about this. I now see what you mean. Basically, people should be able to `streamBinaryBody[IO, Stream]` w/o specifying the content-type and `DecodeStream`, by definition, should have a notion of content type.

I agree we need two type-classes but I'm not sure I like the name `StreamFromReader`. How about `LiftReader`?

vkostyukov: To clarify:

 - `streamBinaryBody[F, S]: Endpoint[F, S[F, Buf]]` should not take Content-Type and should only require `LiftReader`
 - `streamJsonBody[F, S, A]: Endpoint[F, S[F, A]]` should require `DecodeStream`.

vkostyukov:
Hey @sergeykolbasov! I was thinking we could ship this in 0.27, before the NY. What's your plan with this?

sergeykolbasov:
@vkostyukov it should be fine by now

core/src/main/scala/io/finch/streaming/StreamFromReader.scala
```diff
@@ -0,0 +1,27 @@
...
+
+}
+
+object StreamFromReader {
+
+  def instance[S[_[_], _], F[_]](fn: Reader[Buf] => S[F, Buf]): StreamFromReader[S, F] = new StreamFromReader[S, F] {
+    def apply(reader: Reader[Buf]): S[F, Buf] = fn(reader)
+  }
+
+  type AsyncStreamF[F[_], A] = AsyncStream[A]
```
vkostyukov: Let's just strip out any async-stream APIs - there are better streaming APIs we're providing.

iteratee/src/main/scala/io/finch/iteratee/package.scala
vkostyukov: Sorry, I don't see where we redefined them. Mind pointing me out? I believe we should keep these for the sake of backwards compatibility.

core/src/main/scala/io/finch/EndpointResult.scala
```diff
@@ -45,6 +45,11 @@ sealed abstract class EndpointResult[F[_], +A] {
     case _ => None
   }
 
+  final def map[B](fn: A => B)(implicit F: Effect[F]): EndpointResult[F, B] = this match {
```
vkostyukov: Why this is needed?

core/src/main/scala/io/finch/EndpointStreamModule.scala
```diff
@@ -0,0 +1,26 @@
+package io.finch
+
+import cats.effect.Effect
+import com.twitter.io.Buf
+import io.finch.streaming.{DecodeStream, StreamFromReader}
+
+trait EndpointStreamModule[S[_[_], _], F[_]] extends EndpointModule[F] {
```
vkostyukov: I'm starting to think this could be an overkill at this point. I'd imagine the following still works (please, correct me if I'm wrong):

```
import io.finch.catsEffect._
import io.finch.fs2._

val e = streamBinaryBody[fs2.Stream]
```

I know I suggested that `StreamModule` idea in the first place. After Thinking about this more, the types seem less scary now, even if we use just a bare bones `EndpointModule`. What do you think?

core/src/main/scala/io/finch/internal/package.scala
```diff
@@ -122,4 +123,9 @@ package object internal {
     }
   }
 
+  /**
+    * Convert outlaw effect F[_] hidden inside of cats.Eval into lawful E[_]
+    */
+  def futureToEffect[E[_] : Effect, F[_], A](f: => F[A])(implicit te: io.finch.ToEffect[F, E]): E[A] = te(f)
```
vkostyukov: Why this is needed?

core/src/main/scala/io/finch/streaming/StreamFromReader.scala
vkostyukov: I'm still liking the `LiftReader` name more, but it's not a big deal - we can revisit the names in the later cleanup (these are internal and easy to change).

circe/src/main/scala/io/finch/circe/Decoders.scala
```diff
@@ -26,15 +30,33 @@ trait Decoders {
...
+  }
+
+  implicit def fs2Circe[F[_] : Sync, A : Decoder](implicit
+    monadError: MonadError[F, Throwable]
+  ): DecodeStream.Json[Stream, F, A] = {
+    DecodeStream.instance[Stream, F, A, Application.Json]((stream, cs) => {
+      val parsed = cs match {
+        case StandardCharsets.UTF_8 =>
+          stream
+            .flatMap(buf => Stream.fromIterator(buf.asByteArray.toIterator))
```
vkostyukov: Is there a way to avoid bouncing through `Iterator`?

fs2/src/main/scala/io/finch/fs2/package.scala
```diff
@@ -0,0 +1,62 @@
...
+import com.twitter.io.Buf
+import io.finch.internal.futureToEffect
+import io.finch.streaming.StreamFromReader
+import shapeless.Witness
+
+package object fs2 extends StreamInstances {
+
+  implicit def streamFromReader[F[_] : Effect]: StreamFromReader[Stream, F] = StreamFromReader.instance { reader =>
+    Stream
+      .repeatEval(futureToEffect(reader.read()))
```
vkostyukov: Thinking out loud. Maybe we shouldn't need to change `ToEffect` and instead, wrap the call to it with `F.delay` here?

What I'm getting at is I'm not sure `ToEffect` has to be lazy. As you mentioned, in the `Mapper`, we sort of already wrap it with `() =>`.

fs2/src/main/scala/io/finch/fs2/package.scala
```diff
@@ -0,0 +1,62 @@
...
+      .unNoneTerminate
+      .onFinalize(Effect[F].delay(reader.discard()))
```
vkostyukov: Will `onFinalize` fire if a failure is encountered?

fs2/src/main/scala/io/finch/fs2/package.scala
```diff
@@ -0,0 +1,62 @@
...
+import io.finch.internal.futureToEffect
+import io.finch.streaming.StreamFromReader
+import shapeless.Witness
+
+package object fs2 extends StreamInstances {
+
+  implicit def streamFromReader[F[_] : Effect]: StreamFromReader[Stream, F] = StreamFromReader.instance { reader =>
+    Stream
+      .repeatEval(futureToEffect(reader.read()))
+      .unNoneTerminate
```
vkostyukov: This is 🔥!

fs2/src/main/scala/io/finch/fs2/package.scala
```diff
@@ -0,0 +1,62 @@
...
+        .map(e.apply(_, cs))
+        .evalMap(buf => futureToEffect(writer.write(delimiter match {
+          case Some(d) => buf.concat(d)
+          case _ => buf
+        })))
+        .onFinalize(futureToEffect(writer.close()))
+        .compile
+        .drain
+
+      Effect[F].toIO(effect).unsafeRunAsyncAndForget()
```
vkostyukov: I'm thinking it would be great if we could somehow mix this `F[_]` with the one we return from an endpoint so we run them together, at the very last step.

This doesn't have to happen as part of this change as I understand we already do that for `iteratee`. Just trying to understand if we should file a ticket and follow up here. What do you think?

The "fix" could be as easy as having `ToResponse` returning `F[Response]`.

circe/src/main/scala/io/finch/circe/Decoders.scala
sergeykolbasov: The problem is that `circe-fs2` decoder expects stream of bytes as an input instead of a stream of arrays. Do you have any suggestion?

core/src/main/scala/io/finch/EndpointResult.scala
sergeykolbasov: I tried to make `streamJsonBody` simpler. We still need the access to `Input.request.charsetOrUtf8` to run the decoder, so it's impossible to just use `streamBody.map(a => b)`.
Check implementation of this endpoint.

circe/src/main/scala/io/finch/circe/Decoders.scala
vkostyukov: I believe we can use `Stream.chunk` and [`Chunk.array`](https://github.com/functional-streams-for-scala/fs2/blob/series/1.0/core/shared/src/main/scala/fs2/Chunk.scala#L586):

```
Stream.chunk(Chunk.array(buf.asByteArray))
```

fs2/src/main/scala/io/finch/fs2/package.scala
vkostyukov: Created https://github.com/finagle/finch/issues/1057

core/src/main/scala/io/finch/streaming/StreamFromReader.scala
vkostyukov: I also created a ticket so we kill async-stream all together #1055

fs2/src/main/scala/io/finch/fs2/package.scala
sergeykolbasov: Done

core/src/main/scala/io/finch/EndpointStreamModule.scala
sergeykolbasov: Removed

core/src/main/scala/io/finch/streaming/StreamFromReader.scala
sergeykolbasov: Removed

iteratee/src/main/scala/io/finch/iteratee/package.scala
sergeykolbasov: I put it back

core/src/main/scala/io/finch/EndpointModule.scala
sergeykolbasov: `AsyncStream` is going to be deprecated

core/src/main/scala/io/finch/streaming/StreamFromReader.scala
sergeykolbasov: Renamed into `LiftReader`

core/src/main/scala/io/finch/internal/package.scala
sergeykolbasov: Should be resolved by now

circe/src/main/scala/io/finch/circe/Decoders.scala
sergeykolbasov: Done

sergeykolbasov:
@vkostyukov mind to give it another look?

Let's make this PR a New Year resolution 😄

vkostyukov:
Going to look tonight @sergeykolbasov!

What's your opinion on this vs #1056. Which one should go first?

core/src/main/scala/io/finch/streaming/DecodeStream.scala
```diff
@@ -0,0 +1,48 @@
...
+
+import com.twitter.io.Buf
+import io.finch.Application
+import scala.annotation.implicitNotFound
+
+/**
+  * Stream HTTP streamed payload represented as S[F, Buf] into
+  * a S[F, A] of arbitrary type `A`.
+  */
+trait DecodeStream[S[_[_], _], F[_], A] {
```
vkostyukov: What do you think about defining this under `io.finch` (no need for `streaming`)?

core/src/main/scala/io/finch/internal/ToEffect.scala
```diff
@@ -1,12 +1,13 @@
 package io.finch.internal
 
 import cats.effect.Effect
-import cats.~>
 import com.twitter.util.{Future => TwitterFuture, Return, Throw}
 import scala.concurrent.{Future => ScalaFuture}
 import scala.util.{Failure, Success}
 
-trait ToEffect[A[_], B[_]] extends ~>[A, B]
+trait ToEffect[F[_], G[_]] {
```
vkostyukov: Looks like we don't have to change this?

core/src/main/scala/io/finch/streaming/DecodeStream.scala
sergeykolbasov: Done

core/src/main/scala/io/finch/internal/ToEffect.scala
sergeykolbasov: Yep

travisbrown:
@sergeykolbasov Hoping to have circe-fs2 published soon—it's tests turned up a bug in Jawn 0.14.0 that I thought we could merged pretty quickly: https://github.com/typelevel/jawn/pull/144

sergeykolbasov:
Thanks for the info, Travis!

Waiting for the resolution then

fs2/src/main/scala/io/finch/fs2/package.scala
vkostyukov: This should be possible after #1056.

PR #1043, opened 2018-11-19
fix typo in user guide regarding io.finch.Error

peterkrieg:
- should be "io.finch.error.NotPresent" instead of "io.finch.error.NotFound"


vkostyukov:
Thank you @peterkrieg! This LGTM.

Issue #1044, opened 2018-11-26
add support for FunctionK


jcouyang:
previously finch support any data type that can be convert to `TwitterFuture` by typeclass `ToTwitterFuture`

but since 0.25 it only support typeclass `Effect`, `Future` and `TwitterFuture`, which is not very friendly to customize data type that can be converted to `Effect` or `Future`

e.g. I compose all business logic to `Free[Program, Unit]`, in old finch I can simply define `ToTwitterFuture[Free[Program, ?]]` then finch can convert my type to Future

while in the new finch it can't

## maybe we can make this generic by using FunctionK
add the following to `Mapper.scala`
```scala
implicit def mapperToEffectOutputFunction[A, B, F[_], G[_]:Effect](f: A => F[Output[B]])(implicit cov:F~>G): Mapper.Aux[G, A, B] = new Mapper[G, A] {
    type Out = B
    def apply(e: Endpoint[G,A]) : Endpoint[G, B] = e.mapOutputAsync(a => cov.apply(f(a)))
  }

  implicit def mapperToEffectOutputValue[A, B, F[_], G[_]:Effect](f: => F[Output[B]])(implicit cov: F ~> G): Mapper.Aux[G, A, B] = new Mapper[G, A] {
    type Out = B
    def apply(e: Endpoint[G,A]) : Endpoint[G, B] = e.mapOutputAsync(a => cov.apply(f))
  }

  implicit def mapperFromKindOutputHFunction[F[_] : Effect, G[_], A, B, FN, FOB](f: FN)(implicit
    conv: G ~> F,
    ftp: FnToProduct.Aux[FN, A => FOB],
    ev: FOB <:< G[Output[B]]
  ): Mapper.Aux[F, A, B] = new Mapper[F, A] {
    type Out = B
    def apply(e: Endpoint[F,A]) : Endpoint[F, B] = e.mapOutputAsync(a => conv.apply(ev(ftp(f)(a))))
  }

```
then user can simply define how to convert their kind to some kind of `Effect`
```scala
case class MyType[A](value: A)

implicit val convert: MyType ~> IO  = Lambda[MyType ~> IO](a=>IO(a.value))
```

thus, use can define response in custom kind, not constrain to only `Effect` or `Future`

```scala
  def hello = get("hello" :: path[String]) { s: String =>
    MyType(Ok(Message(s)))
  }

```

vkostyukov:
[...]

PR #1045, opened 2018-11-29
mapper for `F[_]` should be generic at kind level

jcouyang:
just for demonstrating #1044

while it's generic all `mapperFrom*Future*` can be deleted

and user can implement `F ~> E:Effect` for their own F whenever/where ever they want

same thing should done to mapper for response as well


vkostyukov:
@sergeykolbasov @rpless Do you have an opinion?

@jcouyang Mind fixing a build in the meantime? I think scalafix wasn't happy about something - just check the build logs or verify it with `sbt validate` locally.

sergeykolbasov:
I'm still not sold to the `FunctionK` approach instead of custom type class due to the fact that this implicit conversion is supposed to be used all around the place and not only in `Mapper`. With this approach we have to carry around import with implicits inside instead of letting compiler to find it in companion object on itself.

jcouyang:
again @sergeykolbasov i'm not saying not using typeclass

what i'm trying to say is definitely we need typeclass, just you don't need to define your own since cats have it already

imagine that your typeclass would probably looks like
```scala
trait ToEffect[F[_], E[_]:Effect] {
  def apply[A](a: F[A]): E[A]
}
```
just look at the type, it's exactly the same as `FunctionK`/`~>`
```
trait FunctionK[F[_], G[_]] {
  def apply[A](fa: F[A]): G[A]
}
```

either way the implementation of the typeclass for future or any user custom type should be exactly the same

core/src/test/scala/io/finch/MethodSpec.scala
```diff
@@ -3,6 +3,8 @@ package io.finch
 import cats.effect.IO
 import com.twitter.finagle.http.Response
 import com.twitter.util.{Future => TwitterFuture}
+import io.finch.instances.scalaFuture._
```
jcouyang: also moved scala and twitter future into `instance` package because they are just implementation of typeclass `?~>Effect`, as cats convention
it still works if leave it in Mapper but i think it shouldn't belong to mapper anymore, mapper is generic for any kind now

sergeykolbasov: Well, that's exactly my point. We could avoid imports here and there using home-backed type class with companion object.
https://github.com/finagle/finch/blob/b03b88ad2339cebc1a106bb4f203d64895fde56e/core/src/main/scala/io/finch/internal/LiftToEffect.scala

jcouyang: Still not getting your point. You can move where ever you want, they are implicits
I just think they should organized this way in cats philosophy
Imagine If I implement your type class LiftToEffect for my type in my code, those future will be auto import to my context which I don't need at all.
How is that better than explicitly import

sergeykolbasov: > I just think they should organized this way in cats philosophy

Based on my experience working with different people through the years, this is one of the most frustrating pain points they had with Scala. It's not enough to write an actual code anymore, but also requires a user to know what and when to import.

> those future will be auto import to my context which I don't need at all.

They aren't auto-imported, but only by need. As far as you don't use these specific types in your context, compiler doesn't care much about the those implicit instances. This pattern is widely adopted in finch. Examples:
- [Encode](https://github.com/finagle/finch/blob/master/core/src/main/scala/io/finch/Encode.scala#L57) and [DecodeEntity](https://github.com/finagle/finch/blob/6f5006fd4395c8c8066692c44c889bdd0ea43d5d/core/src/main/scala/io/finch/DecodeEntity.scala#L28)
- [ToResponse](https://github.com/finagle/finch/blob/6f5006fd4395c8c8066692c44c889bdd0ea43d5d/core/src/main/scala/io/finch/ToResponse.scala#L72)

jcouyang: > It's not enough to write an actual code anymore but also requires a user to know what and when to import

I totally understand what your concern is and I think
cats already have pretty good pattern to cater both kind of users

for peeps don't know what to import, they simply import everything and let the compile pick for them
```scala
import cats.implicits._
```
for peeps know exactly what they need, i.e. functor instance of list
```scala
import cats.instances.list._
```
they both behave the same from user point of view but I believe the second one is more gentle to the compiler and more explicit about the intention

> They aren't auto-imported, but only by need

they aren't at run time but are there for compiler search

> This pattern is widely adopted in finch

i think decoupling typeclass and data type is the pattern that functional programming different from OO. it make sense to put all instance of data type in data type's companion object but not in typeclass's companion object 
i.e.
when` import cats.effect.IO` you have all instances of datatype IO for typeclass Monad, Effect etc

but when you `import cats.effect.Effect` you should only have typeclass `Effect`, because when i import the typeclass, it means I want to implement it for my data type, which means I'm using custom data type, not IO, all instance of IO for that typeclass is the last thing I need.

since finch adopted cats, i reckon keeping only one pattern is easier for both maintainer and user

sergeykolbasov: I see your point here, but in my opinion it's rather subjective and based on the experience with certain environment. My argument comes from the UX point of view, rather than FP (or any other) purity. Extrapolating the example of `Encode` and `ToResponse` toward the state where every implicit is placed under a separate package, we end up either with dozens of imports for each primitive type like `Int`, `Boolean` or `String`, or introduce the single import of truth `io.finch.implicits._`, which is, sorry for being blunt, but just a cargo cult.

jcouyang: From UX point of view, I reckon user choose Finch over other is because fp since Finch branding itself so isn't it. At least I am because it's compatible with cats ecosystem and I expected seamless UX as cats.
If user come from other experience other than cats isn't finatra the a better choice then

rpless: I'd like to respond as user, not a maintainer. The team I work with chose Finch 3 years ago because it was easy to use, very concurrent, and gave us all the benefits of Finagle to use as we needed. We've never followed how Cats does usability things (and my teammates have struggled with Cats' conventions), I that find while I understand the import strategy, I'm constantly teaching it or reminding others of it. While I'd personally be willing to accept another import (and this PR), we can't discount the users. I'm with @sergeykolbasov on this one.I think we should let it be discovered by the compiler.

core/src/main/scala/io/finch/package.scala
```diff
@@ -1,13 +1,16 @@
...
-
+import cats.~>
 /**
  * This is a root package of the Finch library, which provides an immutable layer of functions and
  * types atop of Finagle for writing lightweight HTTP services.
- */
+  */
 package object finch extends Outputs with ValidationRules {
 
+  trait ToEffect[F[_], E[_]] extends (F ~> E)
```
vkostyukov: Can this live in it's own file? Also, I'd probably lobby for `io.finch.internal` package for this.

core/src/main/scala/io/finch/instances/IO.scala
```diff
@@ -0,0 +1,30 @@
+package io.finch
+package instances
+
+import cats.effect.IO
+import cats.~>
+import com.twitter.util.{Future => TwitterFuture, Return, Throw}
+import scala.concurrent.{Future => ScalaFuture}
+import scala.util.{Failure, Success}
+
+trait ToIO {
```
vkostyukov: What's the reasoning for this being a trait? Why not move these two instances under `ToEffect` companion object?

core/src/main/scala/io/finch/internal/Mapper.scala
```diff
@@ -1,13 +1,11 @@
-package io.finch.internal
```
vkostyukov: I think normally we just leave the package name as a single fqn.

core/src/main/scala/io/finch/instances/IO.scala
jcouyang: abstract them as trait so `ToEffect` could be more extensible, since `IO` is only one data type that implemented `Effect[_]`, there could be other i.e. monix's `Task`, so that we could easily just implement `Effect[Tasks]` as `trait ToTask` and
```scala
object ToEffect extends instances.ToIO with instances.ToTask
```

core/src/main/scala/io/finch/package.scala
jcouyang: Absolutely can do, but I don't think it's for internal, the typeclass is intent to let user implement for their own type if they are not using *Future, my intention is keep them in `io.finch.implicits._` just like cats, doobie, http4s etc
but to address @sergeykolbasov and @rpless 's concern for user confusing about import implicits
this is a good place when user simply just `import io.finch._` and get all those implicits

core/src/main/scala/io/finch/instances/IO.scala
sergeykolbasov: Would it be shorter to generalize it for any `E : Effect` instead?

jcouyang: 🤔 good idea, these Future things just instances of `ToEffect[?, E]`

core/src/main/scala/io/finch/package.scala
vkostyukov: Sounds good.

core/src/main/scala/io/finch/internal/IdToEffect.scala
```diff
@@ -0,0 +1,10 @@
+package io.finch.internal
+
+import cats.effect.Effect
+import io.finch.ToEffect
+
+trait IdEffect {
```
vkostyukov: I lost our previous conversion on this, but I'm still not sure why we need these instances defined outside of `ToEffect` companion object?

jcouyang: the rational is that comparing to put everything in one single big object, it's more extensible and modular by putting `ToEffect`'s instances in different files
the implementation of `ToEffect[E,E]` here is totally irrelevant to the implementation of `ToEffect[*Future, E]`(ideally TwtterFuture and ScalaFuture should be separated too but I think it's fine that they share some similarity)

vkostyukov: Extensible in a sense of for Finch users or for Finch developers? I'm trying to understand if there is any value in the latter.

jcouyang: for finch developers, e.g. if finch want to natively support some new type other then Future, Id, it's easier to add new trait for `object ToEffect` to extends

user shouldn't care these traits so that's why I move them back to `internal` folder

vkostyukov: Sorry, I'm a little skeptical on what's easier: create a new trait and extend it in an object, or just put your stuff under the same object. I personally like the latter more b/c it's just less code and fewer indirection and fewer symbols. Perhaps, we can revisit it when we have 10-20 instances the grouping is the only vehicle to easier reasonability and discoverability.

jcouyang: Does that means you're ok with a big object live in package.scala?
I personally really don't like putting irrelevant thing in single file, and keeping package.scala as concise as possible

vkostyukov: I'd prefer a separate file for this, `ToEffect.scala` but also curious to hear what other think.

sergeykolbasov: Agree with @vkostyukov 

Also, one thing to consider. If at some point we'd like to ship Monix support with finch, most likely it'll be separate `finch-monix` submodule anyway

core/src/main/scala/io/finch/internal/FutureToEffect.scala
```diff
@@ -0,0 +1,28 @@
+package io.finch.internal
+
+import cats.effect.Effect
+import com.twitter.util.{Future => TwitterFuture, Return, Throw}
+import io.finch.ToEffect
+import scala.concurrent.{Future => ScalaFuture}
+import scala.util.{Failure, Success}
+
+trait FutureToEffect {
+  implicit def twFutureToEffect[E[_]: Effect]: ToEffect[TwitterFuture, E] = new ToEffect[TwitterFuture, E] {
```
sergeykolbasov: I wonder if we can `@deprecate` those parameters to encourage users to upgrade their codebase to lawful effects.

vkostyukov: Maybe we don't have to? Maybe deprecating these wasn't the good idea. These converters are pretty slim and going from Future to Effect turned out to be pretty straightforward. I'm thinking to keep them as is and revisit it later, after 1.0 or after we rework our mapper entirely.

core/src/main/scala/io/finch/internal/IdToEffect.scala
jcouyang: I don't think we need another modules to support monix , actually we probably don't need to do anything and monix should be supported already, since we have generic endpointmodule[E[_]], monix task is an Effect already or at least for sure it has Async implemention already

core/src/main/scala/io/finch/internal/FutureToEffect.scala
jcouyang: That's what I was thinking

core/src/main/scala/io/finch/internal/IdToEffect.scala
sergeykolbasov: @jcouyang It goes a bit further then just a `Mapper` support:
https://github.com/finagle/finch/blob/master/core/src/main/scala/io/finch/package.scala#L11

It allows to enable simple syntax for all the `Endpoint.` methods

jcouyang: I'm not talking about mapper, I'm talking about endpointmodule [F[_]]

The F is generic as Effect already isn't it?

sergeykolbasov: Shortly, `EndpointModule[F[_]]` is a container for all the aliases for `Endpoint.` methods such as `get`, `post`, `path` and so on. Otherwise user would neeed to type manually each time something like:
```scala
Endpoint.path[IO, String]("foo")

//equal to
import io.finch.catsEffect._
path("foo")
```

There are multiple ways to enable such syntax, one of them is to have the object extending `EndpointModule[AnyEffect]`. Details: https://github.com/finagle/finch/releases/tag/v0.25.0

jcouyang: > I'd prefer a separate file for this, ToEffect.scala

Hmm🤔 A second thought, if move to separate file, object mapper can simpliy extend the trait and has the ability to convert none effect to effect, which is more composable and less implicit

And type alias for `F ~>E` can be eliminated too, user don't have to import anything from Finch to implement `F ~>E` for their own type then

Not sure if it makes sense to you, I'll refactor it to demonstrate what I meant anyway

jcouyang: > object mapper can simpliy extend the trait and has the ability to convert none effect to effect

forget it, that theory didn't work out

> I'd prefer a separate file for this, ToEffect.scala

@vkostyukov just to be clear,  i dont want to change back and forth again
IIUC is this what you saying?
- combine IdEffect.scala and FutureToEffect.scala into `ToEffect.scala`

**or**
- move everything including the `trait ToEffect` from `package.scala` into `ToEffect.scala`

sergeykolbasov:
LGTM

vkostyukov:
Thanks again @jcouyang! This is great work!

PR #1048, opened 2018-12-05
Use .recoverWith instead of .handleErrorWith

sergeykolbasov:
`.handleErrorWith` doesn't re-raise the error if the partial function doesn't match, so it leads to `MatchError` exception pointing to `ToService.handler` in the `Future` from `Service`.

Users lose the information about the type of exception caused the original problem.


sergeykolbasov:
@vkostyukov could we roll out .1 release afterward?

Issue #1041, opened 2018-11-16
Can we get rid of Bootstrap.negotiateContentType option


vkostyukov:
And, instead derive it fro what's passed into the `ContentType` type. See [this gitter discussion.](https://gitter.im/finagle/finch?at=5bef3371b86c70503f40ac40)

sergeykolbasov:
Closed in #1050

PR #1050, opened 2018-12-06
Enable content negotiation automatically

sergeykolbasov:
Use `shapeless.OrElse` to resolve implicit evidence of `CTH` being a Coproduct or return `DummyImplicit` otherwise. Then fold this structure into boolean value to understand if we need to negotiate on Content-Type.

#1041

### DIFF
--- a/core/src/main/scala/io/finch/ToService.scala
+++ b/core/src/main/scala/io/finch/ToService.scala
@@ -44,7 +44,6 @@ object ToService {
   final case class Options(
     includeDateHeader: Boolean,
     includeServerHeader: Boolean,
-    negotiateContentType: Boolean,
     enableMethodNotAllowed: Boolean,
     enableUnsupportedMediaType: Boolean
   )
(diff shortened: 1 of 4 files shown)


core/src/main/scala/io/finch/ToService.scala
```diff
@@ -98,17 +97,19 @@ object ToService {
       }
   }
 
+  type IsCoproduct[C] = OrElse[C <:< Coproduct, DummyImplicit]
```
vkostyukov: Is this type-alias for readability or it doesn't work w/o it?

sergeykolbasov: Just for the sake of readability, yes.

vkostyukov: 👍 Maybe rename it to `IsNegotiable` so it make sense in our domain.

core/src/main/scala/io/finch/ToService.scala
```diff
@@ -117,8 +118,8 @@ object ToService {
 
             Trace.captureIfNeeded(trc)
 
-            val accept =
-              if (opts.negotiateContentType) req.accept.map(a => Accept.fromString(a)) else Nil
+            val negotiateContent = isNegotiable.fold(_ => true, _ => false)
```
vkostyukov: I think we can move this out of `apply` (request path) and reuse the value instead of recomputing it.

Issue #1040, opened 2018-11-15
Provide Enumerate.Text instance out of the box


vkostyukov:
See #1037 for more context.

vkostyukov:
Done in #1056.

Issue #1038, opened 2018-11-15
Input should be able to create chunked requests


vkostyukov:
See #1037 for a discussion.

vkostyukov:
It should be easier to do when #1042 lands (as we would have all the needed abstractions in place).

vkostyukov:
I'm thinking something along the lines:

```scala
val i = Input.post("/").withStream[Application.Json](stream)
```

Where stream is anything that has a `EncodeStream` instance (similar to `DecodeStream` from #1042).

vkostyukov:
Done in #1056.

PR #1056, opened 2018-12-21
Abstract over Stream type in the outbound path

vkostyukov:
This to address #1038 as well as make it more pluggable to adopt new streaming libraries (eg FS2). The type-class name is inspired by what Sergey is doing in #1042. The new machinery allows us to reuse the same instances between `ToResponse` (when serving streaming content) and `Input.withStream` allowing users to created chunked requests for testing.

Here is how it looks and feels:

```scala
import io.finch._, io.finch.iteratee._, io.interatee.Enumerator, cats.effect.IO

val i = Input
  .post("/")
  .withStream[Text.Plain](Enumerator.enumerate[IO, String]("foo", "bar", "baz"))
```

Also possible to encode JSON (as you would imagine). This does the right thing and encodes ND (newline-delimited) JSON stream, reusing the same code that we use on the outbound path:

```scala
import io.finch._, io.finch.iteratee._, io.interatee.Enumerator, cats.effect.IO
import io.finch.circe._, io.circe.generic.auto._

case class Foo(s: String)

val i = Input
  .post("/")
  .withBody[Application.Json](Enumerator.enumerate[IO, Foo](Foo("foo"), Foo("bar"), Foo("baz")))
```

As a bonus, this addresses #1040.

### DIFF
--- a/core/src/main/scala/io/finch/ToResponse.scala
+++ b/core/src/main/scala/io/finch/ToResponse.scala
@@ -76,12 +76,24 @@ trait HighPriorityToResponseInstances extends LowPriorityToResponseInstances {
     w: Witness.Aux[CT]
   ): Aux[A, CT] = instance { (a, cs) =>
     val buf = e(a, cs)
-    val rep = Response()
+    val rep = Response(Version.Http11, Status.Ok)
 
     if (!buf.isEmpty) {
       rep.content = buf
       rep.headerMap.setUnsafe("Content-Type", w.value)
     }
+
+    rep
+  }
+
+  implicit def streamToResponse[S[_[_], _], F[_], A, CT <: String](implicit
+    E: EncodeStream.Aux[S, F, A, CT],
+    W: Witness.Aux[CT]
+  ): Aux[S[F, A], CT] = instance { (a, cs) =>
+    val stream = E(a, cs)
+    val rep = Response(Version.Http11, Status.Ok, stream)
+    rep.headerMap.setUnsafe("Content-Type", W.value)
+    rep.headerMap.setUnsafe("Transfer-Encoding", "chunked")
 
     rep
   }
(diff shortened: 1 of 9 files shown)


vkostyukov:
@sergeykolbasov Do you prefer this or #1042 being merged this?

core/src/main/scala/io/finch/ToResponse.scala
```diff
@@ -85,6 +85,17 @@ trait HighPriorityToResponseInstances extends LowPriorityToResponseInstances {
 
     rep
   }
+
+  implicit def streamToResponse[S[_[_], _], F[_], A, CT <: String](implicit
```
sergeykolbasov: 👍

vkostyukov:
@sergeykolbasov I'm thinking we should probably go the opposite way - merge smaller pieces when possible. I was worried you'd need to resolve a lot of conflicts in your PR hence kept it open. If you're up for a challenge (and it sounds like you're), I'd say we merge this first and rebase your work.

I wouldn't even mind if we split your PR into couple of independent changes if possible.

vkostyukov:
Given #1042 is blocked on a new circe-fs2 release (am I right, @sergeykolbasov ?), should we merge this first and, perhaps, cut a release?

BenFradet:
I don't think I can do so myself but I'm all for it :+1:

travisbrown:
Okay, circe-fs2 is ready: https://github.com/circe/circe-fs2/releases/tag/v0.11.0

vkostyukov:
Thank you, Travis!

sergeykolbasov:
@vkostyukov FYI: just merged #1042

vkostyukov:
Awesome @sergeykolbasov! I will rebase this tonight.

vkostyukov:
Rebased. Some pieces feel repetitive but it's workable. I think we should merge this in and keep iterating.

PR #1066, opened 2019-01-10
New names for streaming bodies

vkostyukov:
What I like in this PR:

 - Renames streaming endpoints to better match the current schema. For example, users can type `binaryBody` and see all options available in the IDE autocompletion: `binaryBody`, `binaryBodyStream`, `binaryBodyOption`
 - This also adds aliases for `textBodyStream`
 - New streaming endpoint, `stringBodyStream` to work with raw strings
 - Changes the chunk type from `Buf` to `Array[Byte]` for binary streams to match how `binaryBody` is defined (in terms of `Array[Byte]`).
- Scaladoc

What I'm not stocked about in this PR:
 - I had to change the `LiftReader` signature so it also accepts a `process` function. This is needed for zero-overhead `stringBodyStream` endpoint. Alternatively, we can summon  `Functor` instance for a stream and just map over it. I say we revisit this later.


sergeykolbasov:
I had exactly opposite intention actually 😄 
To show what streaming options are available once person start typing `stream..`

vkostyukov:
To be honest, IntelliJ autocompletion should still be able to show streaming bodies if `stream` is typed. I still like these names more b/c they follow the pattern we use for bodies:

 - `body`, `bodyOption`, `bodyStream` - raw body
 - Add a `json` prefix: `jsonBody`, `jsonBodyOption`, `jsonBodyStream`
 - Add `binary` prefix: `binaryBody`, `binaryBodyOption`, `binaryBodyStream`, etc

rpless:
I prefer the naming scheme in this PR. Keeping the names consistent with the rest of bodies feels good to me.

I also feel like we shouldn't need to pass `process` into `LiftReader`, but I don't have an alternative at this point. 

:+1: here.

Issue #1057, opened 2018-12-21
We should only run streams at the very last step in ToService


vkostyukov:
Given the type signature of `ToResponse`, we have to run those streams out of band (not as part of the `IO` we run in `ToService`). Let's figure out how we can merge them together and run at once.

PR #1073, opened 2019-01-18
Do not run stream out of band

vkostyukov:
See #1057.

This refactors `ToResponse` so it returns `F[Response]`, instead of just `Response` to allow for in-band stream evaluation.

UPDATE:

Nothing concerning in the benchmark:

```
BEFORE:

[info] JsonBootstrapBenchmark.foos                                thrpt   10   28777.149 ± 1351.610   ops/s
[info] JsonBootstrapBenchmark.foos:·gc.alloc.rate.norm            thrpt   10  103896.055 ±    0.019    B/op
[success] Total time: 37 s, completed Jan 19, 2019 5:09:30 PM


AFTER:

[info] Benchmark                                                   Mode  Cnt       Score      Error   Units
[info] JsonBootstrapBenchmark.foos                                thrpt   10   29342.207 ± 1092.418   ops/s
[info] JsonBootstrapBenchmark.foos:·gc.alloc.rate.norm            thrpt   10  104088.054 ±    0.018    B/op
```

### DIFF
--- a/core/src/main/scala/io/finch/EncodeStream.scala
+++ b/core/src/main/scala/io/finch/EncodeStream.scala
@@ -6,21 +6,21 @@ import java.nio.charset.Charset
 /**
  * A type-class that defines encoding of a stream in a shape of `S[F[_], A]` to Finagle's [[Reader]].
  */
-trait EncodeStream[S[_[_], _], F[_], A] {
+trait EncodeStream[F[_], S[_[_], _], A] {
 
   type ContentType <: String
 
-  def apply(s: S[F, A], cs: Charset): Reader[Buf]
+  def apply(s: S[F, A], cs: Charset): F[Reader[Buf]]
 }
 
 object EncodeStream {
 
-  type Aux[S[_[_], _], F[_], A, CT <: String] =
-    EncodeStream[S, F, A] { type ContentType = CT }
+  type Aux[F[_], S[_[_], _], A, CT <: String] =
+    EncodeStream[F, S, A] { type ContentType = CT }
 
-  type Json[S[_[_],_], F[_], A] = Aux[S, F, A, Application.Json]
+  type Json[F[_], S[_[_],_], A] = Aux[F, S, A, Application.Json]
 
-  type Text[S[_[_],_], F[_], A] = Aux[S, F, A, Text.Plain]
+  type Text[F[_], S[_[_],_], A] = Aux[F, S, A, Text.Plain]
 }
 
 
--- a/core/src/main/scala/io/finch/ToResponse.scala
+++ b/core/src/main/scala/io/finch/ToResponse.scala
@@ -1,5 +1,7 @@
 package io.finch
 
+import cats.{Applicative, Functor}
+import cats.effect.Effect
 import com.twitter.finagle.http.{Response, Status, Version}
 import java.nio.charset.Charset
 import scala.annotation.implicitNotFound
@@ -8,10 +10,10 @@ import shapeless._
 /**
  * Represents a conversion from `A` to [[Response]].
  */
-trait ToResponse[A] {
+trait ToResponse[F[_], A] {
   type ContentType
 
-  def apply(a: A, cs: Charset): Response
+  def apply(a: A, cs: Charset): F[Response]
 }
 
 trait ToResponseInstances {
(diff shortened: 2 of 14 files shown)


iteratee/src/main/scala/io/finch/iteratee/package.scala
```diff
@@ -68,19 +68,18 @@ trait IterateeInstances {
...
+    def apply(s: Enumerator[F, A], cs: Charset): F[Reader[Buf]] = {
       val p = new Pipe[Buf]
       val run = s
         .ensure(F.suspend(TE(p.close())))
         .map(chunk => encodeChunk(chunk, cs))
         .into(writeIteratee(p))
 
-      F.toIO(run).unsafeRunAsyncAndForget()
-      p
+      F.guarantee(F.pure(p: Reader[Buf]))(F.delay(F.toIO(run).unsafeRunAsyncAndForget()))
```
vkostyukov: This is the relevant piece. As we can't just `flatMap` from `run` (it only resolves stream terminates), we need a way to asynchronously mix an effect.

I'm wondering if there is a better way to do that (looking for suggestions).

joroKr21: You could `F.runAsync(run)(_ => IO.unit).to[F]`. But I'm not sure I understand the goal here. Is it supposed to guarantee that everything gets encoded into `p` even on cancellation? `F.pure` can't fail so cancellation is the only thing left.

vkostyukov: Oh yeah, very good point about cancelation. I guess I'd need to use `guaranteeCase` to handle that?

fs2/src/main/scala/io/finch/fs2/package.scala
```diff
@@ -68,12 +70,11 @@ trait StreamInstances {
         .compile
         .drain
 
-      F.toIO(run).unsafeRunAsyncAndForget()
-      p
+      F.productR(F.runAsync(run)(this).to[F])(F.pure(p))
```
vkostyukov: @joroKr21 I took your suggestion.

iteratee/src/main/scala/io/finch/iteratee/package.scala
vkostyukov: To answer your question: we previously unconditionally evaluated a stream when we call `ToResponse` (from within `ToService`) We want to change that and stay within `F[_]` as long as possible, and run everything together.

The cancelation part is really interesting. I'd imagine to properly support it here, we need to require `ConcurrentEffect` so we can `runCancellable`? Otherwise, I'm not sure the cancelation signal will get propagated via `runAsync`.

fs2/src/main/scala/io/finch/fs2/package.scala
joroKr21: Wait, what does it mean to pass `this` as the callback?

sergeykolbasov:
LGTM 👍

fs2/src/main/scala/io/finch/fs2/package.scala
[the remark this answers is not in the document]
sergeykolbasov: @vkostyukov I just tried something similar. It's not propagated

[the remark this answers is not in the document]
joroKr21: Yes you need `runCancelable` which is defined in `ConcurrentEffect` to propagate cancellation. But if you have `Concurrent`, you can just `start` a fiber without "running" anything.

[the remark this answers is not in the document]
vkostyukov: Thank you both. Maybe we should just use fibers then? It looks like the following should do it:

```scala
F.bracket(F.start(run))(_ => F.pure(p: Reader[Buf]))(_.cancel)
```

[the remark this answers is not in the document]
vkostyukov: It's pretty counterintuitive though.

This hangs in the test (fiber never starts?) 

```scala
F.bracket(F.start(run))(_ => F.pure(p: Reader[Buf]))(_.cancel)
```

This works but lacks resource cleanup.
```scala
F.productR(F.start(run))(F.pure(p: Reader[Buf]))
```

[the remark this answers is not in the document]
vkostyukov: I think I know what it is. I just cancel it immediately, no matter how `use` is terminated.

fs2/src/main/scala/io/finch/fs2/package.scala
```diff
@@ -68,14 +70,15 @@ trait StreamInstances {
         .compile
         .drain
 
-      F.toIO(run).unsafeRunAsyncAndForget()
-      p
+      F.bracketCase(F.start(run))(_ => F.pure(p: Reader[Buf])) {
```
vkostyukov: @sergeykolbasov @joroKr21 Please take a look.

examples/src/main/scala/io/finch/iteratee/Main.scala
```diff
@@ -35,7 +37,7 @@ import scala.util.Random
   *   $ curl -X POST --header "Transfer-Encoding: chunked" -d '{"i": 40} {"i": 42}' localhost:8081/streamPrime
   * }}}
   */
-object Main extends EndpointModule[IO] {
+object Main extends IOApp {
```
vkostyukov: Note: `ConcurrentEffect[IO]` instances requires `ContextSwitch[IO]` hence only compiles within `IOApp`.

fs2/src/main/scala/io/finch/fs2/package.scala
joroKr21: Sorry, I was really busy these days. This change LGTM but it's unfortunate that we have to ask for ConcurrentEffect since this is the most powerful typeclass in the hierarchy and permits the least amount of implementations (e.g. Scalaz Task can implement Effect but not ConcurrentEffect). I'm trying but failing to understand why evaluation of the Stream has to happen  here.

vkostyukov: Thanks a lot for looking into this. What we simply need to do is to run `run: F[Unit]` and forget about it (in a sense only worrying about cancelations) within [`ToService`](https://github.com/finagle/finch/blob/master/core/src/main/scala/io/finch/ToService.scala#L126-L143). So effectively we need to include `run: F[Unit]` evaluation as part of the effect we return here, from `ToResponse`, but not wait for its result.

One way to do that is to `runAsyncAndForget` but we figured this rules out cancelation signals. Another way is to use `ConcurrentEffect` and fibers as we're basically racing two effects. I'm not sure there are other ways to express that but I'm wiling to spend more time researching (there is no rush on merging this).

Also, good point about the callback.

vkostyukov:
@sergeykolbasov @joroKr21 I want to revert back to the previous idea, `.runAsync(run)(_ => IO.unit).to[F]`, for now so we don't require `ConcurrentEffect` just yet. I believe we'll be in a better shape and can revisit this again, later. Thoughts?

core/src/main/scala/io/finch/EncodeStream.scala
```diff
@@ -6,21 +6,21 @@ import java.nio.charset.Charset
 /**
  * A type-class that defines encoding of a stream in a shape of `S[F[_], A]` to Finagle's [[Reader]].
  */
-trait EncodeStream[S[_[_], _], F[_], A] {
+trait EncodeStream[F[_], S[_[_], _], A] {
```
joroKr21: Does this improve type inference?

core/src/main/scala/io/finch/ToResponse.scala
```diff
@@ -72,59 +79,60 @@ object ToResponse extends ToResponseInstances {
...
+      new FromCoproduct[F, C] {
         type ContentType = CT
-        def apply(c: C, cs: Charset): Response = fn(c, cs)
+        def apply(c: C, cs: Charset): F[Response] = fn(c, cs)
       }
 
-    implicit def cnilToResponse[CT <: String]: Aux[CNil, CT] =
-      instance((_, _) => Response(Version.Http10, Status.NotFound))
+    implicit def cnilToResponse[F[_], CT <: String](implicit
+      F: Effect[F]
```
joroKr21: I think this could be relaxed to `Applicative`

joroKr21:
@vkostyukov I think that's fine

> I want to revert back to the previous idea, .runAsync(run)(_ => IO.unit).to[F], for now so we don't require ConcurrentEffect just yet. I believe we'll be in a better shape and can revisit this again, later. Thoughts?

Maybe it would be best to provide both instances (they would have to be prioritized) so that when the user does have `ConcurrentEffect[F]` they can benefit from cancellation. I'm not sure it's possible though. In theory it should be. I can give it a go after you merge this PR.

core/src/main/scala/io/finch/ToResponse.scala
vkostyukov: 👍

core/src/main/scala/io/finch/EncodeStream.scala
vkostyukov: It's mostly for consistency. Our other types take `F[_]` (effect) as a first argument.

vkostyukov:
> > I want to revert back to the previous idea, .runAsync(run)(_ => IO.unit).to[F], for now so we don't require ConcurrentEffect just yet. I believe we'll be in a better shape and can revisit this again, later. Thoughts?
> 
> @vkostyukov I think that's fine
> 
> Maybe it would be best to provide both instances (they would have to be prioritized) so that when the user does have `ConcurrentEffect[F]` they can benefit from cancellation. I'm not sure it's possible though. In theory it should be. I can give it a go after you merge this PR.

Yea, we sort of do that in `ToService` but I was meaning to double check if it works properly (after working on this PR, I'm not longer sure).

core/src/main/scala/io/finch/EncodeStream.scala
joroKr21: Ok, I think it probably won't matter for type inference that much. At least not in the same way that `EitherT[F, E, A]` matters. Users would probably rarely work directly with `EncodeStream`.

Issue #1061, opened 2019-01-08
Effect-aware Service and Filter


sergeykolbasov:
I have the following idea.

Introduce two additional abstractions that are aware of effect type:

```scala
trait Service[F[_]] {
  def apply(req: Request): F[Either[Error, Response]]
  def compose(service: Service[F]): Service[F]
}

trait Filter[F[_]] {
  def apply(req: Request, service: Service[F]): F[Either[Error, Response]]
  def andThen(service: Service[F]): Service[F]
  def andThen(filter: Filter[F]): Filter[F]
}
```

At Zalando we have a case to run auth and tracing inside of the Finagle filters, but the functions we're calling might return `IO`, `Task` or some `F[_]` as a result instead of `Future`. Running effects in multiple places is a thing we'd like to avoid.

Having a `compose` method on `Service` also should enable the ability of applying filters to some specific services only:

```scala
val filterA: Filter[IO] = ...
val serviceA: Service[IO] = filterA.andThen(someService)
val serviceB: Service[IO] = anotherService

val final: Service[IO] = serviceA.compose(serviceB) //filterA is not applied to the part B of final service
```

Now on `Either[Error, Response]`. This one is required for services composition, as we need to understand whether we should call next service in the chain (without applying the filter) or return the response. So if the underlying `Endpoint[A :+: B :+: C]` doesn't match the request, we should return `Left(Error.NotFound)` that will be intercepted by composition.

In the very end everything is converted into usual Finagle Service.

What do you think folks?

vkostyukov:
I've been thinking a lot about making auth (and, perhaps, tracing) easier to do with endpoints. Not super sure we need these abstractions for this though (not both of these for sure). It feels like when a user bootstraps their endpoint, they should be able to configure auth/tracing/logging for each of them separately. Technically, all of this is more or less possible via a simple wrapping `Endpoint[F[_], A](underlying: Endpoint[F, A])`.

Maybe this warrants a new combinator on endpoint (not sure about the name and the type-signature yet)?

```scala
trait Endpoint {
  def before(f: Input => F[Output[A]]): Endpoint
}
```

sergeykolbasov:
If I understand correctly, we shall not invoke `underlying.apply` if
```scala
F[Output[_]] =:= F[Output.failure]
```

It would mean that this combinator is obliged to return `F[_]`, not an `Endpoint` or `EndpointResult`. This is probably would be a mess.

sergeykolbasov:
[...]

Issue #1051, opened 2018-12-08
Can/should Bootstrap.serve take a Finagle Filter?


vkostyukov:
This is a very popular use-case: add a filter to certain endpoints wile leaving others. Here is the recent request [about auth](https://gitter.im/finagle/finch?at=5c0bda3126de6f0822a1924c).

Not sure about how API would look like and would love to hear others thoughts on this.

The sketch may look as:

```scala
Bootstrap
  .serveFiltered[Appication.Json](authFilter, helloEndpoint)
```

This means, in addition to keeping track of endpoints in an `HList`, we also need to keep track of filters.

sergeykolbasov:
I'm curious about semantics. So the given filter should be applied to the whole `Service` output, or only to the endpoints, it was mentioned with?

prayagupd:
Hi, my use case was all of the API endpoints are secured(`api/*`), except two which are`/heartbeat` (gives API health) and `/health` (overall health including connectivity to external dependent APIs), which will be pinged by kubernetes. (For ops purpose basically)

So, the secured endpoints are intercepted with the filter and make sure the clients are authorized to access. (Using `api-secret` http header)

My current solution is using [`finagle.http.HttpMuxer`](https://twitter.github.io/finagle/docs/com/twitter/finagle/http/HttpMuxer.html),

```scala
  val heartbeat: Endpoint[String] = get("heartbeat") {
    NoContent[String]
  }

  val healthcheck: Endpoint[Map[String, Int]] = get("healthcheck") {
    Ok(
      Map(
        "service1" -> 200,
        "service2" -> 200
      )
    )
  }

 val secureEndpoints = aEndpoint :+: bEndpoint :+: cEndpoint

  val allEndpoints = HttpMuxer
    .withHandler("api/", EndpointOps.authFilter.andThen(secureEndpoints.toService))
    .withHandler("heartbeat", heartbeat.toService)
    .withHandler("health", healthcheck.toService)

Await.ready(Http.server.serve(":8080", allEndpoints))
```

I was thinking if Service could be thought as Semigroup so that they can be combined but I'm not sure. 
The `.serve` that accepts filter can be good one as well, so that above solution can work as

```scala
Bootstrap
  .serve[Appication.Json](authFilter, (aEndpoint :+: bEndpoint :+: cEndpoint))
  .serve[Application.Json](healthEndpoint)
  .serve[Text.Plain](heatbeatEndpoint)
```

sergeykolbasov:
[...]

PR #1080, opened 2019-01-26
Introduce Endpoint.Compiled, a Kleisli middleware

sergeykolbasov:
I give https://github.com/finagle/finch/issues/1061 a try and I think it looks nice.

**Separation of concerns**
I like the separation of concerns between converting
`Endpoint[A :+: B] -> (Request => F[Response])` and
`(Request => F[Response]) -> Service[Request, Response]`, it makes it cleaner IMHO.

**Effect-aware filters**
It partially solves https://github.com/finagle/finch/issues/1051, but in the better way as filters now are effect aware. Although, nothing prevents users from sticking to the Finagle filters still.

**Safety first**
Having filter on top of a service guarantees that no `Endpoint.apply` method is executed before filters are passed. It seems impossible to achieve it on `Endpoint` level with the current signature of `Endpoint.apply(input: Input): Endpoint.Result[A]`. Check the example for explanation.

**Faster Tracing** 
As we have our own abstraction over `Service`, I decided to return `Trace` along with `Response`. At the cost of tuple allocation we are getting rid of slower context mechanism that also doesn't work correctly across async boundaries of Twitter Future and some effect `F` (see https://github.com/finagle/finch/issues/1061#issuecomment-457759336). Now tracing could be nicely solved with `Filter[F]`

Those changes solve almost all of our team problems with the latest migration to monixofinch at the cost of some small backward incompatibility on instantiating `Bootstrap`.

**Example**
```scala
val filter = new Filter[IO] {
  def predicate(req: Request): IO[Boolean] = ???
  def apply(req: Request, compiled: Endpoint.Compiled[IO]): IO[(Trace, Request)] = {
    //predicate returns F[Boolean] that will never execute
    //the underlying endpoint if predicate has failed
    predicate(req).flatMap {
      case true => compiled(req)
      case false => (Trace.empty, Response(Status.BadRequest))
    }
  }
}
val compiled: Endpoint.Compiled[IO] = Bootstrap[IO].serve(endpoint).compile
val finagleService = filter.andThen(service).toFinagleService
```

I'd be happy to get some feedback on idea and implementation.


core/src/main/scala/io/finch/Bootstrap.scala
```diff
@@ -68,7 +66,7 @@ class Bootstrap[ES <: HList, CTS <: HList](
 
   def serve[CT]: Serve[CT] = new Serve[CT]
 
-  def toService(implicit ts: ToService[ES, CTS]): Service[Request, Response] = {
+  def toService(implicit ts: ToService[F, ES, CTS]): Service[F] = {
```
sergeykolbasov: maybe rename it to `compile` instead?

core/src/main/scala/io/finch/ToService.scala
```diff
@@ -102,42 +101,23 @@ object ToService {
...
           if (opts.enableUnsupportedMediaType) respond415.orElse(respond400) else respond400
 
         private[this] val negotiateContent = isNegotiable.fold(_ => true, _ => false)
         private[this] val underlying = es.head.handle(handler)
 
-        def apply(req: Request): Future[Response] = underlying(Input.fromRequest(req)) match {
+        def apply(req: Request): F[(Trace, Response)] = underlying(Input.fromRequest(req)) match {
           case EndpointResult.Matched(rem, trc, out) if rem.route.isEmpty =>
 
-            Trace.captureIfNeeded(trc)
```
sergeykolbasov: This is backward-incompatible also. Should we stick to propagate it both through the context and return type?

core/src/main/scala/io/finch/EndpointCompiled.scala
```diff
@@ -0,0 +1,37 @@
...
+import cats.effect.{ConcurrentEffect, Effect, IO}
+import com.twitter.finagle.Service
+import com.twitter.finagle.http.{Request, Response}
+import com.twitter.util.{Future, Promise}
+
+/**
+  * A compiled endpoint that is represented as a function of `Request => F[(Trace, Response)]`
+  * with ability to convert it to `com.twitter.finagle.Service`
+  */
+abstract class EndpointCompiled[F[_]] extends (Request => F[(Trace, Response)]) { self =>
```
vkostyukov: What do you think about `EndpointCompiler[F[_]]` being just an alias for a [Kleisli](https://typelevel.org/cats/datatypes/kleisli.html)? This would allow us to reuse most of the Cats' goodness w/o needing to introduce a `Filter` abstraction.

sergeykolbasov: Yea, it's very much possible

sergeykolbasov: Except, I don't know where to put `toService` conversion then

vkostyukov: I believe we can just define it for `Kleisli[F[_], Request, Response]`.

core/src/main/scala/io/finch/Filter.scala
```diff
@@ -0,0 +1,20 @@
+package io.finch
+
+import com.twitter.finagle.http.{Request, Response}
+
+trait Filter[F[_]] extends ((Request, EndpointCompiled[F]) => F[(Trace, Response)]) {
```
rpless: I'm wondering if we want to change this name too. Not sure we want to have the naming collision with Finagle Filters.

core/src/main/scala/io/finch/EndpointCompiled.scala
sergeykolbasov: Not really.

I had to work around it with `Endpoint.run` method

sergeykolbasov:
@vkostyukov 
Hm, that's true. We could `attempt` the effect to have `(Trace, Either[Throwable, Response])`, but it's not really an exit.

core/src/main/scala/io/finch/EndpointCompiled.scala
sergeykolbasov: I would prefer not to have implicits inside of wildcard package that is imported everywhere, but choices are limited here.

core/src/main/scala/io/finch/Filter.scala
sergeykolbasov: Yep, removed in favor of `Kleisli.compose`

core/src/main/scala/io/finch/EndpointCompiled.scala
vkostyukov: I agree. Let's have it as `CompileOps` for now but keep brainstorming.

core/src/main/scala/io/finch/Compile.scala
```diff
@@ -79,72 +78,55 @@ object ToService {
...
-        private[this] val underlying = es.head.handle(handler)
-
-        def apply(req: Request): Future[Response] = underlying(Input.fromRequest(req)) match {
+  ): Compile[F, Endpoint[F, A] :: ET, CTH :: CTT] = new Compile[F, Endpoint[F, A] :: ET, CTH :: CTT] {
+    def apply(es: Endpoint[F, A] :: ET, opts: Options, ctx: Context): Endpoint.Compiled[F] = {
+      val handler = if (opts.enableUnsupportedMediaType) respond415.orElse(respond400) else respond400
+      val negotiateContent = isNegotiable.fold(_ => true, _ => false)
+      val underlying = es.head.handle(handler)
+
+      new Endpoint.Compiled[F]((req: Request) => {
```
vkostyukov: Can't we omit `new`? `Kleisli` is a case class.

core/src/main/scala/io/finch/Endpoint.scala
```diff
@@ -419,6 +419,11 @@ object Endpoint {
    */
   type Module[F[_]] = EndpointModule[F]
 
+  /**
+    * Representation of function Request => F[(Trace, Response)]
+    */
+  type Compiled[F[_]] = Kleisli[F, Request, (Trace, Response)]
```
vkostyukov: One thing we may consider with regards to `Trace` is to embed it into our `Error` model, such that if `IO` resolves into `io.finch.Error`, it will have a `trace: Trace` member.

Not exactly sure I like this, just throwing out ideas.

core/src/main/scala/io/finch/package.scala
```diff
@@ -23,4 +26,34 @@ package object finch extends Outputs with ValidationRules {
     case object MultipleItems extends RequestItem("request")
   }
 
+  implicit class CompiledOps[F[_]](compiled: Endpoint.Compiled[F]) {
```
vkostyukov: Let's chat here what others ideas we might have around this?

Just a simple class `ToService` could do it:

```scala
case class ToService(k: Endpoint.Compiled[IO]) extends Service[Req, Rep] { ... }

// Usage:
val k: Endpoint.Compiled[IO] = Endpoint[IO].const("foo").compile
val s = ToService(k)
```

By the way, what do you think about adding `compile` on `Endpoint`?

core/src/main/scala/io/finch/Endpoint.scala
sergeykolbasov: Customers might throw their own exceptions

vkostyukov: I mean, we can wrap them. Not sure it's a great way of doing this but it's workable.

core/src/main/scala/io/finch/package.scala
sergeykolbasov: Personally, I would prefer to push users toward using `Bootstrap`

vkostyukov: I agree. But we do have `toService` on an `Endpoint`. I think it wouldn't hurt to add compile as a shortcut. I don't think it's confusing whether or not people should use `Bootstrap` (no semantic changes to that).

core/src/main/scala/io/finch/Compile.scala
sergeykolbasov: No, due to the type aliasing

core/src/main/scala/io/finch/Endpoint.scala
sergeykolbasov: What about `Request => (Trace, F[Response])`? I know, it's not Kleisli anymore, which is sad but on the other hand it's minimum allocations and a guarantee of having tracing all the time

vkostyukov: I'd probably prefer us to stay within a "normal" `Kleisli` as it gives us insane composability for free.

We can also re-consider shoving `Trace` onto both `Request.ctx` and `Response.ctx` as we originally discussed.

core/src/main/scala/io/finch/Compile.scala
vkostyukov: Nm then.

core/src/main/scala/io/finch/Endpoint.scala
vkostyukov: We sort of already doing `handle` within `Compile` and pretty much convert any unhandled exceptions into `io.finch.Error` (please, correct me if I'm wrong). That said, there is nothing else can expect in `Compiled` beyond either successful `Response` or an `Error`. Why not do `Error.trace` than? Seems very reasonable to me to be honest (assuming I'm correct about users exceptions not being propagated there).

sergeykolbasov: No, I don't think so. I'm sleepy already but what I see is:

```scala
private val respond400: PartialFunction[Throwable, Output[Nothing]] = {
    case e: io.finch.Error => Output.failure(e, Status.BadRequest)
    case es: io.finch.Errors => Output.failure(es, Status.BadRequest)
  }

  private val respond415: PartialFunction[Throwable, Output[Nothing]] = {
    case e: io.finch.Error if e.getCause eq Decode.UnsupportedMediaTypeException =>
      Output.failure(e, Status.UnsupportedMediaType)
  }
```

We don't catch & wrap user exceptions here

sergeykolbasov: Honestly, more I think about it and try it locally, more I doubt the idea of
`Kleisli[F, Request, Response]`.

Having something like `Request => (Trace, F[Response])` solves the case of happy/unhappy `Trace` almost for free and cases with complex filters where you want to have both `Request` and `F[Response]` (i.e. monitoring, time measurements and distributed tracing). While `Kleisli` helps with simple situations in `compose` and `andThen`, it also brings additional complexity when we go outside the usual composition.

So I'd go with our custom `Filter` or similar name. We can boost it with cats typeclasses if it's neccesary

vkostyukov: Beyond `Trace` is there anything else that bothers you? I'm not convinced we should give up on the Kleisli idea this easy. Both of the approaches of handling taces seem pretty reasonable to me. Unless there is something else concerning about `Kleisli`, I'd probably prefer we keep discussing it.

http4s uses Kleilsi as its middleware and they seem to handle things like this.

sergeykolbasov: - Embedding user exceptions into `io.finch.Error` might lead to unexpected behavior for users when their error handlers made on top of pattern matching don't work anymore. The "best" part that they wouldn't even notice any warnings in compile time in that case.
- `F.attempt` is going to hit benchmarks for sure, also turning the type signature into a nice mess.

Could you provide the example of how the Kleisli based filter and their composition would look like? May be I just can't express it nicely

vkostyukov: I can give it a try based on my very limited exposure that Klesili. 

Auth could be a `flatMap`:

```scala
scala> val in = Kleisli.ask[cats.effect.IO, Request]
in: cats.data.Kleisli[cats.effect.IO,com.twitter.finagle.http.Request,com.twitter.finagle.http.Request] = Kleisli(cats.data.KleisliFunctions$$Lambda$5376/280682234@58b00155)

scala> val middleware = Kleisli.pure[cats.effect.IO, Request, Response](Response())
middleware: cats.data.Kleisli[cats.effect.IO,com.twitter.finagle.http.Request,com.twitter.finagle.http.Response] = Kleisli(cats.data.KleisliFunctions$$Lambda$5396/1176582514@70aa5f1)

scala> in.flatMap { req => if (req.content.length > 0) middleware else Kleisli.pure[cats.effect.IO, Request, Response](Response(Status.BadRequest)) }
res1: cats.data.Kleisli[cats.effect.IO,com.twitter.finagle.http.Request,com.twitter.finagle.http.Response] = Kleisli(cats.data.Kleisli$$$Lambda$5401/1855768135@24fff07f)
```

Something like logging would probably be `andThen`:

```scala
scala> val logging = Kleisli.ask[cats.effect.IO, Request].map { req => println(req); req }
logging: cats.data.Kleisli[cats.effect.IO,com.twitter.finagle.http.Request,com.twitter.finagle.http.Request] = Kleisli(cats.data.Kleisli$$Lambda$5403/1811915277@36350e9e)

scala> logging.andThen(middleware)
res2: cats.data.Kleisli[cats.effect.IO,com.twitter.finagle.http.Request,com.twitter.finagle.http.Response] = Kleisli(cats.data.Kleisli$$$Lambda$5401/1855768135@648c6ebb)
```

`tapWith` would allow to extend logging on responses too:

```scala
scala> middleware.tapWith { (req, rep) => println(req); println(rep); rep }

res3: cats.data.Kleisli[cats.effect.IO,com.twitter.finagle.http.Request,com.twitter.finagle.http.Response] = Kleisli(cats.data.Kleisli$$Lambda$5492/986060877@2eb34475)
```

Stats can probably be a bimap:

```scala
scala> middleware.bimap(req => ({println("start"); req}), rep => ({println("finish"); rep}))
```

vkostyukov: So exposing Kleisli isn't necessarily mean limiting options. I actually believe it's the opposite: in addition to the combinators `Kleisli` already expose, you can also drop down to a filter-like structure discussed in the description of this PR that just wraps `Kleisli`. I'm not super sure yet it should be our responsibility to determine a shape of those filter-like structures. In my mind's eye, these are application-level concerns.

I'd be curious to see the benchmark numbers for `IO[Either[Trace -> Throwable, Trace -> Response]`. I'm doubtful the penalty is that bad. The ergonomics, Kleisli provide seems to outweigh the anticipated performance hit, keeping in mind we've been always overserving in performance. I'd be happy to trade some of it for developer's productivity.

sergeykolbasov: So I've changed the signature to:
```scala
type Endpoint.Compiled[F] = Kleisli[F, Request, (Trace, Either[Throwable, Response])]
```
and made some measurements. Actually, it's not _that_ bad.

**Before** (master):
```
[info] Benchmark                                                                    Mode  Cnt       Score       Error   Units
[info] JsonAndTextNegotiatedBootstrapBenchmark.foos                                thrpt   10   32358.348 ±  1809.015   ops/s
[info] JsonAndTextNegotiatedBootstrapBenchmark.foos:·gc.alloc.rate                 thrpt   10    1609.298 ±    90.256  MB/sec
[info] JsonAndTextNegotiatedBootstrapBenchmark.foos:·gc.count                      thrpt   10      74.000              counts
[info] JsonAndTextNegotiatedBootstrapBenchmark.foos:·gc.time                       thrpt   10     143.000                  ms
[info] JsonBootstrapBenchmark.foos                                                 thrpt   10   22776.430 ±  1863.079   ops/s
[info] JsonBootstrapBenchmark.foos:·gc.alloc.rate                                  thrpt   10    1503.841 ±   123.266  MB/sec
[info] JsonBootstrapBenchmark.foos:·gc.count                                       thrpt   10      70.000              counts
[info] JsonBootstrapBenchmark.foos:·gc.time                                        thrpt   10     117.000                  ms
[info] TextBootstrapBenchmark.foos                                                 thrpt   10   64812.622 ±  6849.459   ops/s
[info] TextBootstrapBenchmark.foos:·gc.alloc.rate                                  thrpt   10    1972.065 ±   209.078  MB/sec
[info] TextBootstrapBenchmark.foos:·gc.count                                       thrpt   10      92.000              counts
[info] TextBootstrapBenchmark.foos:·gc.time                                        thrpt   10     159.000                  ms
```
**After** (this branch):
```
[info] Benchmark                                                                    Mode  Cnt       Score       Error   Units
[info] JsonAndTextNegotiatedBootstrapBenchmark.foos                                thrpt   10   31949.906 ±  4026.098   ops/s
[info] JsonAndTextNegotiatedBootstrapBenchmark.foos:·gc.alloc.rate                 thrpt   10    1577.434 ±   213.755  MB/sec
[info] JsonAndTextNegotiatedBootstrapBenchmark.foos:·gc.count                      thrpt   10      72.000              counts
[info] JsonAndTextNegotiatedBootstrapBenchmark.foos:·gc.time                       thrpt   10     147.000                  ms
[info] JsonBootstrapBenchmark.foos                                                 thrpt   10   21939.217 ±  1327.161   ops/s
[info] JsonBootstrapBenchmark.foos:·gc.alloc.rate                                  thrpt   10    1452.948 ±    88.720  MB/sec
[info] JsonBootstrapBenchmark.foos:·gc.count                                       thrpt   10      67.000              counts
[info] JsonBootstrapBenchmark.foos:·gc.time                                        thrpt   10     120.000                  ms
[info] TextBootstrapBenchmark.foos                                                 thrpt   10   65395.311 ±  3172.109   ops/s
[info] TextBootstrapBenchmark.foos:·gc.alloc.rate                                  thrpt   10    2001.061 ±    97.205  MB/sec
[info] TextBootstrapBenchmark.foos:·gc.count                                       thrpt   10      92.000              counts
[info] TextBootstrapBenchmark.foos:·gc.time                                        thrpt   10     154.000                  ms
```

sergeykolbasov:
Hm, I think I even managed to keep it fully backward-compatible on the user API surface.

So `Bootstrap` now has a default type `F[_] =:= Id` but once `.serve.apply` is called, actual effect type of served endpoint is taking a place instead.

core/src/main/scala/io/finch/package.scala
sergeykolbasov: How about `Endpoint.run` or `Endpoint.service` as the function to convert compiled endpoint into `Service`?

vkostyukov: Here is a terrible idea. The only type we can hijack in `Endpoint.Compiled[F]` is `Trace`. Turns out if we put our syntax extension under `Trace` companion object, the compiler is able to find it:

```scala
scala> import cats.data.Kleisli, cats.effect.IO
import cats.data.Kleisli
import cats.effect.IO

scala> trait Trace; object Trace { implicit class KlesliSyntax(k: Kleisli[IO, Int, (Trace, Either[Throwable, String])]) { def toService: Unit = () } }
       trait Trace; object Trace { implicit class KlesliSyntax(k: Kleisli[IO, Int, (Trace, Either[Throwable, String])]) { def toService: Unit = () } }
                                                                                                                              ^
defined trait Trace
defined object Trace

scala> val k = Kleisli.pure[IO, Int, (Trace, Either[Throwable, String])](new Trace{} -> Right("foo"))
k: cats.data.Kleisli[cats.effect.IO,Int,(Trace, Either[Throwable,String])] = Kleisli(cats.data.KleisliFunctions$$Lambda$4443/1717142056@5206bb69)

scala> k.toService
()                          ^
```

sergeykolbasov: While it kinda hides it from the users radar, I think it's quite ugly from the perspective of internal maintance and code organization 😄

core/src/main/scala/io/finch/Compile.scala
```diff
@@ -79,72 +78,57 @@ object ToService {
...
+        underlying(Input.fromRequest(req)) match {
           case EndpointResult.Matched(rem, trc, out) if rem.route.isEmpty =>
 
             Trace.captureIfNeeded(trc)
```
vkostyukov: I don't think this should be happening here as we're not dealing with Twitter Futures yet. Let's move that under `toService`.

core/src/main/scala/io/finch/Trace.scala
```diff
@@ -97,12 +97,14 @@ object Trace {
    *   Trace.capture { foo(Request()).map(_ => Trace.captured) }
    * }}}
    */
+  @deprecated("Use Bootstrap.compile together with Kleisli middleware to get an access to Trace", "0.28.0")
```
vkostyukov: Do you think we should deprecate it? These seem to do the trick for people wanting to stick with Twitter Futures and Finagle Filters. I think it's fine if we keep them for now.

core/src/main/scala/io/finch/Compile.scala
sergeykolbasov: 👍

core/src/main/scala/io/finch/Trace.scala
sergeykolbasov: fair enough

core/src/main/scala/io/finch/package.scala
vkostyukov: I agree. I still like `toService` as a name just because it's something we've been using for quite some time. Although I'm out of ideas where to define it.

sergeykolbasov:
@vkostyukov I think it should be fine by now. I'm going to add also example on the middleware as a part of this PR

joroKr21:
nice :+1:

PR #1082, opened 2019-01-29
Fs2 EncodeStream for both ConcurrentEffect and Effect monads

joroKr21:
The `ConcurrentEffect` instances are prioritized higher.
This means users can benefit from cancellation when possible
without loss of generality.

@vkostyukov this is what I had in mind with providing both versions and prioritizing them. But I don't know how to test that cancellation actually works. Any ideas?

### DIFF
--- a/fs2/src/test/scala/io/finch/fs2/Fs2StreamingSpec.scala
+++ b/fs2/src/test/scala/io/finch/fs2/Fs2StreamingSpec.scala
@@ -1,16 +1,25 @@
 package io.finch.fs2
 
 import _root_.fs2.Stream
-import cats.effect.IO
+import cats.effect.{ConcurrentEffect, Effect, IO}
 import com.twitter.io.Buf
 import io.finch._
 import org.scalatest.prop.GeneratorDrivenPropertyChecks
 
 class Fs2StreamingSpec extends FinchSpec with GeneratorDrivenPropertyChecks {
 
-  checkAll("fs2.streamBody", StreamingLaws[Stream, IO](
-    list => Stream(list:_*),
-    _.map(array => Buf.ByteArray.Owned(array)).compile.toList.unsafeRunSync()
-  ).all)
+  checkEffect[IO]
+  checkConcurrentEffect[IO]
 
+  def checkEffect[F[_]](implicit F: Effect[F]): Unit =
+    checkAll("fs2.streamBody[F[_]: Effect]", StreamingLaws[Stream, F](
+      list => Stream(list:_*),
+      stream => F.toIO(stream.map(array => Buf.ByteArray.Owned(array)).compile.toList).unsafeRunSync()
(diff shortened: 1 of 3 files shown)


joroKr21:
It works just fine on 2.12 :disappointed:

joroKr21:
I'm not too concerned about compile times. After scala/scala#7067 implicits in Scala 2.12 and 2.13 are not that expensive anymore.

I'm more concerned about:
  1. How to test cancellation actually works?
  2. How to document the behavior (although that not be as bad if we think about this as an optimization).

vkostyukov:
@joroKr21 Mind rebasing this?

fs2/src/test/scala/io/finch/fs2/Fs2StreamingSpec.scala
```diff
@@ -1,16 +1,25 @@
...
 import org.scalatest.prop.GeneratorDrivenPropertyChecks
 
 class Fs2StreamingSpec extends FinchSpec with GeneratorDrivenPropertyChecks {
 
-  checkAll("fs2.streamBody", StreamingLaws[Stream, IO](
-    list => Stream(list:_*),
-    _.map(array => Buf.ByteArray.Owned(array)).compile.toList.unsafeRunSync()
-  ).all)
+  checkEffect[IO]
+  checkConcurrentEffect[IO]
```
vkostyukov: So how this compiles? I think we need instances of `ContextSwitch[IO]` and `Timer[IO]` to summon `ConcurrentEffect[IO]`?

joroKr21: Hmm, there is a `ContextShift` instance in `MissingInstances` :thinking: 
It uses the global execution context.

vkostyukov: Oh you're right. Ignore me.

sergeykolbasov:
👍 

I just solved the `Fiber` and `bracketCase` for stream cancellation on my own and then re-checked this PR. Oops 😄 

On cancellation topic, could we get some inspiration from cats-effect laws: https://github.com/typelevel/cats-effect/blob/master/laws/shared/src/test/scala/cats/effect/IOCancelableTests.scala#L76-L89 ?

vkostyukov:
FWIW, I'm happy if we decide to add a specific cancelation test later, in the upcoming commit. Let's merge this, it's been on review for quite some time.

joroKr21:
> On cancellation test topic, could we get some inspiration from cats-effect laws: https://github.com/typelevel/cats-effect/blob/master/laws/shared/src/test/scala/cats/effect/IOCancelableTests.scala#L76-L89 ?

First we have to migrate the tests to use a `TestContext` which can `tick`.

PR #1083, opened 2019-01-29
Example on using Endpoint.Compiled

sergeykolbasov:
So I wanted to make a usability check on the new mechanism. Would be nice to get a feedback if it's the right way to compose functions.

### DIFF
--- a/examples/src/main/scala/io/finch/middleware/Main.scala
+++ b/examples/src/main/scala/io/finch/middleware/Main.scala
@@ -1,0 +1,58 @@
+package io.finch.middleware
+
+import cats.effect.IO
+import com.twitter.finagle.Http
+import com.twitter.finagle.http.{Response, Status}
+import com.twitter.util.{Await, Time}
+import io.finch._
+
+/**
+  * Small Finch hello world application serving endpoint protected by serious authentication
+  * where each request & response are also logged and measured.
+  *
+  * This is achieved using Kleisli-based middleware together with [[Bootstrap.compile]]
+  *
+  * Use the following curl commands to test it:
+  *
+  * {{{
+  *   curl -v -H "Authorization: secret" http://localhost:8081/hello
+  *   curl -V -H "Authorization: wrong" http://localhost:8081/hello
+  * }}}
+  */
+object Main extends App with Endpoint.Module[IO] {
+
+  val helloWorld: Endpoint[IO, String] = get("hello") {
+    Ok("Hello world")
+  }
+
+  def auth(compiled: Endpoint.Compiled[IO]): Endpoint.Compiled[IO] = {
+    Endpoint.Compiled[IO] {
+      case req if req.authorization.contains("secret") => compiled(req)
+      case _ => IO.pure(Trace.empty -> Right(Response(Status.Unauthorized)))
+    }
+  }
+
+  def logging(compiled: Endpoint.Compiled[IO]): Endpoint.Compiled[IO] = {
+    compiled.tapWith((req, res) => {
+      print(s"Request: $req\n")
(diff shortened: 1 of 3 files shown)


examples/src/main/scala/io/finch/middleware/Main.scala
```diff
@@ -0,0 +1,58 @@
...
+      val start = Time.now
+      compiled.map {
+        case r @ (trace, _) =>
+          print(s"Response time: ${Time.now.diff(start)}. Trace: $trace\n")
+          r
+      }.run(req)
+    }
+  }
+
+  val compiled = stats(logging(auth(Bootstrap.serve[Text.Plain](helloWorld).compile)))
```
rpless: Potentially, we could curry/compose the functions and make it a little nicer to look at. Maybe something like:
```scala
val filters = stats _ andThen logging andThen auth
val compiled = filters(Bootstrap.serve[Text.Plain](helloWorld).compile)
```
I find this a little bit easier to read, but calling the functions how it is now is also fine.

examples/src/main/scala/io/finch/middleware/Main.scala
```diff
@@ -0,0 +1,58 @@
...
+  *   curl -V -H "Authorization: wrong" http://localhost:8081/hello
+  * }}}
+  */
+object Main extends App with Endpoint.Module[IO] {
+
+  val helloWorld: Endpoint[IO, String] = get("hello") {
+    Ok("Hello world")
+  }
+
+  def auth(compiled: Endpoint.Compiled[IO]): Endpoint.Compiled[IO] = {
```
rpless: I wonder if we can define these as `val`s like the examples in the [Cats Docs](https://typelevel.org/cats/datatypes/kleisli.html#configuration). I think its fine the way it is, but I might play with that later today.

examples/src/main/scala/io/finch/middleware/Main.scala
sergeykolbasov: agree. It's possible to define functions as
```scala
val fn: Endpoint.Compiled[IO] => Endpoint.Compiled[IO]
```

So `andThen` would be available naturally.

joroKr21: I would even do `val filters = Function.chain(stats, logging, auth)`

examples/src/main/scala/io/finch/middleware/Main.scala
```diff
@@ -0,0 +1,58 @@
...
+  * This is achieved using Kleisli-based middleware together with [[Bootstrap.compile]]
+  *
+  * Use the following curl commands to test it:
+  *
+  * {{{
+  *   curl -v -H "Authorization: secret" http://localhost:8081/hello
+  *   curl -V -H "Authorization: wrong" http://localhost:8081/hello
+  * }}}
+  */
+object Main extends App with Endpoint.Module[IO] {
```
vkostyukov: @sergeykolbasov I think you're on a right track with this. I looked at what http4s is doing and they pretty much follow the same structure: given a Kleilsi, wrap it with a bonus Klaisli. Take a look at these (https://github.com/http4s/http4s/tree/master/server/src/main/scala/org/http4s/server/middleware) for more inspiration.

sergeykolbasov:
@vkostyukov @rpless shall we call it a day, merge and cut a release?

rpless:
:+1: to that plan

vkostyukov:
> @vkostyukov @rpless shall we call it a day, merge and cut a release?

Maybe wait for a Finagle Feb release? Should happen next week.

sergeykolbasov:
Sound legit

PR #1089, opened 2019-02-24
Least powerful type classes + Endpoint.transformF

sergeykolbasov:
- Replace `Effect` type class boundaries with the least powerful one wherever it's possible. Mostly with `Sync`/`Async`
- Introduce `Endpoint.transformF` for `F[A] => F[B]` transformation, similar to `Kleisli.mapF`

In combination, those two changes open very promising doors together with `Endpoint.Compiled[F]`.

As an example, there is no `Effect` instance for `StateT` monad but there is one available for `Async`. Shortly speaking, it allows to maintain state over the Kleisli compositions of compiled Endpoint, so it piggybacks on top of `F[_]`. Instead of abusing `Request.ctx` users would be able to pass and change some request local state using `State[F, State, ?]` as async type:

```scala
case class State(msg: String)

type ST[A] = StateT[IO, State, A]

val state: Endpoint[ST, State] = root.mapF(state => state.get)
val hello: Endpoint[ST, String] = get("hello" :: state) { (state: State) =>
  Ok(state.msg)
}

val finale = hello.compiled.compose[ST, State] { state =>
  StateT.set(State("foo"))
} //optionally run .mapK after to convert StateT ~> IO

finale(Request("/hello")) //there is foo Response inside
```

Also, `WriterT` might be the option too.

### DIFF
--- a/core/src/main/scala/io/finch/Endpoint.scala
+++ b/core/src/main/scala/io/finch/Endpoint.scala
@@ -97,7 +97,7 @@ trait Endpoint[F[_], A] { self =>
 
   final def mapF[B](fn: F[A] => F[B])(implicit F: Monad[F]): Endpoint[F, B] =
     new Endpoint[F, B] {
-      def apply(input: Input): Result[F, B] =
+      def apply(input: Input): Endpoint.Result[F, B] =
         self(input) match {
           case EndpointResult.Matched(rem, trc, out) =>
             EndpointResult.Matched[F, B](rem, trc, out.flatMap { o =>
--- a/core/src/main/scala/io/finch/Endpoint.scala
+++ b/core/src/main/scala/io/finch/Endpoint.scala
@@ -96,24 +96,6 @@ trait Endpoint[F[_], A] { self =>
     }
 
   /**
-    * Maps this endpoint to the given function `F[A] => F[B]`
-    */
-  final def mapF[B](fn: F[A] => F[B])(implicit F: Monad[F]): Endpoint[F, B] =
-    new Endpoint[F, B] {
-      def apply(input: Input): Endpoint.Result[F, B] =
-        self(input) match {
-          case EndpointResult.Matched(rem, trc, out) =>
-            EndpointResult.Matched[F, B](rem, trc, out.flatMap { o =>
-              o.traverse(a => fn(F.pure(a)))
-            })
-          case skipped: EndpointResult.NotMatched[F] => skipped
-        }
-
-      final override def item = self.item
-      final override def toString: String = self.toString
-    }
-
-  /**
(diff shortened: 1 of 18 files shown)


sergeykolbasov:
@joroKr21 working on it.

There are issues with implicits in `Mapper` and complex types like `StateT[F, State, ?]` I couldn't overcome last night

joroKr21:
Ah I see. I've had similar problems with ReaderT but couldn't find a good solution. Mapper is a thorny part of the Finch API.

core/src/main/scala/io/finch/Endpoint.scala
```diff
@@ -95,6 +95,24 @@ trait Endpoint[F[_], A] { self =>
       final override def toString: String = self.toString
     }
 
+  /**
+    * Maps this endpoint to the given function `F[A] => F[B]`
+    */
+  final def mapF[B](fn: F[A] => F[B])(implicit F: Monad[F]): Endpoint[F, B] =
```
vkostyukov: Is it is a common name for this kind of combinator?

We already have `transform(f: F[Output[A]] => F[Output[B]])` that looks fairly similar to this. Maybe we should use symmetric name here as well?

I feel like it should be 

 - `transform(f: F[A] => F[B])`
 - `transformOutput(f: F[Output[A]] => F[Output[B]])`

to match the current naming schema. I'm open to run this through a deprecation cycle and fix the names.

sergeykolbasov: I followed Kleisli naming here:
https://github.com/typelevel/cats/blob/89b13d6d7c0933fe5639f639a89745dcf571fc1b/core/src/main/scala/cats/data/Kleisli.scala#L33

But `transform` is also a common name there.

vkostyukov: Linked `mapF` from Kleisli is different though, it maps over an effect type, `F[_] = G[_]`.

What do you think about deprecating `transform` in this release and renaming to `transformOutput` so we can introduce a proper transform in the next milestone?

sergeykolbasov: not really a map over the _type_ but also over the value `F[B] => N[C]`.
`mapK` is for the transformation `F ~> G`

This one is actually quite important in our current projects. Could we somehow speed it up?

Ofc we could make it on our own using implicits for now.

vkostyukov: Sure. What about `transformF` for this release? We can do the renaming/deprecations later.

sergeykolbasov: Done

PR #1091, opened 2019-02-26
Update only circe-core to 0.11.1

sergeykolbasov:
circe-fs2 is 0.11.0 yet, but shouldn't be a problem


travisbrown:
Right, circe-fs2 and the core modules aren't necessarily in sync at the patch version level.

PR #1096, opened 2019-03-04
Replace references to deprecated GeneratorDrivenPropertyChecks

joroKr21:
Deprecated since scalatest 3.0.6

The drop-in replacement is `ScalaCheckDrivenPropertyChecks`.

### DIFF
--- a/core/src/test/scala/io/finch/MethodSpec.scala
+++ b/core/src/test/scala/io/finch/MethodSpec.scala
@@ -5,13 +5,13 @@ import cats.effect.IO
 import com.twitter.finagle.http.Response
 import com.twitter.util.{Future => TwitterFuture}
 import org.scalacheck.Arbitrary
-import org.scalatest.prop.GeneratorDrivenPropertyChecks
+import org.scalatestplus.scalacheck.ScalaCheckDrivenPropertyChecks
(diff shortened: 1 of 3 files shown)


travisbrown:
It's worth noting that ScalaTest 3.0.6 breaks binary compatibility, so updating now would cause serious problems for anyone using both finch-test and circe-testing (or cats-laws, etc.). There should be a 3.0.7 that fixes compat soon, and it sounds like eventually Cats 2.0 will use ScalaTest 3.1.

joroKr21:
Do you mean that incompatibility is related to `GeneratorDrivenPropertyChecks` or something else? I think `finch-test` with ScalaTest 3.0.6 hasn't been released yet so we could revert to 3.0.5 or wait for 3.0.7.

travisbrown:
@joroKr21 Yes, `GeneratorDrivenPropertyChecks` is one of the broken parts in 3.0.6, where the old class is gone (the deprecation is on an alias). I'm not sure about the timing on 3.0.7, but I think Finch should either revert to 3.0.5 (in which case this PR would need to go on hold) or wait for it.

vkostyukov:
If we merge #1101, this should be good to go?

joroKr21:
I think so yes.

PR #1098, opened 2019-03-08
streaming with Monix's Iterant

alexeygorobets:


build.sbt
```diff
@@ -18,6 +18,7 @@ lazy val iterateeVersion = "0.18.0"
 lazy val refinedVersion = "0.9.4"
 lazy val catsEffectVersion = "1.2.0"
 lazy val fs2Version =  "1.0.4"
+lazy val monixVersion = "3.0.0-RC2"
```
vkostyukov: Is there a non-RC version of Monix we can depend on?

vkostyukov:
Hey, @alexeygorobets! Can you rebase your branch and adopt a cancelation logic similar to #1082?

build.sbt
rpless: Looks like RC2 is the latest. Otherwise we'd have to depend on Monix 2.0 which doesn't have Iterant.

joroKr21:
I think this would have to wait for CE3 support in Monix: https://github.com/monix/monix/pull/1533

PR #1114, opened 2019-05-16
Upgrade to Finagle 19.4.0 #1113

### LINKED ISSUE #1113: Upgrade to Finagle 19.4.0

hderms:
fix #1113

### DIFF
--- a/core/src/main/scala/io/finch/Input.scala
+++ b/core/src/main/scala/io/finch/Input.scala
@@ -1,15 +1,14 @@
 package io.finch
 
+import java.nio.charset.{Charset, StandardCharsets}
+
 import cats.Eq
 import cats.effect.Effect
-import com.twitter.finagle.http.{Method, Request}
-import com.twitter.finagle.netty3.ChannelBufferBuf
+import com.twitter.finagle.http.{Method, Request, RequestBuilder}
 import com.twitter.io.{Buf, Reader}
-import java.nio.charset.{Charset, StandardCharsets}
-import org.jboss.netty.handler.codec.http.{DefaultHttpRequest, HttpMethod, HttpVersion}
-import org.jboss.netty.handler.codec.http.multipart.{DefaultHttpDataFactory, HttpPostRequestEncoder}
 import scala.collection.mutable.ListBuffer
 import shapeless.Witness
+
 
 /**
  * An input for [[Endpoint]] that glues two individual pieces together:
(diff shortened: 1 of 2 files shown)


core/src/main/scala/io/finch/Input.scala
```diff
@@ -67,29 +67,14 @@ final case class Input(request: Request, route: List[String]) {
...
-        .toVector
-      )
-    } else ChannelBufferBuf.Owned(req.getContent)
-
-    withBody[Application.WwwFormUrlencoded](content, StandardCharsets.UTF_8)
+     val postRequest: Request = RequestBuilder()
+       .setHeader(Fields.UserAgent, "MyBot")
+       .setHeader(Fields.Connection, HttpHeaders.Values.KEEP_ALIVE)
+       .addFormElement(params: _*)
+       .url("http://foo")
```
hderms: this is extremely regrettable as we can't construct a Request without a url. I tried a lot of other ways to accomplish this same goal of producing a buf from a form encoded request but I'm at a loss as to any way to do it cleanly, while relying on Finagle code as much as possible that doesn't involve setting a bogus url on the request

hderms:
would love to know if there was a better way to accomplish this without setting a bogus URL

core/src/main/scala/io/finch/Input.scala
vkostyukov: Oh yea, that's right. Maybe just use `http://localhost` instead?

core/src/main/scala/io/finch/Input.scala
```diff
@@ -67,29 +67,14 @@ final case class Input(request: Request, route: List[String]) {
...
-        Iterator.continually(encoder.nextChunk())
-        .takeWhile(c => !c.isLast)
-        .map(c => ChannelBufferBuf.Owned(c.getContent))
-        .toVector
-      )
-    } else ChannelBufferBuf.Owned(req.getContent)
-
-    withBody[Application.WwwFormUrlencoded](content, StandardCharsets.UTF_8)
+     val postRequest: Request = RequestBuilder()
+       .setHeader(Fields.UserAgent, "MyBot")
```
vkostyukov: Do we need these two headers? I'd imagine these makes little to no difference for local testing.

hderms: @vkostyukov You're right. I removed them.

hderms:
@vkostyukov thanks for the help and kind words. As always, Finch is an amazing project

core/src/main/scala/io/finch/Input.scala
```diff
@@ -1,16 +1,16 @@
...
 import cats.Eq
 import cats.effect.Effect
-import com.twitter.finagle.http.{Method, Request}
-import com.twitter.finagle.netty3.ChannelBufferBuf
+import com.twitter.finagle.http.{Fields, Method, Request, RequestBuilder}
 import com.twitter.io.{Buf, Reader}
-import java.nio.charset.{Charset, StandardCharsets}
-import org.jboss.netty.handler.codec.http.{DefaultHttpRequest, HttpMethod, HttpVersion}
-import org.jboss.netty.handler.codec.http.multipart.{DefaultHttpDataFactory, HttpPostRequestEncoder}
+import io.netty.handler.codec.http.HttpHeaders
```
vkostyukov: Unused import?

core/src/main/scala/io/finch/Input.scala
```diff
@@ -67,29 +67,12 @@ final case class Input(request: Request, route: List[String]) {
    */
   def withForm(params: (String, String)*): Input = {
     // TODO: Figure out way to do that w/o Netty.
```
vkostyukov: I think this comment can be removed now.

rpless:
This looks great. Thanks @hderms!

PR #1125, opened 2019-06-17
Change Sigma to Merit in Adopters list

omerzach:
Sigma is now known as Merit (and is still using Finch!)
https://www.merits.com/blog/sigma-rebrands-as-merit


rpless:
Cool! Thanks for updating this @omerzach.

Issue #1138, opened 2019-08-17
effectless endpoint kills zio fiber


Phill101:
Hi. Trying to use finch with zio.Task (through cats-interop). But I've met a problem. 
```scala
get("div" :: path[Int] :: path[Int]) { (a: Int, b: Int) =>
  Ok((a / b).toString)
} handle {
  case e: Exception => BadRequest(e)
}
```
This simple endpoint kills fiber, when we try divide by zero. `handle` block do nothing and client receive 500.
If we wrap expression in Task.effect constructor all works as expected.

Seems like it happens in Output.traverseFlatten:55 when Endpoint.mapOutputAsync:109.
In cats.IO, and monix.Task you can throw an exception in functions that you pass as an argument. But not in ZIO. There is `map` awaits for a pure function. And it seems reasonable for me.

Can we refactor this two methods? I would like to try to help

Phill101:
#1142

Phill101:
Added test. Probably it has not better name.

sergeykolbasov:
Fixed in #1142

PR #1142, opened 2019-08-23
catch possible exceptions of mapOutputs function

Phill101:
fix for #1138
Idea is catch every exception that may happen in function that provided to Endpoint#mapOutput from Mapper.


sergeykolbasov:
Thanks, @Phill101!
LGTM

/cc @vkostyukov @rpless

sergeykolbasov:
Would you like to add a spec that reproduces & covers #1138?

Phill101:
added small fix

> Would you like to add a spec that reproduces & covers #1138?
yep. In progress

Phill101:
Added test. Probably it has not better name.

sergeykolbasov:
Thanks, @Phill101! I think it's good

PR #1158, opened 2019-09-18
Add Endpoint.mapK

sergeykolbasov:
Add support for natural transformation of endpoint's effect

### DIFF
--- a/core/src/main/scala/io/finch/Endpoint.scala
+++ b/core/src/main/scala/io/finch/Endpoint.scala
@@ -1,6 +1,6 @@
 package io.finch
 
-import cats.{Alternative, Applicative, ApplicativeError, Id, Monad, MonadError}
+import cats.{~>, Alternative, Applicative, ApplicativeError, Id, Monad, MonadError}
 import cats.data._
 import cats.effect._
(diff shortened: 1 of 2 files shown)


core/src/main/scala/io/finch/Endpoint.scala
```diff
@@ -170,6 +170,21 @@ trait Endpoint[F[_], A] { self =>
       final override def toString: String = self.toString
     }
 
+  /**
+    * Transform effect of endpoint given natural transformation `F ~> G`
+    */
+  final def mapK[G[_]](nat: F ~> G): Endpoint[G, A] =
```
vkostyukov: This seems to be a generally adopted name for this (at least across cats). Although, I like fs2's naming schema for this a little bit better. They use `translate`, which I think fits a little nicer into our set of APIs. What do you think?

I'm not feeling super strong about this and open for merging this as is. Just decided to mention this other name that I found.

sergeykolbasov: My idea is that since we're relying on cats type classes and `FunctionK` type, it's natural to follow this convention

PR #1169, opened 2019-10-07
Update dependencies (except Finagle)

joroKr21:
It seems appropriate to do these updates atomically,
because we bump major versions of many interconnected libraries:

  * cats to 2.0.0
  * cats-effect to 2.0.0
  * fs2 to 2.0.1
  * circe to 0.12.2

Also we add the `-Xfatal-warnings` option to ensure deprecated
API usages are replaced.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -8,13 +8,13 @@ lazy val buildSettings = Seq(
 )
 
 lazy val twitterVersion = "19.10.0"
-lazy val circeVersion = "0.11.2"
+lazy val circeVersion = "0.12.2"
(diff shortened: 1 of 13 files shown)


joroKr21:
Hmm, didn't realize there is no circe 0.12 for Scala 2.11 :thinking:

vkostyukov:
Yeah, we're going to be in this weird state for a while. There is no Finagle for 2.13 so we can't drop 2.11 target yet. Historically, at least, we've been trying to target two scala versions.

joroKr21:
Closing in favour of #1204 - it has been too long and this PR is stale.

PR #1170, opened 2019-10-07
Use HTTPS for more resolvers

joroKr21:
Looks like we are still calling `http://dl.bintray.com/tpolecat/maven` but I'm not sure why.


vkostyukov:
This is good, thanks @joroKr21!

PR #1178, opened 2019-11-11
Add Endpoint.coproduct for summing a list of endpoints

joroKr21:
It's equivalent to `List(endpoints: _*).foldK` but a bit more convenient.

### DIFF
--- a/core/src/main/scala/io/finch/Endpoint.scala
+++ b/core/src/main/scala/io/finch/Endpoint.scala
@@ -1118,4 +1118,16 @@ object Endpoint {
    */
   def multipartAttributesNel[F[_]: Sync, A: DecodeEntity: ClassTag](name: String): Endpoint[F, NonEmptyList[A]] =
     new Attribute[F, NonEmptyList, A](name) with Attribute.NonEmpty[F, A] with Attribute.MultipleErrors[F, NonEmptyList, A]
+
+  /**
+   * Sequentially composes the given `endpoints` by using [[Endpoint!.coproduct]].
+   *
+   * The resulting endpoint will match if at least one of the provided endpoints matches.
+   * If the sequence of provided endpoints is empty, the empty endpoint is returned, which never matches.
+   *
+   * @see [[Endpoint!.coproduct]] for the exact composition semantics.
+   * @see [[Endpoint.empty]] for the semantics of the empty endpoint.
+   */
+  def coproductAll[F[_], A](endpoints: Endpoint[F, A]*): Endpoint[F, A] =
+    if (endpoints.isEmpty) empty else endpoints.reduce(_ coproduct _)
 }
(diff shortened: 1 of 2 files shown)


core/src/test/scala/io/finch/EndpointSpec.scala
```diff
@@ -305,6 +305,19 @@ class EndpointSpec extends FinchSpec {
     e2(b).remainder shouldBe Some(b.withRoute(b.route.drop(2)))
   }
 
+  it should "be greedy in terms of Endpoint.coproduct" in {
```
joroKr21: I was very lazy here. Do we need more tests?

sergeykolbasov:
👍

core/src/main/scala/io/finch/Endpoint.scala
```diff
@@ -1118,4 +1118,16 @@ object Endpoint {
...
+  /**
+   * Sequentially composes the given `endpoints` by using [[Endpoint!.coproduct]].
+   *
+   * The resulting endpoint will match if at least one of the provided endpoints matches.
+   * If the sequence of provided endpoints is empty, the empty endpoint is returned, which never matches.
+   *
+   * @see [[Endpoint!.coproduct]] for the exact composition semantics.
+   * @see [[Endpoint.empty]] for the semantics of the empty endpoint.
+   */
+  def coproduct[F[_], A](endpoints: Endpoint[F, A]*): Endpoint[F, A] =
```
vkostyukov: I think that's good to have but I'm not super sure about the name, simply b/c we already have `Endpoint.coproduct`. Is there anything else we can name this?

joroKr21: We have `Endpoint.coproduct` as an instance method but this one is on the companion object. Granted, they don't do exactly the same but are very closely related. Think of it as having `concat` on collections both as an instance and companion object method (although we only have `++`, but very close).

Anyway if you suggest another name it's fine by me.

sergeykolbasov: How about `fold` or `reduce`?

joroKr21: `fold` and `reduce` already have a specific meaning in Scala. I could suggest also `sum`, `oneOf` or `atLeastOneOf`.

vkostyukov: `coproductAll`?

sergeykolbasov: Semantically, I like `oneOf`. That's exactly what this combinator does 🤔

joroKr21: So which one should it be? I called it `coproductAll` right now. Or it could be `coproductOf` as well.

zhenwenc: Not sure if it's already discussed, but have you considered refactoring the existing `coproduct` instance method to support variable argument? Or overload with
```
final def coproduct[B >: A](other: Endpoint[F, B], rest: Endpoint[F, B]*): Endpoint[F, B]
```

It should be backward compatible. Cheers!

joroKr21: I don't like it that much. It would look like there is special significance to the first endpoint (the method reciever) when there really isn't one.

PR #1179, opened 2019-11-13
Remove an allocation of Either in decodeCirce

joroKr21:
The result is already of type `Either`.
Also simplify the implementation a bit.

The second commit adds a specialized `CirceError` type,
that forwards to `cats.Show` for printing the error message.


vkostyukov:
Looks great, thank you @joroKr21!

sergeykolbasov:
👍

PR #1193, opened 2019-12-19
docs: fix 404 link to Jackson

ComFreek:


rpless:
Great catch. Thanks @ComFreek!

PR #1204, opened 2020-02-10
Add Support for Scala 2.13 and drop support for Scala 2.11

### LINKED ISSUE #1202: Support Scala 2.13
I think the title says it all. Having artefacts published to Maven Central for Scala 2.13 would help with adoption in up-to-date code bases.

rpless:
Resolves #1202 by adding support for 2.13. We're also forced to drop support for 2.11 here because a number of dependencies that are not cross built for both. Big thanks to @sergeykolbasov for figuring the weird SAM issue that caused the Mapper to not get resolved in a bunch of cases.

Couple things about this PR:
- Introduces a bunch of warnings from Scalatest, I plan on cleaning these up in a separate PR
- I'd still like to investigate cleaning up the sbt settings, I really like how Odin is does it.
- Examples and docs cannot be cross built until Twitter Server is cross built for 2.13

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -1,6 +1,8 @@
 import ReleaseTransformations._
 import microsites.ExtraMdFileConfig
 
+parallelExecution := false
+
 lazy val buildSettings = Seq(
   organization := "com.github.finagle",
   scalaVersion := "2.12.7",
@@ -19,7 +21,7 @@ lazy val refinedVersion = "0.9.12"
 lazy val catsEffectVersion = "2.0.0"
 lazy val fs2Version =  "2.1.0"
 
-lazy val compilerOptions = Seq(
+def compilerOptions(scalaVersion: String) = Seq(
   "-deprecation",
   "-encoding", "UTF-8",
   "-feature",
@@ -30,6 +32,19 @@ lazy val compilerOptions = Seq(
   "-Ywarn-dead-code",
   "-Ywarn-numeric-widen",
   "-Xlint"
+) ++ (CrossVersion.partialVersion(scalaVersion) match {
+  case Some((2, scalaMajor)) if scalaMajor == 12 => scala212CompilerOptions
+  case Some((2, scalaMajor)) if scalaMajor == 13 => scala213CompilerOptions
+})
+
+lazy val scala212CompilerOptions = Seq(
+  "-Yno-adapted-args", 
+  "-Ywarn-unused-import",
+  "-Xfuture"
+)
+
+lazy val scala213CompilerOptions = Seq(
+  "-Wunused:imports"
 )
 
 val testDependencies = Seq(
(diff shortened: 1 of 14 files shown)


rpless:
Hrm, not sure what's up with that coverage report. I might try to re-apply these changes directly to master instead of merging master into it.

build.sbt
```diff
@@ -235,8 +235,9 @@ lazy val finch = project.in(file("."))
   .settings(libraryDependencies ++= Seq(
     "io.circe" %% "circe-generic" % circeVersion
   ))
+  // Bring back examples when Twitter Server is published for Scala 2.13.
   .aggregate(
-    core, fs2, iteratee, generic, argonaut, circe, benchmarks, test, jsonTest, examples, refined
+    core, fs2, iteratee, generic, argonaut, circe, benchmarks, test, jsonTest, /*examples,*/ refined
```
sergeykolbasov: Doesn't `examples` work just fine with `crossScalaVersions := Seq("2.12.7")` setting?

rpless: It works fine for `sbt validate` which doesn't cross build, but breaks on things like `sbt +compile` trying to fetch 2.13 dependencies. I _think_ its because its inheriting the crossVersions of the aggregate project and not respecting the child projects override.

rpless:
Ok I've managed to fix the coverage report by reapplying my changes to master instead of having the merge commit. I also cleaned up a couple unneeded things. AND because Travis released Circe 0.13 between when I posted this PR and when I got home, I was able to use that instead of the milestone build.

build.sbt
sergeykolbasov: Yeah, I see it now:

```
 Issuing a cross building command, but not all sub projects have the same cross build configuration. This could result in subprojects cross building against Scala versions that they are not compatible with. Try issuing cross building command with tasks instead, since sbt will be able to ensure that cross building is only done using configured project and Scala version combinations that are configured.
```

I wish I'd understand whatever the `Try issuing cross building command with tasks instead` means

rpless:
I think I have all of the scalacOptions figured out. We were just missing the one in the 2.13 list.

sergeykolbasov:
LGTM

sergeykolbasov:
Hi @rpless

twitter-server got released for 2.13
https://search.maven.org/artifact/com.twitter/twitter-server_2.13/20.3.0/jar

I guess we could finish this PR now and then merge it 😎

rpless:
@sergeykolbasov I agree. I just pushed something to crossbuild examples and docs to 2.13. If it builds and you think its good I'll merge.

build.sbt
```diff
@@ -1,44 +1,57 @@
 import ReleaseTransformations._
 import microsites.ExtraMdFileConfig
 
+parallelExecution := false
```
sergeykolbasov: What's up with parallel execution? Also, does this line affect anything as it's not a part of any project?

rpless: Oh, that's not needed, I'll take it out. I had in there cause one of the machines I develop on didn't quite have the horsepower to run all the tests at once and would lock up.

build.sbt
```diff
@@ -4,41 +4,52 @@ import microsites.ExtraMdFileConfig
...
   scalaVersion := "2.12.7",
-  crossScalaVersions := Seq("2.11.12", "2.12.7")
+  crossScalaVersions := Seq("2.12.7", "2.13.1")
 )
 
-lazy val twitterVersion = "19.10.0"
-lazy val circeVersion = "0.11.2"
-lazy val circeIterateeVersion = "0.12.0"
-lazy val circeFs2Version = "0.11.0"
+lazy val twitterVersion = "20.1.0"
```
sergeykolbasov: I think it should be `20.3.0`

rpless: Yep good catch.

rpless:
@sergeykolbasov I've fixed the version of Twitter Server and removed that parallelExecution directive. Build looks good.

sergeykolbasov:
LGTM 👍

Issue #1002, opened 2018-10-09
Auto-publish the website when release completes


vkostyukov:
This would also mean we need to include docs type-checking on `sbt validate`.

PR #1231, opened 2020-05-25
Fix microsite compile

rpless:
Our microsite build has been broken and outdated while we've been moving to an Effect Agnostic `Endpoint`. This PR fixes the `sbt docs/makeMicrosite` command. The code examples in the docs all compile but the text that surrounds them is still out of date. My current thinking is that we can use this as a base to fix the text, although I'm open to alternative approaches to fixing our docs. If we decide to merge this, I don't think we should publish the docs to the website until all the text has been fixed.

Some things to note / questions:
- I replaced the Server Sent Event example's use of `AsyncStream` with `Fs2`
- The latest sbt-microsites seems to replace favicons and icons with its own. I've noticed this on other microsites like [cats effect](https://typelevel.org/cats-effect/).
- Set the docs compilation tool to Tut, but we should probably consider moving to mdoc in the future
- contributing.md probably needs to get updated to no longer mention Waffle [which no longer exists](https://web.archive.org/web/20190331134811/https:/waffle.io/)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,3 +11,4 @@ local.*
 
 .bloop
 .metals
+project/metals.sbt
--- a/build.sbt
+++ b/build.sbt
@@ -181,6 +181,7 @@ lazy val docSettings = allSettings ++ Seq(
   micrositeName := "Finch",
   micrositeDescription := "Scala combinator library for building Finagle HTTP services",
   micrositeAuthor := "Vladimir Kostyukov",
+  micrositeCompilingDocsTool := WithTut,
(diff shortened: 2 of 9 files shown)


sergeykolbasov:
What I did in Odin is to compile docs on each PR, so it becomes a part of the check, as well as automatically compile and publish documentation back on GitHub during release. I guess we might onboard similar mechanism here to avoid severe discrepancy between documentation and actual code.

rpless:
I like the sound of publishing on release. It might also be related to #1002. I'm going to merge this and will look at more fixes so we can get ready for a release.

PR #1267, opened 2020-09-25
Remove scalastyle in favor of scalafmt

sergeykolbasov:
I've also added wartremover as a linter but will enable it in a separate PR.


kingslef (commented):
There [a few mentions to scalastyle](https://github.com/finagle/finch/search?q=scalastyle) but perhaps at least change the contributing instructions in https://github.com/finagle/finch/blob/master/docs/mdoc/contributing.md#write-code and https://github.com/finagle/finch/blob/master/CONTRIBUTING.md#write-code and remove the `scalastyle-config.xml`.

.scalafix.conf
```diff
@@ -0,0 +1,4 @@
+OrganizeImports {
+  groupedImports = Keep
+  coalesceToWildcardImportThreshold = 7
```
Doikor, line 3: I would use the organize imports to remove all unused imports.


```suggestion
  coalesceToWildcardImportThreshold = 7
  removeUnused = true
```

Doikor: Also it looks like the coalesce to wildcard is now breaking some stuff (importing both shapeless and cats `Id` into same scope)

PR #1268, opened 2020-09-25
user-guide.md: add a few examples and remove mentions about futures

kingslef:
Add a few examples to the user guide docs and remove the mentions about futures so that it is a bit closer to what's actually in finchx.

Also add a link to giter8 template to the index.

### DIFF
--- a/docs/mdoc/user-guide.md
+++ b/docs/mdoc/user-guide.md
@@ -162,9 +162,18 @@ requested type:
 - `path[Boolean]: Endpoint[Boolean]`
 - `path[UUID]: Endpoint[java.lang.UUID]`
 
-Each extracting endpoint has a corresponding _tail extracting_ endpoints.
+For example,
+```scala mdoc
+import io.finch._, io.finch.catsEffect._
 
-There are also tail extracting endpoints available out of the box. For example, the `strings`
+val intEndpoint = path[Int]
+
+intEndpoint(Input.get("/foo")).isMatched
+
+intEndpoint(Input.get("/123")).isMatched
+```
+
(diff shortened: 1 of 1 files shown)


docs/mdoc/index.md
```diff
@@ -53,6 +53,8 @@ object Main extends App {
 }
 ```
 
+Try also [our giter8 template](https://github.com/finch/finch.g8) to get started: `sbt new finch/finch.g8`.
```
sergeykolbasov, line 56: that one requires some update before being mentioned in a doc

docs/mdoc/user-guide.md
```diff
@@ -756,7 +776,7 @@ or `beShorterThan(n: Int)`.
 
 ### Errors
 
-An endpoint may fail (it may evaluate into a `Future.exception`) by a number of reasons: it was
+An endpoint may fail (it may evaluate into a `Throwable`) by a number of reasons: it was
```
sergeykolbasov, line 779: I think it might evaluate into `F[Throwable]`

docs/mdoc/index.md
kingslef, line 56: I didn't realize it was quite old. Perhaps merge https://github.com/finch/finch.g8/pull/2 to get it into a better shape?

kingslef: I'll drop it, but I think it would be nice to have it here once it is updated.

docs/mdoc/user-guide.md
kingslef: fixed.

PR #1407, opened 2021-09-14
[WIP] Rewrite to Scala 3

sergeykolbasov:
Done:
- Get rid of `HList` in favor of new `Tuple`
- Get rid of `Witness.T` in favor of native literal types & `ValueOf`
- Get rid of magnet pattern for `get("hello") { }` apply methods, thanks to the new fancy `@targetName`. That API looks much user-friendlier now, see `Mappable`. I had to do some repetitive work, but it's still cleaner for the users I hope.
- Drop `ToAsync` stuff that enabled Scala & Twitter Futures in `Mappable`. I don't think the HTTP library should do it anyway. If users want to use different effects in their app, they shall convert between them explicitly.
- Drop the iteratee support
- Still using Finagle & Shapeless from 2.13. We still need shapeless for coproducts, because union types are meh.

TODO:
- Fix compilation & specs outside of core
- Use the latest libraries
- Adjust core to cats-effect 3


sergeykolbasov:
Ideally, I'd move builds to Github Action, but I need support from @vkostyukov for that

spockz:
> Still using Finagle & Shapeless from 2.13. We still need shapeless for coproducts, because union types are meh.

@sergeykolbasov I haven't worked with scala 3 too much but union types got me excited. Can you share why they are "meh"?

sergeykolbasov:
@spockz Yeah, I got excited as well. Later I found out that due to their commutative nature [it's impossible to define a proper type classes derivation](https://github.com/lampepfl/dotty/issues/12246).

It's also untagged comparing to `Coproduct`, so if you have generic union: 
```scala
def test[A, B](v: A | B): Unit = v match {
  case a: A => println(a)
  case b: B => println(b)
}
```

It's pretty much useless pattern matching without macrosing all the way down due to type erasure. I've hacked some macro indeed, but there were issues when `A | A |  B` was smashed into `A | B` during macro compilation. And it is kinda unacceptable in finch logic. So `Coproduct` it is.

sergeykolbasov:
Update:
Completed the migration to CE3 and refurbished the rest of the project.

The issue I'm facing now is that circe for Scala 3 doesn't support incomplete JSON decoding, so our examples fail to compile. Trying to figure it out now.

joroKr21:
Why do we even need `Coproduct`? It can easily be replaced by `OneOf[Tuple]` and an integer.

PR #1451, opened 2022-04-05
Update twitterVersion to 22.3.0

tpetillot:


joroKr21:
I wonder what should be our approach to updating Finagle? Looks like major bumps occur rather often.
We're still on 0.x so we can do this confidently but I wonder about the future.

tpetillot:
Can't we just rely on breaking API changes?

joroKr21:
> Can't we just rely on breaking API changes?

Not sure what you mean? 🤔

joroKr21:
If a dependency makes a breaking change then by bumping that dependency we force it onto users therefore also making a breaking change.

tpetillot:
Right, forget about it. 😅

IMO, as Finagle didn't broke API used by Finch in two major version, I think we could just bump a minor (or major? as Finch is a finch layer atop Finagle, maybe it make sense...) when those API are broked and we enforce user upgrade of Finagle.

joroKr21:
Maybe we can make an exception about Finagle and only bump minor versions.
I'm not sure what is the version scheme applied by Finagle - it a major version means breaking API or not 🤔 
Anyway we're still in 0.x so we shouldn't bump 1.x unless we are ready to promise compatibility.
Maybe after upgrading to CE3 we can release 1.0

PR #1457, opened 2022-04-05
Scala steward configuration

tpetillot:
chore: use circe version for circe-fs2
---

**Motivation:**
Must be aligned

chore: add scala steward configuration

---

**Modification:**
* freeze `cats-effect` to major version 2
* freeze `fs2-core` to major version (next use CE3)
* freeze `circe-fs2` to minor version (next use fs2 > v3)

### DIFF
--- a/.scala-steward.conf
+++ b/.scala-steward.conf
@@ -1,0 +1,5 @@
+updates.pin = [
+  { groupId = "co.fs2", artifactId="fs2-core", version="2." },
+  { groupId = "io.circe", version="0.13." },
+  { groupId = "org.typelevel", artifactId="cats-effect", version="2." },
+]
--- a/build.sbt
+++ b/build.sbt
@@ -10,7 +10,6 @@ lazy val buildSettings = Seq(
 lazy val twitterVersion = "22.3.0"
 lazy val circeVersion = "0.13.0"
 lazy val circeIterateeVersion = "0.13.0-M2"
-lazy val circeFs2Version = "0.13.0"
 lazy val shapelessVersion = "2.3.9"
 lazy val catsVersion = "2.7.0"
 lazy val argonautVersion = "6.3.8"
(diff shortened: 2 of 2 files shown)


.scala-steward.conf
```diff
@@ -0,0 +1,5 @@
+updates.pin = [
+  { groupId = "co.fs2", artifactId="fs2-core", version="x." },
+  { groupId = "io.circe", artifactId="circe-fs2", version="x.y" },
+  { groupId = "org.typelevel", artifactId="cats-effect", version="x." },
```
joroKr21, line 4:
```suggestion
  { groupId = "co.fs2", artifactId="fs2-core", version="2." },
  { groupId = "io.circe", version="0.13." },
  { groupId = "org.typelevel", artifactId="cats-effect", version="2." },
```

joroKr21 (changes requested):
Another thing I wonder about is if we should support both Cats Effect 2.x and 3.x on separate branches or just aim to move to 3.x soonish.

tpetillot:
3.x soonish sound good to me (it'll motivate us to do the upgrade, and quit quill)

tpetillot:
I think we're good for a release after this one? (which is not required into the release at all)

.scala-steward.conf
```diff
@@ -0,0 +1,5 @@
+updates.pin = [
+  { groupId = "co.fs2", artifactId="fs2-core", version="2." },
+  { groupId = "io.circe", artifactId="circe-fs2", version="0.13." },
```
joroKr21, line 3: I think we need to pin all of Circe
```suggestion
  { groupId = "io.circe", version="0.13." },
```

joroKr21:
> I think we're good for a release after this one? (which is not required into the release at all)

Yeah, feel free to cut a release 👍 

> Then, I may check what's needed to upgrade to CE3

Awesome!

joroKr21 (approved):

tpetillot:
756cea1a61ae82418ba7c3413f05448d71dde402 release doesn't seems to be triggered...

PR #1466, opened 2022-04-08
chore: format on compile

tpetillot:
**Motivation:**
Trigger scalafix and scalafmt on compile

**Modification:**
* validate should check formatting before compiling (as compile now trigger formatting)
* disable format on compile when on CI environment

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -52,6 +52,8 @@ val testDependencies = Seq(
   "org.typelevel" %% "discipline-scalatest" % "2.1.5"
 )
 
+val formatOnCompile = System.getenv("CI") != "true"
+
 val baseSettings = Seq(
   libraryDependencies ++= Seq(
     "com.chuusai" %% "shapeless" % shapelessVersion,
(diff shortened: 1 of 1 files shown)


build.sbt
```diff
@@ -52,6 +52,8 @@ val testDependencies = Seq(
   "org.typelevel" %% "discipline-scalatest" % "2.1.5"
 )
 
+val formatOnCompile = !sys.env.getOrElse("CI", "false").toBoolean
```
joroKr21, line 55:
```suggestion
val formatOnCompile = System.getenv("CI") != "true"
```

joroKr21 (approved):

vkostyukov:
I have a question. Why do we need to validate at all, if compile would format the code for us?

> validate should check formatting before compiling (as compile now trigger formatting)

build.sbt
```diff
@@ -416,9 +420,9 @@ lazy val benchmarks = project
 
 val validateCommands = List(
   "clean",
-  "compile",
   "scalafix --check",
   "scalafmtCheckAll",
```
vkostyukov, line 424: Do we still need these two if compile would trigger format already?

vkostyukov:
> I have a question. Why do we need to validate at all, if compile would format the code for us?
> 
> > validate should check formatting before compiling (as compile now trigger formatting)

I think I misunderstood it. I left a better question in the comments.

build.sbt
joroKr21: Because there is no guarantee you would compile the code locally.

vkostyukov (approved):

PR #1467, opened 2022-04-08
Re-encrypt credentials so they work with github actions.

rpless:
Re-encrypting the files so they can work with github actions.


arron-green:
yay! once this is merged we can merge #1468

build/build.sh
```diff
@@ -46,4 +46,4 @@ if [[ "${GITHUB_EVENT_NAME}" == "push" ]]; then
 else
     echo "${GITHUB_EVENT_NAME} build"
     ${SBT_CMD}
-fi
+fi
```
tpetillot, line 49: revert change

rpless: Yea I had some stuff locally from testing this. Lemme fix real quick.

PR #1461, opened 2022-04-07
publish finch artifacts using github actions

arron-green:
worked with @rpless to prepare deploying artifacts with github actions.  

still needs a github actions file to trigger workflow

PR #1468, opened 2022-04-08
moves decrypt within push branch

arron-green:
finishes work left from #1461 

only decrypt release files on `push`

### DIFF
--- a/build/build.sh
+++ b/build/build.sh
@@ -3,25 +3,20 @@ set -e
 
 SBT_CMD="sbt +validate"
 
-# test openssl bits
-for FILE in \
-    secring.gpg \
-    pubring.gpg \
-    credentials.sbt \
-    deploy_key.pem \
-    ; do
-    openssl aes-256-cbc -salt -pbkdf2 -d \
-        -in "./build/${FILE}.enc" \
-        -out "local.${FILE}" \
-        -pass env:ENCRYPTION_PASSWORD
-done
-
-exit 0
-
 if [[ "${GITHUB_EVENT_NAME}" == "push" ]]; then
     SBT_CMD+=" +coverageOff +publish"
 
-
+    for FILE in \
+        secring.gpg \
+        pubring.gpg \
+        credentials.sbt \
+        deploy_key.pem \
+        ; do
+        openssl aes-256-cbc -salt -pbkdf2 -d \
+            -in "./build/${FILE}.enc" \
+            -out "local.${FILE}" \
+            -pass env:ENCRYPTION_PASSWORD
+    done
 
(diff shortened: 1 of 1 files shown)


joroKr21 (approved):

rpless:
Ok I've merged mine. Once this build passes we can merge this too and then attempt a deploy.

rpless:
Ok. We should be good to release now. @tpetillot @joroKr21 do one of you want to attempt another release?

rpless:
Uh well hold on. the master branch seems to be broken even though this passed.

rpless:
There appears to be a problem with the content of the credentials.sbt file. I'm investigating.

joroKr21:
Sure I can push a commit

joroKr21:
@rpless it didn't work: https://github.com/finagle/finch/runs/5887648393?check_suite_focus=true

rpless:
Hrm, not sure why it's getting that error. I can do some digging later today.

vkostyukov:
This the error extracted from logs:

```

[info] git@github.com: Permission denied (publickey).
[info] fatal: Could not read from remote repository.
[info] Please make sure you have the correct access rights
[info] and the repository exists.
```

vkostyukov:
FYI: I was able to log in to sonatype using the credentials in the encrypted credentials.sbt.

vkostyukov:
Ah ok, maybe it's Github? Not Sonatype? I think we might be missing a key.

vkostyukov:
I just re-added the same deploy (public) key. I have no idea why Github decided to remove them but I recall this already happened to me in the other Finagle repos. Can you folks give it a try one more time?

rpless:
Awesome thanks @vkostyukov. I'll re-try the job.

vkostyukov:
It looks like it's gpg keys now:

```
[info] gpg: no default secret key: No secret key
[info] gpg: signing failed: No secret key
[error] java.lang.RuntimeException: Failure running 'gpg --batch --pinentry-mode
```

vkostyukov:
Trying to move gpg keys under gnupg folder here: https://github.com/finagle/finch/pull/1469

sergeykolbasov:
@vkostyukov Did you set the expiration date on the public key by chance? I remember I ran into the same issue without any helpful error message to later find out that the pub key expired a couple of months ago.

rpless:
It looks like that last commit fixed and the artifacts have been published. Great find @vkostyukov.

vkostyukov:
> @vkostyukov Did you set the expiration date on the public key by chance? I remember I ran into the same issue without any helpful error message to later find out that the pub key expired a couple of months ago.

I don't remember, honestly. It was a few years ago but the build/release still chugging so I think we're good?

PR #1470, opened 2022-04-11
Rename artifacts finchx to finch

vkostyukov:
Looks like we said [Finch 0.32.1](https://github.com/finagle/finch/releases/tag/v0.32.1) was the last release where we publish both artifacts. I don't think we actually published the legacy (Twitter Future) artifacts back then so I say we just rename artifacts now and publish one more release.

### DIFF
--- a/README.md
+++ b/README.md
@@ -48,18 +48,9 @@ libraryDependencies ++= Seq(
 )
 ```
 
-Cats Effect and Twitter Futures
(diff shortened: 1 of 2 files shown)


joroKr21 (approved):
Oh no, completely forgot about that 🙈

tpetillot (approved):

PR #1472, opened 2022-04-12
Migration to cats-effect 3

tpetillot:


tpetillot:
🚫 compilation failed in scala 2.13...

build.sbt
```diff
@@ -8,15 +8,16 @@ lazy val buildSettings = Seq(
...
+lazy val circeFs2Version = "0.14.0"
 lazy val circeIterateeVersion = "0.13.0-M2"
 lazy val shapelessVersion = "2.3.9"
 lazy val catsVersion = "2.7.0"
 lazy val argonautVersion = "6.3.8"
 lazy val iterateeVersion = "0.20.0"
 lazy val refinedVersion = "0.9.28"
-lazy val catsEffectVersion = "2.5.4"
-lazy val fs2Version = "2.5.10"
+lazy val catsEffectVersion = "3.3.11"
```
vkostyukov, line 19: We used to pin CE version to 2.x in scala-steward's config. Should we unpin now?

benchmarks/src/main/scala/io/finch/benchmarks.scala
```diff
@@ -13,13 +15,18 @@ import shapeless._
...
 import java.util.concurrent.{ThreadLocalRandom, TimeUnit}
+import scala.concurrent.Future
 
 @BenchmarkMode(Array(Mode.AverageTime))
 @OutputTimeUnit(TimeUnit.NANOSECONDS)
 @Warmup(iterations = 5, time = 1, timeUnit = TimeUnit.SECONDS)
 @Measurement(iterations = 5, time = 1, timeUnit = TimeUnit.SECONDS)
 @Fork(2)
 abstract class FinchBenchmark extends Endpoint.Module[IO] {
+  implicit val dispatcher: Dispatcher[IO] = new Dispatcher[IO] {
```
vkostyukov, line 26: Isn't there an instance of a Dispatcher for IO already?

core/src/main/scala/io/finch/Bootstrap.scala
```diff
@@ -81,7 +82,7 @@ class Bootstrap[F[_], ES <: HList, CTS <: HList](
     ts.apply(endpoints, opts, ctx)
   }
 
-  def toService(implicit F: Effect[F], ts: Compile[F, ES, CTS]): Service[Request, Response] =
+  def toService(implicit F: Async[F], dispatcher: Dispatcher[F], ts: Compile[F, ES, CTS]): Service[Request, Response] =
```
vkostyukov, line 85: I have very limited knowledge of CE3, but from what I'm reading we don't need to ask for Dispatcher[F] if we already have Async[F]. Looks like we can just summon a dispatcher: https://typelevel.org/cats-effect/docs/std/dispatcher

core/src/main/scala/io/finch/Endpoint.scala
```diff
@@ -297,7 +299,8 @@ trait Endpoint[F[_], A] {
     * Consider using [[io.finch.Bootstrap]] instead.
     */
   final def toService(implicit
-      F: Effect[F],
+      F: Async[F],
+      dispatcher: Dispatcher[F],
```
vkostyukov, line 303: Same question about Dispatcher. Maybe it's enough just to have Async?

core/src/main/scala/io/finch/Endpoint.scala
```diff
@@ -696,12 +693,9 @@ object Endpoint {
     *     ...
     * }}}
     */
-  def filesystemAsset[F[_]](path: String)(implicit
-      F: Sync[F],
-      S: ContextShift[F]
-  ): Endpoint[F, Buf] = {
+  def filesystemAsset[F[_]](path: String, blockingEc: ExecutionContext)(implicit F: Async[F]): Endpoint[F, Buf] = {
```
vkostyukov, line 696: Should we take implicit EC?

core/src/main/scala/io/finch/EndpointResult.scala
```diff
@@ -40,32 +40,29 @@ sealed abstract class EndpointResult[F[_], +A] {
     case _                                 => None
   }
 
-  def awaitOutput(d: Duration = Duration.Inf)(implicit F: Effect[F]): Option[Either[Throwable, Output[A]]] = this match {
+  def awaitOutput(d: Duration = Duration.Inf)(implicit dispatcher: Dispatcher[F]): Option[Either[Throwable, Output[A]]] = this match {
```
vkostyukov, line 43: I think we should ask for Async[F] and summon Dispatcher instead.

Something along the lines:

```
Dispatcher[F].use { dispatcher =>
case EndpointResult.Matched(_, _, out) =>
      try Some(Right(dispatcher.unsafeRunTimed(out, d)))
      catch {
        case _: TimeoutException => Some(Left(new TimeoutException(s"Output wasn't returned in time: $d")))
        case e: Throwable        => Some(Left(e))
      }
}
```

core/src/main/scala/io/finch/Input.scala
```diff
@@ -103,17 +103,17 @@ object Input {
       Input(copied, i.route)
     }
 
-    def apply[F[_]: Effect, S[_[_], _], A](s: S[F, A])(implicit
+    def apply[F[_]: Dispatcher, S[_[_], _], A](s: S[F, A])(implicit
```
vkostyukov, line 106: Same question here:

I think we should keep asking for Async type-class and derive Dispatcher out of it later.

core/src/test/scala/io/finch/EndToEndSpec.scala
```diff
@@ -36,6 +36,7 @@ class EndToEndSpec extends FinchSpec {
     implicit val encodeException: Encode.Text[Exception] =
       Encode.text((_, cs) => Buf.ByteArray.Owned("ERR!".getBytes(cs.name)))
 
+    println("I'M GONNA SERVE THIS!")
```
vkostyukov, line 39: Don't forget to remove this.

core/src/test/scala/io/finch/EndpointSpec.scala
```diff
@@ -438,6 +448,11 @@ class EndpointSpec extends FinchSpec {
         def apply[A](fa: IO[A]): WriterT[IO, List[String], A] = WriterT.liftF(fa)
       }
 
+      implicit val dispatcherW: Dispatcher[W] = new Dispatcher[W] {
```
vkostyukov, line 451: Same: there might be already an instance of `Async` for Writer so we can summon Dispatcher.

core/src/test/scala/io/finch/MissingInstances.scala
```diff
@@ -1,21 +1,27 @@
...
 import com.twitter.io.Buf
 
-import scala.concurrent.ExecutionContext
+import scala.concurrent.Future
 
 /** Type class instances for non-Finch types.
   */
 trait MissingInstances {
+
+  implicit val dispatcherIO: Dispatcher[IO] = new Dispatcher[IO] {
```
vkostyukov, line 15: We probably wouldn't need to do that if we defer to `Dispatcher[IO].use` API.

examples/src/main/scala/io/finch/middleware/Main.scala
```diff
@@ -19,7 +20,7 @@ import io.finch._
   *
   * !Disclaimer: most likely you would need to use proper libraries for logging, auth, and metrics instead but their use will be quite similar.
   */
-object Main extends App with Endpoint.Module[IO] {
+object Main extends IOApp with Endpoint.Module[IO] {
```
vkostyukov, line 23: Oh that's very cool. We should do that for other examples too if we can.

examples/src/main/scala/io/finch/wrk/Finch.scala
```diff
@@ -16,5 +19,10 @@ import io.finch.circe._
   *   - c = t * n * 1.5
   */
 object Finch extends Wrk {
-  serve(Endpoint[IO].lift(Payload("Hello, World!")).toServiceAs[Application.Json])
+
+  implicit val ioRunTime: IORuntime = IORuntime.global
+
+  Dispatcher[IO].use { implicit dispatcher =>
```
vkostyukov, line 25: We wouldn't need to do that if stop requiring an implicit Dispatcher. It would be easier to interact with Endpoint API that way.

vkostyukov (changes requested):
This is very cool, @tpetillot!

I'm totally onboard to drop CE2 and move to CE3, but I think we'd need to figure out the proper use of a `Dispatcher` utility, which feels a little out of order now.

benchmarks/src/main/scala/io/finch/benchmarks.scala
joroKr21: Yes there is but it returns a `Resource`

core/src/main/scala/io/finch/Endpoint.scala
joroKr21: `Async` already has a blocking EC baked in

core/src/main/scala/io/finch/EndpointResult.scala
joroKr21: Wouldn't that change the signature?

core/src/main/scala/io/finch/Bootstrap.scala
tpetillot: I'm not very confident about this... Looking at `Dispatcher[IO]`, it seems to me that instantiation is expensive, but at the same time doesn't seem to be that much.

Looking forward, I've sturborned across a discussion between [Fabio Labella and @joroKr21 on Gitter ](https://gitter.im/typelevel/cats-effect?at=5fbebdbafe857d71dc953fe3)on the same concern, where it's stated that Dispatcher instantiation:

> isn't that heavyweight either, just a Fiber and some AtomicReferences
> it's very very lightweight in fact

tpetillot: But no matter, dispatcher instanciation return a `Resource[F, Dispatcher[F]]`, which we cannot run synchronously on its own...

core/src/main/scala/io/finch/Endpoint.scala
tpetillot: @joroKr21 thanks, completly missed it!

build.sbt
tpetillot: Agree, removed!

core/src/main/scala/io/finch/Bootstrap.scala
vkostyukov: I guess that's my point exactly? Shouldn't we `use` that resource instead of accessing Dispatcher directly? I think we're side-stepping the API for some reason.

This is the use method I'm referring to: https://github.com/typelevel/cats-effect/blob/series/3.x/kernel/shared/src/main/scala/cats/effect/kernel/Resource.scala#L206

joroKr21: `use` also would change the signature to `IO`

vkostyukov: Ok, I see. I'm reading this doc (https://typelevel.org/cats-effect/docs/migration-guide) and it, at least, says that we should use Dispatcher as a type-class, b/c it's not one:

> An alternative approach to creating a Dispatcher on our own would be taking one as a parameter. It is recommended to pass it explicitly, as Dispatcher isn't a type class and there can be many instances of it in an application.

We somehow need to bubble Dispatcher construction up, I assume. Should it maybe happen as part of a Bootstrap? Like Bootstrap.toService would return you a `Resource[Service]` with `Dispatcher` fused in?

I understand we also need to figure out how to run those EndpointResult methods since they are happening outside of Bootstrap. Maybe for those we should use ask for Dispatcher explicitly and expect users to instantiate one themselves? Maybe it's not too much trouble since it's only happening in tests.

tpetillot: Updated accordingly

core/src/test/scala/io/finch/BodySpec.scala
```diff
@@ -36,29 +36,29 @@ class BodySpec extends FinchSpec {
   it should "not match on streaming requests" in {
     val req = Request()
     req.setChunked(true)
-    body[Foo, Text.Plain].apply(Input.fromRequest(req)).awaitValueUnsafe() shouldBe None
+    body[Foo, Text.Plain].apply(Input.fromRequest(req)).awaitValueUnsafe(dispatcherIO) shouldBe None
```
vkostyukov, line 39: Maybe CE already provides some kind of test base class that instantiates a dispatcher on before and shuts it down on after?

examples/src/main/scala/io/finch/iteratee/Main.scala
```diff
@@ -63,13 +65,14 @@ object Main extends IOApp {
...
+    Resource.make(IO(Http.server.withStreaming(enabled = true).serve(":8081", service))) { server =>
+      IO.defer(implicitly[ToAsync[Future, IO]].apply(server.close()))
+    }
 
-    server.use(_ => IO.never).as(ExitCode.Success)
-  }
+  def run(args: List[String]): IO[ExitCode] =
+    (for {
+      service <- (sumJson :+: streamJson :+: isPrime).toServiceAs[Application.Json]
+      server <- serve(service)
```
vkostyukov, line 76: This structure makes so much sense now, I love it!

examples/src/main/scala/io/finch/middleware/Main.scala
```diff
@@ -52,6 +53,14 @@ object Main extends App with Endpoint.Module[IO] {
   val filters = Function.chain(Seq(stats, logging, auth))
   val compiled = filters(Bootstrap.serve[Text.Plain](helloWorld).compile)
 
-  Await.ready(Http.server.serve(":8081", Endpoint.toService(compiled)))
+  def serve(service: Service[Request, Response]): Resource[IO, ListeningServer] =
+    Resource.make(IO(Http.server.serve(":8081", service))) { server =>
```
vkostyukov, line 57: An idea for another PR: maybe we should make this function a utility and provide our wrapper around IOApp.

examples/src/main/scala/io/finch/todo/Main.scala
```diff
@@ -31,21 +32,28 @@ object Main extends TwitterServer {
...
+      id <- Resource.eval(Ref[IO].of(0))
+      store <- Resource.eval(Ref[IO].of(Map.empty[Int, Todo]))
+      app = new App(id, store)
+      service <- app.toService
+      server <- serve(service)
+    } yield server).use(_ => IO(Await.ready(adminHttpServer)) >> IO.never)
 
-      srv.serve(s":${port()}", app.toService)
+  def main(): Unit = {
+    val dispatcher = new Dispatcher[IO] {
```
vkostyukov, line 54: Can we do this instead?

```
val run = Dispatcher[IO].use { dispathcer =>
  // run on the things
}

// since run is IO
run.unsafeRunSync()
```

vkostyukov (commented):
This looks pretty awesome! Thank you for following through the dispatcher changes, @tpetillot!

I think there is a few places left that we might need to clean up, but we're almost there. I want to look around the CE3 APIs today to see if we're missing anything else. I like the use of Resouce[F, Service[_, _]] already but perhaps we'd need to spend a little more thinking to make sure we don't compromise the testing experience for users.

A couple of things I want to find answers to:
 - In tests (our own and users), can we somehow avoid instantiating our own dispatcher
 - Can we make experience of EndpointResult.await methods better?

examples/src/main/scala/io/finch/iteratee/Main.scala
```diff
@@ -63,13 +65,14 @@ object Main extends IOApp {
...
+    } yield server).use(_ => IO.never[ExitCode])
```
joroKr21, line 77:
```suggestion
    } yield server).useForever
```

examples/src/main/scala/io/finch/middleware/Main.scala
```diff
@@ -52,6 +53,14 @@ object Main extends App with Endpoint.Module[IO] {
...
+      IO.defer(implicitly[ToAsync[Future, IO]].apply(server.close()))
+    }
 
+  override def run(args: List[String]): IO[ExitCode] =
+    (for {
+      service <- Endpoint.toService(compiled)
+      server <- serve(service)
+    } yield server).use(_ => IO.never[ExitCode])
```
joroKr21, line 65:
```suggestion
    } yield server).useForever
```

examples/src/main/scala/io/finch/wrk/Finagle.scala
```diff
@@ -5,11 +5,12 @@ import com.fasterxml.jackson.module.scala.DefaultScalaModule
 import com.twitter.finagle.Service
 import com.twitter.finagle.http.{Request, Response, Status}
 import com.twitter.io.Buf
-import com.twitter.util.Future
+import com.twitter.util.{Await, Future}
 import io.finch.internal._
 
 /** How to benchmark this:
   *
+  *   1. Run the server: sbt 'examples/runMain io.finch.wrk.Finagle' 2. Run wrk: wrk -t4 -c24 -d30s http://localhost:8081/
```
joroKr21, line 13: This line is a duplicate

examples/src/main/scala/io/finch/wrk/Finch.scala
```diff
@@ -1,12 +1,18 @@
...
+import com.twitter.finagle.{ListeningServer, Service}
+import com.twitter.util.Future
 import io.circe.generic.auto._
 import io.finch._
 import io.finch.circe._
+import io.finch.internal.ToAsync
 
 /** How to benchmark this:
   *
+  *   1. Run the server: sbt 'examples/runMain io.finch.wrk.Finch' 2. Run wrk: wrk -t4 -c24 -d30s http://localhost:8081/
```
joroKr21, line 15: Delete this line

examples/src/main/scala/io/finch/wrk/Finch.scala
```diff
@@ -15,6 +21,18 @@ import io.finch.circe._
...
+  def serveR(service: Service[Request, Response]): Resource[IO, ListeningServer] =
+    Resource.make(IO(serve(service))) { server =>
+      IO.defer(implicitly[ToAsync[Future, IO]].apply(server.close()))
+    }
+
+  override def run(args: List[String]): IO[ExitCode] =
+    (for {
+      service <- Endpoint[IO].lift(Payload("Hello, World!")).toServiceAs[Application.Json]
+      server <- serveR(service)
+    } yield server).use(_ => IO.never[ExitCode])
```
joroKr21, line 37:
```suggestion
    } yield server).useForever
```

examples/src/main/scala/io/finch/wrk/Finch.scala
```diff
@@ -15,6 +21,18 @@ import io.finch.circe._
   *   - t = n
   *   - c = t * n * 1.5
   */
-object Finch extends Wrk {
-  serve(Endpoint[IO].lift(Payload("Hello, World!")).toServiceAs[Application.Json])
+object Finch extends IOApp with Wrk {
+
+  implicit val ioRunTime: IORuntime = IORuntime.global
```
joroKr21, line 26: Why do we need it? If we use `IOApp`?

joroKr21 (changes requested):
This looks much better than I thought it would.
See my comments for some improvements, but overall amazing work 👍

examples/src/main/scala/io/finch/middleware/Main.scala
vkostyukov: Wow, sweet!

core/src/test/scala/io/finch/BodySpec.scala
vkostyukov: Thinking out loud: what if we give up providing this awaitX method on EndpointResult and maybe move that responsibility to users? If you think about it, it's not too bad of an experience after all, given that F is already defined as IO. Folks can just can runSyncUnsafe  (or whatever it's called) on their IO instances in tests directly. We could still make this experience nicer if we provide convenient functions transforming the output, but running it should be something we leave our users to deal with.

joroKr21: That's surely an option - but we would have to change our own tests quite a bit. I would leave that for another PR.

core/src/test/scala/io/finch/FinchSpec.scala
```diff
@@ -284,17 +285,18 @@ trait FinchSpec extends AnyFlatSpec with Matchers with Checkers with AllInstance
       (r.method.toString, r.version.toString, r.path, Buf.ByteArray.Owned.extract(r.content))
     }
 
-  implicit def arbitraryEndpoint[F[_]: Effect, A](implicit A: Arbitrary[A]): Arbitrary[Endpoint[F, A]] = Arbitrary(
+  implicit def arbitraryEndpoint[F[_]: Sync, A](implicit A: Arbitrary[A]): Arbitrary[Endpoint[F, A]] = Arbitrary(
     Gen.oneOf(
       Gen.const(Endpoint[F].empty[A]),
       A.arbitrary.map(a => Endpoint[F].const(a)),
-      Arbitrary.arbitrary[Throwable].map(e => Endpoint[F].liftOutputAsync(Effect[F].raiseError[Output[A]](e))),
+      // FIXME: use exception instead of throwable, as dispatcher use promise, wrapping error into ExecutionException (see unsafeToFutureCancelable)
```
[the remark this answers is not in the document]
tpetillot: Dispatcher use promise here: https://github.com/typelevel/cats-effect/blob/series/3.x/std/shared/src/main/scala/cats/effect/std/Dispatcher.scala#L194
And promise box error here: https://github.com/scala/scala/blob/2.12.x/src/library/scala/concurrent/impl/Promise.scala#L87

Breaking test when `Arbitrary.arbitrary[Throwable]` throw an `Error`.

core/src/main/scala/io/finch/endpoint/endpoint.scala
```diff
@@ -11,23 +11,20 @@ import java.io.InputStream
...
-      val n = stream.read(buffer)
-      if (n == -1) F.pure(left)
-      else readLoop(left.concat(Buf.ByteArray.Owned(buffer, 0, n)), stream)
-    }
+    private def readLoop(left: Buf, stream: InputStream): F[Buf] =
+      for {
+        buffer <- F.delay(new Array[Byte](1024))
+        n <- F.blocking(stream.read(buffer))
+        buf <- if (n == -1) F.pure(left) else readLoop(left.concat(Buf.ByteArray.Owned(buffer, 0, n)), stream)
+      } yield buf
```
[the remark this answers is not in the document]
tpetillot: Nice catch! Added defer for stack safety.

core/src/test/scala/io/finch/internal/ToEffectSpec.scala
```diff
@@ -10,6 +11,8 @@ import scala.concurrent.{Future => ScalaFuture}
 
 class ToEffectSpec extends FinchSpec {
 
+  implicit val ioRuntime: IORuntime = IORuntime.global
```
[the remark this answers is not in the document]
tpetillot: Used by `unsafeRunSync()`, for a reason I don't understand, when importing `cats.effect.unsafe.implicits.global` implicit is not resolved. 😞

examples/src/main/scala/io/finch/middleware/Main.scala
tpetillot: Definitely! Is there any place to add feature request?

core/src/main/scala/io/finch/Input.scala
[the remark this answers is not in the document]
tpetillot: Updated to take it explicitly

benchmarks/src/main/scala/io/finch/benchmarks.scala
```diff
@@ -190,11 +197,13 @@ abstract class BootstrapBenchmark[CT](init: Bootstrap[Id, HNil, HNil])(implicit
 
   protected def issueRequest(): Request = Request()
 
-  private val foo: Service[Request, Response] =
+  private val foo: Resource[IO, Service[Request, Response]] =
     init.serve[CT](Endpoint[IO].const(List.fill(128)(Foo(scala.util.Random.alphanumeric.take(10).mkString)))).toService
 
   @Benchmark
-  def foos: Response = Await.result(foo.apply(issueRequest()))
+  def foos: Response = Await.result(dispatcher.unsafeRunSync(foo.use { service =>
```
vkostyukov, line 204: I think it's ok if we keep this as is for this change, but we'd need to revisit this sometime later. Ideally, we'd need to `use` the resource on each benchmark invocation. Maybe we could leverage before/after setup machine to manage the lifecycle of the resource.

core/src/test/scala/io/finch/BodySpec.scala
vkostyukov: Yea I like your plan.

core/src/main/scala/io/finch/Input.scala
```diff
@@ -103,17 +103,16 @@ object Input {
       Input(copied, i.route)
     }
 
-    def apply[F[_]: Effect, S[_[_], _], A](s: S[F, A])(implicit
+    def apply[F[_], S[_[_], _], A](s: S[F, A], dispatcher: Dispatcher[F])(implicit
```
vkostyukov, line 106: This is another place that I'm not in love with but we should leave it for another PR. I don't have a ton of ideas just yet but I'm thinking to extract this and awaitX methods into their own testing package (or syntax?) that's hopefully structured around the existing dispatcher.

core/src/test/scala/io/finch/FinchSpec.scala
```diff
@@ -304,12 +306,12 @@ trait FinchSpec extends AnyFlatSpec with Matchers with Checkers with AllInstance
     *
     * We attempt to verify that two endpoints are the same by applying them to a fixed number of randomly generated inputs.
     */
-  implicit def eqEndpoint[F[_]: Effect, A: Eq]: Eq[Endpoint[F, A]] = new Eq[Endpoint[F, A]] {
+  implicit def eqEndpoint[F[_]: Sync, A: Eq](implicit dispatcher: Dispatcher[F]): Eq[Endpoint[F, A]] = new Eq[Endpoint[F, A]] {
```
vkostyukov, line 309: Oh this eq methods need more attention. Maybe we could move these under missingInstances that's instantiated around the existing dispatcher.

examples/src/main/scala/io/finch/middleware/Main.scala
vkostyukov: Maybe just file a Github issue for now?

examples/src/main/scala/io/finch/todo/Main.scala
```diff
@@ -31,21 +31,23 @@ object Main extends TwitterServer {
...
-  }
+  val run: IO[Unit] =
+    (for {
+      _ <- Resource.eval(IO.println(s"Open your browser at http://localhost:${port()}/todo/index.html")) // scalastyle:ignore
+      id <- Resource.eval(Ref[IO].of(0))
+      store <- Resource.eval(Ref[IO].of(Map.empty[Int, Todo]))
+      app = new App(id, store)
+      service <- app.toService
+      server <- serve(service)
+    } yield server).use(_ => IO(Await.ready(adminHttpServer)) >> IO.never)
```
vkostyukov, line 50: Should we drop IO.never here? I think it we just keep waiting on `Await.ready`, the app would correctly shutdown when /quitquitqui endpoint is hit.

vkostyukov (approved):
This looks pretty nice @tpetillot!

I think I'm ok merging this PR in the current form but keep iterating on the testing pieces. We'd probably want to be good citizens of the CE3 API and stop instantiating our own dispatchers.

I'd wait a ship it from @joroKr21 before merging. And please, use squash commits when merging so we keep the git history more or less clean.

joroKr21 (approved):
I think we can merge and think a bit more about dispatcher and testing.

core/src/test/scala/io/finch/FinchSpec.scala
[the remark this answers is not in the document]
joroKr21: Ok weird, TIL

core/src/test/scala/io/finch/FinchSpec.scala
tpetillot: Existing dispatcher is already set on IO (and eq is used based on higher kind). I'll let it that way until next PR 🚀

tpetillot:
@joroKr21 @vkostyukov merged 🎊 
Looking forward next propositions regarding dispatching (I will try to look into it a bit more as well).

In anticipation of our future upgrades to CE3, when would you estimate a first release candidate?

vkostyukov:
I was thinking we do a couple more changes:
 - testing and dispatcher
 - #1471 
 
 And do a RC1 for 1.0. What do you both think about this plan?

joroKr21:
> And do a RC1 for 1.0. What do you both think about this plan?

Sounds good 👍

PR #1478, opened 2022-04-19
Testing EndpointResults without Dispatcher

vkostyukov:
I only did `BodySpec` to get an idea of how does it look and feel. I will carry on if folks are ok with this approach.


core/src/main/scala/io/finch/EndpointResult.scala
```diff
@@ -86,9 +59,19 @@ object EndpointResult {
...
-    final def output: Option[F[Output[A]]] = self match {
-      case EndpointResult.Matched(_, _, out) => Some(out)
-      case _                                 => None
+    final def outputAttempt(implicit F: ApplicativeError[F, Throwable]): F[Either[Throwable, Output[A]]] =
+      F.attempt(output)
+
+    final def output: F[Output[A]] = self match {
+      case EndpointResult.Matched(_, _, out) => out
+      case _                                 =>
+        throw new IllegalStateException("Endpoint didn't match")
```
tpetillot, line 68: I would rather raise an error than throw it.
```suggestion
      case _                                 => F.raiseError(new IllegalStateException("Endpoint didn't match"))
```

tpetillot (commented):
I really like what you are proposing!

joroKr21:
I like the approach too 👍 - much simpler!

joroKr21:
What about `EntityEndpointLaws`? It looks like it's not so easy.

tpetillot:
If you still want to define a rule set, either we still pass the dispatcher on this one, or, we defined `EntityEndpointLaws` on `IO` as all spec is defined on IO?

PR #1480, opened 2022-04-20
Remove ToAsync

vkostyukov:
We shouldn't be in the business of magically converting Scala/Twitter Futures to IO (or Async[F]).


joroKr21 (approved):
Oh I see we used `ToAsync` to support endpoints that return `Future`.
If we want to bring back this support we could just add specialised mappers.
But maybe we don't need that. Good riddance.

tpetillot (approved):

PR #1481, opened 2022-04-21
Clean up examples to not use toAsync

vkostyukov:
We can just use IO.delay and avoid dealing with TwitterFuture to IO conversions.

### DIFF
--- a/core/src/main/scala/io/finch/Bootstrap.scala
+++ b/core/src/main/scala/io/finch/Bootstrap.scala
@@ -2,21 +2,19 @@ package io.finch
 
 import cats.effect.std.Dispatcher
 import cats.effect.{Async, Resource}
-import com.twitter.finagle.Filter
-import com.twitter.finagle.Http
-import com.twitter.finagle.ListeningServer
 import com.twitter.finagle.http.{Request, Response}
+import com.twitter.finagle.{Filter, Http, ListeningServer, Service}
 import io.finch.internal.TwitterFutureConverter
 import shapeless._
 
-/** Bootstraps a Finagle HTTP service out of the collection of Finch endpoints.
+/** Bootstraps a Finagle HTTP listening server out of the collection of Finch endpoints.
   *
   * {{{
-  * val api: Service[Request, Response] = Bootstrap
-  *  .configure(negotiateContentType = true, enableMethodNotAllowed = true)
+  * val api: Service[Request, Response] = Bootstrap[F]
+  *  .configure(includeServerHeader = false, enableMethodNotAllowed = true)
   *  .serve[Application.Json](getUser :+: postUser)
   *  .serve[Text.Plain](healthcheck)
-  *  .toService
+  *  .listen(":80")
   * }}}
   *
   * ==Supported Configuration Options==
@@ -39,14 +37,14 @@ import shapeless._
   *   https://tools.ietf.org/html/rfc7231#section-6.5.13
   */
 class Bootstrap[F[_], ES <: HList, CTS <: HList](
-    val endpoints: ES,
-    val server: Http.Server = Http.server,
-    val filter: Filter[Request, Response, Request, Response] = Filter.identity[Request, Response],
-    val middleware: Endpoint.Compiled[F] => Endpoint.Compiled[F] = identity _,
-    val includeDateHeader: Boolean = true,
-    val includeServerHeader: Boolean = true,
-    val enableMethodNotAllowed: Boolean = false,
-    val enableUnsupportedMediaType: Boolean = false
+    endpoints: ES,
+    server: => Http.Server = Http.server,
+    filter: Filter[Request, Response, Request, Response] = Filter.identity[Request, Response],
+    middleware: Endpoint.Compiled[F] => Endpoint.Compiled[F] = identity[Endpoint.Compiled[F]] _,
+    includeDateHeader: Boolean = true,
+    includeServerHeader: Boolean = true,
+    enableMethodNotAllowed: Boolean = false,
+    enableUnsupportedMediaType: Boolean = false
 ) { self =>
 
   class Serve[CT] {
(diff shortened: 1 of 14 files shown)


joroKr21:
I don't think that's safe in general - we have to wait for `close` to return

vkostyukov:
> I don't think that's safe in general - we have to wait for `close` to return

I think you're right. I reworked a few things around to make Bootstrap take care of closing the server. This maybe gives Bootstrap too much power so I'm not super sure about this approach yet. Although, I do like the end-user experience working with it, especially where there is no middleware involved.

Take a look at the examples and let me know what do you think about this approach?

core/src/main/scala/io/finch/Bootstrap.scala
```diff
@@ -81,17 +118,27 @@ class Bootstrap[F[_], ES <: HList, CTS <: HList](
...
+
+    val serviced = Dispatcher[F].flatMap { dispatcher =>
+      Resource.make(F.pure(ToService(compiled, dispatcher))) { service =>
+        F.defer(service.close().toAsync[F])
+      }
+    }
+
+    serviced.flatMap { service =>
+      val filtered = filter.andThen(service)
+      Resource.make(F.pure(server.serve(address, filtered))) { listening =>
```
tpetillot, line 132: It is an elegant solution, but it seems to me that there is no input for server configurations. (e.g: `withX` methods on `com.twitter.finagle.Http.Server`)

vkostyukov: I think what I envisioned is that a final version of the server is passed to Bootstrap.apply. A few of our examples leverage that already (e.g, the streaming one).

tpetillot: Oh, I missed it! 👍

tpetillot (commented):
Seems super handy, LGTM!

joroKr21:
I think it's good to make it easier to use a `Resource`-based app but we shouldn't make it harder to have a custom `main`.

vkostyukov:
> I think it's good to make it easier to use a `Resource`-based app but we shouldn't make it harder to have a custom `main`.

The custom main as in Todo app? Can you elaborate?

tpetillot:
@vkostyukov I think it's about letting the compilation public

joroKr21:
I didn't mean anything specific - just a thought.

tpetillot:
FYI, we released an internal version of this branch and use it in production without any problem since June. @joroKr21

core/src/test/scala/io/finch/EndToEndSpec.scala
```diff
@@ -32,20 +33,24 @@ class EndToEndSpec extends FinchSpec {
     "text/event-stream"
   )
 
-  private def testService[E](service: Resource[IO, Service[Request, Response]])(assertions: Service[Request, Response] => E): E =
-    dispatcherIO.unsafeRunSync(service.use(s => IO(assertions(s))))
+  private object ServiceTest {
+    def apply[CTS] = new {
```
joroKr21, line 37: This should be `AnyVal`

examples/src/main/scala/io/finch/todo/App.scala
```diff
@@ -53,7 +53,7 @@ class App(idRef: Ref[IO, Int], storeRef: Ref[IO, Map[Int, Todo]]) extends Endpoi
     }
   }
 
-  final def toService: Resource[IO, Service[Request, Response]] = Bootstrap
+  final val toService: Resource[IO, Service[Request, Response]] = Bootstrap[IO]
```
joroKr21, line 56:
```suggestion
  final val service: Resource[IO, Service[Request, Response]] = Bootstrap[IO]
```

examples/src/main/scala/io/finch/todo/Main.scala
```diff
@@ -27,28 +29,37 @@ import io.finch.internal._
...
   private val port: Flag[Int] = flag("port", 8081, "TCP port for HTTP server")
 
-  private def serve(service: Service[Request, Response]): Resource[IO, ListeningServer] =
-    Resource.make(for {
-      srv <- IO(Http.server.withStatsReceiver(statsReceiver))
-      server <- IO(srv.serve(s":${port()}", service))
-    } yield server) { server =>
-      // TODO: We should abstract this out.
-      IO.defer(server.close().toAsync[IO])
+  def app: IO[App] =
```
joroKr21, line 36:
```suggestion
  val app: IO[App] =
```

examples/src/main/scala/io/finch/todo/Main.scala
```diff
@@ -27,28 +29,37 @@ import io.finch.internal._
...
+    for {
+      id <- Ref[IO].of(0)
+      store <- Ref[IO].of(Map.empty[Int, Todo])
+    } yield new App(id, store)
+
+  def serve(a: App): Resource[IO, ListeningServer] =
```
joroKr21, line 42:
```suggestion
  def serve(app: App): Resource[IO, ListeningServer] =
```

examples/src/main/scala/io/finch/todo/Main.scala
```diff
@@ -27,28 +29,37 @@ import io.finch.internal._
...
+    Bootstrap[IO]
+      .serve[Application.Json](a.getTodos :+: a.postTodo :+: a.deleteTodo :+: a.deleteTodos :+: a.patchTodo)
+      .serve[Text.Html](classpathAsset("/todo/index.html"))
+      .serve[Application.Javascript](classpathAsset("/todo/main.js"))
+      .listen(s":${port()}")
+
+  def run(): IO[Unit] =
```
joroKr21, line 49:
```suggestion
  val run: IO[Unit] =
```

joroKr21 (changes requested):
Some cosmetic changes, but also we need to rename some things which would not be binary compatible later. Sorry for the long wait, I was MIA due to personal circumstances. I will address my own comments a bit later 😅

joroKr21 (approved):

PR #1483, opened 2022-04-26
kind-projector 0.13.2 (was 0.10.3)

SethTisue:
Scala 2.12.16 and 2.13.9 will prohibit the use of `?` in this way, so this changes to the newer `*` syntax which will also work in Scala 3


joroKr21 (approved):
Looks like Scala steward missed that one because of the cross version change

Issue #1254, opened 2020-08-12
Hello World example from readme seems not to work


KovaxG:
Hi, I really like the idea of finch, and I'm trying to test it out, but sadly I keep hitting obstacles.
First there was an issue with the versions, ( #1194) but I finally solved that. But now I can't get the example hello world to work either.

I'm trying 
```scala
import io.finch._
import cats.effect.IO
import com.twitter.finagle.Http

object Server extends App with Endpoint.Module[IO] {
  val api: Endpoint[IO, String] = get("hello") { Ok("Hello, World!") }
  Http.server.serve(":8080", api.toServiceAs[Text.Plain])
}
```
but all I get is
```
Aug 12, 2020 1:16:58 PM com.twitter.finagle.Init$ $anonfun$once$1
INFO: Finagle version 20.3.0 (rev=e3fd6d1bd88c9ed7edeb98a6598e23800924c633) built at 20200308-110626

Process finished with exit code 0
```
I'm not sure if the problem is on my side or with the example.

rpless:
Hi @KovaxG I believe this is an issue with the docs. I think #1250 fixes it. I will try to take a look and merge it this weekend.
Basically, `Http.server.serve` is returning a future that represents the life time of the server. Because it is a future it returns immediately and then the program exits. You can wrap it in `Await.ready` to block it until the server shuts down like so:
```scala
Await.ready(Http.server.serve(":8080", api.toServiceAs[Text.Plain]))
```

hderms:
I can confirm this is how we use Finch, as well

PR #1526, opened 2022-10-17
Update Hello World example

lobzison:
Seems that after the `0.34.0` release some things have changed, but they are not reflected in the examples.

There is a https://github.com/finagle/finch/issues/1254, but it seems to be opened for a different root cause. But it can be closed after the merge


joroKr21 (approved):
Nice, thanks!

PR #1543, opened 2022-11-14
Update cats to `2.9.0` and cats-effect to `3.4.0`

tpetillot:
**Modifications:**
Use previsouly default `Dispatcher.parallel` when converting to finagle
service


tpetillot:
I created this PR also in order to discuss the release of a `0.34.1` to benefit from version upgrades.

core/src/main/scala/io/finch/Bootstrap.scala
```diff
@@ -123,7 +123,7 @@ class Bootstrap[F[_], ES <: HList, CTS <: HList](
 
   def toService(implicit F: Async[F], ts: Compile[F, ES, CTS]): Resource[F, Service[Request, Response]] = {
     val compiled = compile
-    Dispatcher[F].flatMap { dispatcher =>
+    Dispatcher.parallel[F].flatMap { dispatcher =>
```
tpetillot, line 126: This was the default behavior prior 3.4.0, but I wonder if we should not make this to be parametrable.

joroKr21 (approved):
Nice! I don't remember if we merged only updates but feel free to cut a release 👍

PR #1546, opened 2022-11-15
chore: allow to pass a custom dispatcher on `toService`

tpetillot:


tpetillot:
I will trigger a release after this one to benefit from a FIFO policy if needed

